Issue #19, opened 2018-04-05
Allow passing playbooks/roles directly to run()


matburt:
Currently these playbooks and roles must be present in the private_data_dir and playbook is selected as a filename from their.

By implementing this we could remove the  absolute requirement to have the private_data_dir

Issue #18, opened 2018-04-05
Allow taking inventory from a datastructure given to run()


matburt:
Currently this can be overridden on a basic level but we should be able to take a full json or yaml datastructure to use instead of the metadata inventory/ directory

privateip:
I believe this is now closed by #21

PR #21, opened 2018-04-13
adds new function to interface for passing playbooks directly

privateip:
This adds a new function to ansible_runner.interface call api().  This
function accepts the playbook as a native Python object and inventory as
a string. It will write the playbook and inventory to disk in temp and
execute the playbook.

This is the foundation for fully supporting #19


privateip:
This PR addresses #18 and #19

privateip:
@radez good catch!   kwarg description updated

radez:
thx for the update, not sure it came in accurate though?
Shouldn't it read something like: "The inventory kwarg can be provided as either a native Python dict supporting YAML inventory structure or a string providing the path to an INI formatted hosts file."

The update feels like it indicates that you can still pass a string with INI formatted content but the code is actually looking for a file name to an INI file when you pass a string.

Thx @privateip

privateip:
@matburt PR updated with refactoring per comments in review.

Issue #30, opened 2018-04-30
When a file is not loaded, a reason should be present


cdvv7788:
I noticed the following behaviour (test.py just contains the 2 lines in the readme for running from python):
![image](https://user-images.githubusercontent.com/5531776/39442058-9ac657a0-4c75-11e8-83ac-544081007160.png)

The file could fail to load for several reasons. I guess that those are mainly:
- File does not exist
- File is not well formatted

With the current message, the user has no idea if the system is not loading it correctly (because there is some issue with the naming convention) or is not able to parse (syntax issue in the image).

Should that be validated to improve the messages?

privateip:
@cdvv7788 Agreed with your comment.  I'm working on a major refactor of runner_config that will address this issue (among some others) but its still a couple of weeks out before its ready.  Will make a note and try to remember to reference this issue once the PR is up.

privateip:
@cdvv7788 can you confirm #32 addresses your use case?

cdvv7788:
With the proposed fix at https://github.com/ansible/ansible-runner/pull/66 the messages for ssh_key should improve.

matburt:
I think this is taken care of?

```
ERROR! Invalid extra vars data supplied. '@emo/env/extravars' could not be made into a dictionary
```

Let me know and we can re-open and include an example that shows the error.

PR #32, opened 2018-05-15
implement Python logging

privateip:
This change implements the Python logging facility for handling message
logging to both stdout and to a debug log file.  To use the logging
facility perform the following in module code:

```
import logging
log = logging.getLogger('ansible-runner')
log.info('i can now log message')
```

To send messages to stdout use the display function in `utils.py`.

### DIFF
--- a/ansible_runner/interface.py
+++ b/ansible_runner/interface.py
@@ -8,10 +8,37 @@ import json
 import stat
 import sys
 import os
+import logging
 
 from .runner_config import RunnerConfig
 from .runner import Runner
-from .utils import to_artifacts
+from .utils import dump_artifacts, configure_logging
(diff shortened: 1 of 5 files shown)


privateip:
This PR should  resolve #30

matburt:
Looks like there are conflicts with `master` now :open_mouth:

privateip:
> Should we expose more of these details through the command line interface also?

Yes, I figure I would wait until we got #35 merged before bubbling up that option.  Once #35 is agreed upon and merged, I will add that argument.  In the mean time I opened https://github.com/ansible/ansible-runner/issues/40 to track this feature

privateip:

PR #33, opened 2018-05-15
adds loader module to handle file loads

privateip:
This change adds a new module loader.py that will now handle loading of
artifacts from disk.  This module will cache all file contents and
encode the string contents to bytes.

### DIFF
--- a/ansible_runner/loader.py
+++ b/ansible_runner/loader.py
@@ -51,7 +51,7 @@ class ArtifactLoader(object):
         Args:
             contents (string): The contents to deserialize
 
-        Retunrs:
+        Returns:
             dict: If the contents are JSON serialized
 
             None: if the contents are not JSON serialized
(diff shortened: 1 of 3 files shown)


test/unit/test_loader.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # type check fails
+        with raises(ConfigurationError):
+            res = loader.load_file('/tmp/test', dict)
+            assert res is not None
+
+def test_get_contents_ok(loader):
+    with patch('ansible_runner.loader.open') as mock_open:
+        handler = BytesIO()
+        handler.write("test string")
```
matburt: After rebasing against `master` which includes python 3 support this line throws the following error:

```
TypeError: a bytes-like object is required, not 'str'
```

ansible_runner/loader.py
```diff
@@ -0,0 +1,175 @@
...
+        Transform the path to an absolute path
+
+        Args:
+            path (string): The path to transform to an absolute path
+
+        Returns:
+            string: The absolue path to the file
+        '''
+        if path.startswith(os.path.sep) or path.startswith('~'):
+            return os.path.expanduser(path)
```
matburt: `expanduser` could wrap the `os.path.join` on L123 since it will expand `~` if it exists or ignore it otherwise.

privateip:

PR #35, opened 2018-05-19
update ansible-runner arguments

privateip:
This change adds support for directly running roles from the cli interface.  To
execute a role directly without the need for a playbook, use the
following command

ansible-runner run . --role <role-name>

This will execute the role against the inventory.  Two additional
command line switches have been added.

* roles-path - used to set the path to the roles directory
* role-vars - used to inject vars to the role

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -29,13 +29,20 @@ import shlex
 
 from uuid import uuid4
 
-from ansible_runner import run, run_async
+from ansible_runner import run
+from ansible_runner.utils import dump_artifact
+from ansible_runner.runner import Runner
 
 VERSION = pkg_resources.require("ansible_runner")[0].version
-DEFAULT_ROLES_PATH = os.getenv('ANSIBLE_ROLES_PATH', '/etc/ansible/roles')
+
+DEFAULT_ROLES_PATH = os.getenv('ANSIBLE_ROLES_PATH', None)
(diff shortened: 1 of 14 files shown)


ansible_runner/__main__.py
```diff
@@ -0,0 +1,153 @@
...
+import sys
+import os
+import shlex
+
+from uuid import uuid4
+
+from ansible_runner import run, run_async
+
+VERSION = pkg_resources.require("ansible_runner")[0].version
+DEFAULT_ROLES_PATH = os.getenv('ANSIBLE_ROLES_PATH', '/etc/ansible/roles')
```
matburt: This should default to our project base directory (aka private_data_dir) and can also include `/etc/ansible/roles`

Something like `{project}/roles` ?

matburt: In much the same way as we allow `RUNNER_PLAYBOOK` we should also allow `RUNNER_ROLE` also.

matburt: Those would be mutually exclusive I would think.

Adding a demo role would be hot also ;)

setup.py
```diff
@@ -20,7 +20,7 @@
     zip_safe=False,
     entry_points={
         'console_scripts': [
-            'ansible-runner=ansible_runner.interface:main',
-        ],
-    },
+            'ansible-runner = ansible_runner.__main__:main'
```
matburt: :+1: I like the rename to `__main__` for the module and that follows best practice.

privateip:

PR #36, opened 2018-05-21
Support python 3

matburt:
* Correct use of threading
* Fix up encoding issue on event partials
* Fix up dictionary modification when processing passwords and event data

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -2,7 +2,7 @@ import os
 import re
 import yaml
 import pipes
-import thread
+import threading
 import pexpect
 from uuid import uuid4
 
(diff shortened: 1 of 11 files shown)


ansible_runner/utils.py
```diff
@@ -9,7 +8,7 @@
 import hashlib
 
 from collections import Iterable, Mapping
-
+from io import StringIO
```
cchurch: `six.StringIO` unless you're planning on dropping Python 2.7 support entirely.

ansible_runner/runner_config.py
```diff
@@ -177,8 +177,8 @@ def open_fifo_write(self, path, data):
         reads data from the pipe.
         '''
         os.mkfifo(path, 0o600)
-        thread.start_new_thread(lambda p, d: open(p, 'w').write(d), (path, data))
-
+        threading.Thread(target=lambda p, d: open(p, 'w').write(d),
+                         args=(path, data))
```
cchurch: Also need to call `start()` on the resulting `Thread` instance.

ansible_runner/utils.py
matburt: It looks like `io.StringIO` exists in 2.7 at least?

ansible_runner/runner_config.py
matburt: good catch

PR #37, opened 2018-05-23
Adding Openshift deployment support

matburt:
* Move entrypoint and add configuration for openshift random user via passwd magic

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,173 @@
+.git/
+
+### Python template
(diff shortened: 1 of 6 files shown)


j00bar:
You'll want to leave in the `str`'ing of environment variables, methinks. They should never be other types. And a .dockerignore is recommended, even if you don't use mine :)

matburt:
done

PR #43, opened 2018-06-01
Update pexpect to 4.6.0

tbreeds:
In fce332c (Upgrading pexpect (and some ancillary packages) to 4.6) the
Pipfile was updated to use pexpect 4.6.0 but setup.py wasn't.  This
seems strange so update setup.py to match.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ setup(
     packages=find_packages(),
     install_requires=[
         'psutil',
-        'pexpect==4.5.0',
+        'pexpect==4.6.0',
         'python-daemon',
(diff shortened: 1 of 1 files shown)


matburt:
I think I'm going to remove this requirement specific targetting by default, actually.

The reason for pegging it at 4.5 was to help with Fedora packaging for 1.0.3

matburt:
see: https://github.com/ansible/ansible-runner/commit/29c09c501103124765c18140b58ada9e1a05c2cf

tbreeds:
Thanks

PR #45, opened 2018-06-02
adds unit tests for RunnerConfig

privateip:
This change adds unit test coverage for RunnerConfig and makes no
changes to the actual code.  All methods are tested as they are
currently written.


matburt:
:fist_oncoming: :boom:

PR #50, opened 2018-06-06
roll up of fixes for running roles from command line

privateip:
This change includes a role up of a number of fixes to the execution of
roles from the command line.  Namely these changes address problems with
artifacts being left behind after ansible-runner has completed.  Once
ansible-runner completes, it will now clean up any dynamic artificats it
created.

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -26,6 +26,7 @@ import stat
 import sys
 import os
 import shlex
+import shutil
 
 from uuid import uuid4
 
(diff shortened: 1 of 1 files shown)


ansible_runner/__main__.py
```diff
@@ -107,57 +109,81 @@ def main():
...
                 res = run(**kwargs)
                 rc = res.rc
 
+            except AnsibleRunnerException as exc:
+                errmsg = str(exc)
+
             finally:
-                if playbook and os.path.isfile(playbook):
+                if not project_exists and os.path.exists(project_path):
+                    print('removing dynamically generated project folder')
```
matburt: These prints should probably go through the debug interface?

PR #55, opened 2018-06-12
Replaces os.mknod with portable equivalent

benthomasson:
os.mknod is not available to users on Mac OS.  os.mknod causes two tests
to fail: test_cmdline_playbook and test_cmdline_playbook_hosts.
Replacing os.mknod with os.open and setting permissions on create in one
step preserves security, makes the code portable, and makes the
test pass.


cdvv7788:
This seems to fix https://github.com/ansible/ansible-runner/issues/72

PR #59, opened 2018-06-13
Minor format fixing

yujunz:


docs/python_interface.rst
```diff
@@ -14,7 +14,7 @@ process/store it for use later.
 Helper Interfaces
 -----------------
 
-The :mod:`helper interfaces <ansible_runner.interface>` provide a quick way of providing the recommended inputs in order to launch a **Runner** process.
+The helper :mod:`interfaces <ansible_runner.interface>` provide a quick way of providing the recommended inputs in order to launch a **Runner** process.
```
cdvv7788: I think that it should be "provides" instead of "provide".

yujunz: Done

PR #69, opened 2018-06-18
expose ansible-playbook verbosity to cli and api

### LINKED ISSUE #41: How to control verbosity from python
I tried the latest HEAD and I am getting the following with the demo (with the python commands in the readme):

image

I would like to see the full traceback.
As a sidenote, I have psutil installed:
image
And can even run commands with it:
image

privateip:
This change adds a new config option to increase the verbosity of the
ansible playbook output.  In the command line form, you can now specify
`-v`, `-vv`, `-vvv`, etc and it will be passed to the ansible-playbook
at run time.  In the api form, set the kwarg `verbosity` to the number
of `-v`'s to append to `ansible-playbook`.

This closes #41

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -77,7 +77,7 @@ def main():
     parser.add_argument("--inventory")
 
     parser.add_argument("-v", action="count",
-                        help="Increase the verbosity of the ansible-playbook output")
+                        help="Increase the verbosity with multiple v's (up to 5) of the ansible-playbook output")
 
     args = parser.parse_args()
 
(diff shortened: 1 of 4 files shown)


ansible_runner/__main__.py
```diff
@@ -76,6 +76,9 @@ def main():
 
     parser.add_argument("--inventory")
 
+    parser.add_argument("-v", action="count",
+                        help="Increase the verbosity of the ansible-playbook output")
```
matburt: add a note that multiple `v`s will increase verbosity  further.

PR #70, opened 2018-06-19
add new feature for mananging cmdline options

### LINKED ISSUE #63: Cannot pass --tags and --skip-tags to ansible-runner
Cannot pass --tags and --skip-tags to ansible-runner. Is there some alternative way to do this?

privateip:
This change adds a new experimental feature to `ansible-runner` to
manage command line options passed to the `ansible-playbook` executable.
This feature does not validate the command line options so it is up to
the playbook writer to make sure to include the correct values.

In order to use the command line options create a new file in the `env/`
folder called `cmdline` and put the command line string into the file.

```
env/cmdline

--tags one,two --skip-tags three -u ansible --become
```

This change updates the docs as well

This PR will fix #63

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -91,6 +91,9 @@ def main():
     parser.add_argument("-v", action="count",
                         help="Increase the verbosity with multiple v's (up to 5) of the ansible-playbook output")
 
+    parser.add_argument("--cmdline",
+                        help="Command line options to pass to ansible-playbook at execution time")
+
     args = parser.parse_args()
 
     pidfile = os.path.join(args.private_data_dir, 'pid')
@@ -111,7 +114,6 @@ def main():
 
     if args.command in ('start', 'run'):
         if args.role:
-
             role = {'name': args.role}
             if args.role_vars:
                 role_vars = {}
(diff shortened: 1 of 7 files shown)


privateip:
> It might be good to use this as a chance to also support argument pass-through support on the command line in the form of:
$ ansible-runner [runner args, command, dir] -- [additional ansible args]
What do you think?

Agreed... will update PR accordingly

privateip:
I implemented a slight deviation from your proposal.  I added a command line option called `cmdline`.  To use it form the command line run `ansible-runner` like the following example:

```
ansible-runner run /home/sprygada/Workspaces/dev-runner/run -r test --inventory hosts --host localhost --cmdline "--tags bar -vvvvv"
```

I think this PR would make #69 obsolete and therefore it can be closed without merging it.

matburt:
heh welp... I went bottom up when reviewing PRs so I merged that one... how do you want to proceed with this?

My problem with doing it this way (in #70) as opposed to also passing through our own logic in some cases (-v)  is that it gives us less of a chance to be able to change the way runner itself behaves when given these arguments.

I'd love for -v to also increase the verbosity of Runner for example... tags/skip-tags could also be exposed elsewhere (not saying we should do that now or here)

privateip:
resolved conflicts, all tests passing

PR #71, opened 2018-06-20
Enhance the logging and display output

privateip:
This change provides a significant refactoring of the logging and
display facility to better provide output and debug functions.  It moves
the logging capabilities from utils into a separate singleton module
'output' and exposes methods to configure and produce output.

This change updates the entry points for both cli and api operation to
configuring the output.  To send output to there are two convience
functions available: `display` and `debug`.  The `display` function will
send output to `stdout` and to a log file (if enabled).  The `debug`
function will only produce output of the `--debug` option is specified
or the `debug=True` is set from the api interface.

Additionally all logging can be sent to a logfile using the `--logfile`
cli option or `logfile=<path>` for the api interface.

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -34,6 +34,7 @@ from uuid import uuid4
 from yaml import safe_load
 
 from ansible_runner import run
+from ansible_runner import output
 from ansible_runner.utils import dump_artifact
 from ansible_runner.runner import Runner
 from ansible_runner.exceptions import AnsibleRunnerException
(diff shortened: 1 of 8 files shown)


matburt:
A good bit has changed over the course of a few merges... can you rebase this against `master`?

privateip:
merge conflicts resolved, all tests passing

privateip:
As an fyi, this PR will set the foundation that will allow us to honor `-v` in `ansible-runner` as well as pass along to `ansible-playboo`

PR #74, opened 2018-06-23
Accept option for custom artifacts dir

cdvv7788:
As required at https://github.com/ansible/ansible-runner/issues/10, I added the option on RunnerConfig to accept a custom artifact_dir argument. Is this aligned with what you had in mind with your issue @matburt ?
I tested this on the python interface. Will this need to be added somewhere else to be supported by the cli tool?

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -76,6 +76,8 @@ def main():
 
     parser.add_argument("--role-skip-facts", action="store_true", default=False,
                         help="Disable fact collection when executing a role directly")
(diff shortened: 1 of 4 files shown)


matburt:
This is what I had in mind.

Make sure it will default the artifact dir to inside the private data dir and make sure it's specifiable from both the command line interface and the python module interface

matburt:
looks like some integration conflicts so you'll need to rebase.

cdvv7788:
@matburt Rebased and added a test.

PR #77, opened 2018-06-25
Fix issue when loading string types

cdvv7788:
This solves https://github.com/ansible/ansible-runner/issues/65
For more context, check https://github.com/ansible/ansible-runner/pull/66

### DIFF
--- a/ansible_runner/loader.py
+++ b/ansible_runner/loader.py
@@ -21,6 +21,7 @@ import json
 import logging
 
 from yaml import safe_load, YAMLError
+from six import string_types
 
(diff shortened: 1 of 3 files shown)


cdvv7788:
I will add a test covering the issue #65.

cdvv7788:
Updated a test to catch the issue that was happening. Basically, the test was not detecting the issue because it had no line breaks and the objtype was not being specified.
As a side effect of the changes, the load_file method returns a binary instead of a string, if string_types is specified. This happens because before everything passed through a serializer, which converted it.
Everything seems to be working, just wanted to make that clear.

Issue #75, opened 2018-06-24
Where are the usage examples on the new documentation?


cdvv7788:
I could not find them. Were they removed?
The need to be added back (imho). If they are already there, they need to be easier to find.
I can create a pr for any of those options.

PR #79, opened 2018-06-26
Provide python example for docs

cdvv7788:
As mentioned at https://github.com/ansible/ansible-runner/issues/75, the code samples went missing.
This PR intends to display what I mean by adding them back. If you think that they are no longer needed, that's ok too.


matburt:
Thanks for the example, yeah... I think it'd be good to have a few examples directly in the main area of the docs. I reckon I had thought just linking directly to the module docs would be good enough but it *is* useful to see some basic examples here.

Issue #54, opened 2018-06-12
No module named psutil


cdvv7788:
When running the demo python code, I am getting:
![image](https://user-images.githubusercontent.com/5531776/41295389-1d34c67e-6e20-11e8-90e5-d16f70271805.png)

I am running it using the following environment:
![image](https://user-images.githubusercontent.com/5531776/41295456-408f0364-6e20-11e8-90da-48d2965b671d.png)

As it can be seen in the attached image, I do have psutil. Ansible-runner is updated.

cdvv7788:
The issue was happening because I didn't have ansible installed in my venv. Not familiar enough to know if that is the case, but shouldn't ansible be a dependency on setup?
Also, the error does not give a hint about the cause.

matburt:
This has a bit to do with the Runner callback plugin/interface needing psutil in order to handle the processes spawned by Ansible itself. We *do* have psutil installed alongside Runner but if Ansible itself exists *outside* of your venv then it obviously won't have access to it.

Not usually a problem for standard packaging but it is in this scenario. I'm open to suggestions on this one.

cdvv7788:
Well, I didn't have ansible installed at all. At least for the python interface, it may be natural for ansible to be a dependency for this package. After all, this package does depend on ansible to be able to work.
Other than that, I don't know what else to suggest at this time.

matburt:
I've thought about having ansible as a dependency but it takes away some of the flexibility of the tool itself.

I want folks to be able to bring any supported ansible version that they want, and be able to pull it from any source. Even something as flexible as having runner be pip installed in python 3 and ansible installed from distro packaging into python 2.

Adding a hard dependency on Ansible itself makes that very difficult.

cdvv7788:
Well, that makes a lot of sense. Is there a way to detect that ansible is missing? I mean, that may be obvious once you find out, but if for some reason the tool is unable to find ansible, all we have is a message saying that "no module named psutil".

matburt:
Yep, you are right. We could/should add a pre-launch check eithere here:

https://github.com/ansible/ansible-runner/blob/master/ansible_runner/interface.py#L30

or, and this is probably better, as part of these:

https://github.com/ansible/ansible-runner/blob/master/ansible_runner/runner_config.py#L90

[...]

PR #81, opened 2018-06-27
Add check to make sure that ansible is accessible

cdvv7788:
This attempts to solve https://github.com/ansible/ansible-runner/issues/54
The idea is to check if ansible is present before doing anything else.


cdvv7788:
Not compatible with python2. Will add an alternative for that case.

cdvv7788:
Requires rebase and does not work for python 2. I will create another PR with a better approach.

PR #86, opened 2018-07-07
add support for running roles via api interface

### LINKED ISSUE #64: Structure in docs does not correspond with functionality
According to https://ansible-runner.readthedocs.io/en/latest/intro.html#runner-input-directory-hierarchy

The roles directory is outside of project folder:
image
When I set up my structure like that, I get:

    The role xxx was not found in /tmp/tmp2le4eyz0/project/roles:/home/ubuntu/.ansible/roles:/usr/share/ansible/roles:/etc/ansible/roles:/tmp/tmp2le4eyz0/project

Obviously, moving roles to be inside the project folder fixes the issue.
Is this a bug or a documentation error? If the error is in the documentation, let me know and I can make the change.

privateip:
This change expands on the role support to allow roles to be run
directly from the interface module.  Roles can now be called
programatically as well as via the command line.

This commit fixes #64

### DIFF
--- a/ansible_runner/utils.py
+++ b/ansible_runner/utils.py
@@ -101,6 +101,29 @@ def dump_artifacts(kwargs):
     if not os.path.exists(private_data_dir):
         raise ValueError('private_data_dir path is either invalid or does not exist')
 
+    if 'role' in kwargs:
+        role = {'name': kwargs.pop('role')}
+        if 'role_vars' in kwargs:
+            role['vars'] = kwargs.pop('role_vars')
+
+        play = [{'hosts': kwargs.pop('hosts', 'all'), 'roles': [role]}]
+
+        if kwargs.pop('role_skip_facts', False):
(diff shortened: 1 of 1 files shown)


ansible_runner/utils.py
```diff
@@ -101,6 +101,29 @@ def dump_artifacts(kwargs):
...
+        if kwargs.pop('role_skip_facts', False):
+            play[0]['gather_facts'] = False
+
+        kwargs['playbook'] = play
+
+        if 'envvars' not in kwargs:
+            kwargs['envvars'] = {}
+
+        roles_path = kwargs.pop('roles_path', None)
+        if not roles_path:
```
cdvv7788: If there is an if-else block, the if should contain the condition without negation.
Something like:
```
if roles_path:
    # append to roles_path
else:
    # create roles_path
```
Something very minor and that does not matter much, but can make code slightly more readable.

Issue #90, opened 2018-07-11
eventlet 0.20.0 and newer doesn't work with ansibler-runner


cubeek:
ansible-runner uses pexpect that uses poll() for child processes monitoring. select.poll() has been removed from eventlet in 0.20.0 and projects like OpenStack that use eventlet monkey_patching cannot use ansible-runner.

2018-07-11 15:15:45.604 TRACE networking_ansible.ansible_networking Traceback (most recent call last):
2018-07-11 15:15:45.604 TRACE networking_ansible.ansible_networking   File "/opt/stack/networking-ansible/networking_ansible/ansible_networking.py", line 129, in vlan_access_port
2018-07-11 15:15:45.604 TRACE networking_ansible.ansible_networking     segmentation_id, switch_port)
2018-07-11 15:15:45.604 TRACE networking_ansible.ansible_networking   File "/opt/stack/networking-ansible/networking_ansible/ansible_networking.py", line 75, in _run_task
2018-07-11 15:15:45.604 TRACE networking_ansible.ansible_networking     inventory=self.inventory)
2018-07-11 15:15:45.604 TRACE networking_ansible.ansible_networking   File "/usr/lib/python2.7/site-packages/ansible_runner/interface.py", line 94, in run
2018-07-11 15:15:45.604 TRACE networking_ansible.ansible_networking     r.run()
2018-07-11 15:15:45.604 TRACE networking_ansible.ansible_networking   File "/usr/lib/python2.7/site-packages/ansible_runner/runner.py", line 105, in run
2018-07-11 15:15:45.604 TRACE networking_ansible.ansible_networking     searchwindowsize=100)
2018-07-11 15:15:45.604 TRACE networking_ansible.ansible_networking   File "/usr/lib/python2.7/site-packages/pexpect/spawnbase.py", line 341, in expect
2018-07-11 15:15:45.604 TRACE networking_ansible.ansible_networking     timeout, searchwindowsize, async_)
2018-07-11 15:15:45.604 TRACE networking_ansible.ansible_networking   File "/usr/lib/python2.7/site-packages/pexpect/spawnbase.py", line 369, in expect_list
2018-07-11 15:15:45.604 TRACE networking_ansible.ansible_networking     return exp.expect_loop(timeout)
2018-07-11 15:15:45.604 TRACE networking_ansible.ansible_networking   File "/usr/lib/python2.7/site-packages/pexpect/expect.py", line 111, in expect_loop
[...]

PR #91, opened 2018-07-11
Use select() instead of poll()

cubeek:
poll() is not compatible with eventlet >= 0.20.0. This patch switches
ansible runner to use select() when communicating with child processes.
The disadvantage is that select() supports only 1024 file descriptors
but for runner it should be enough.

Fixes issue: #90

### DIFF
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -98,7 +98,7 @@ class Runner(object):
             ignore_sighup=True,
             encoding='utf-8',
             echo=False,
-            use_poll=True,
+            use_poll=self.config.pexpect_use_poll,
         )
         child.logfile_read = stdout_handle
         last_stdout_update = time.time()
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -172,6 +172,8 @@ class RunnerConfig(object):
         self.idle_timeout = self.settings.get('idle_timeout', 120)
         self.job_timeout = self.settings.get('job_timeout', 120)
         self.pexpect_timeout = self.settings.get('pexpect_timeout', 5)
+
+        self.pexpect_use_poll = self.settings.get('pexpect_use_poll', True)
         self.suppress_ansible_output = self.settings.get('suppress_ansible_output', self.quiet)
 
         if 'AD_HOC_COMMAND_ID' in self.env:
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -136,6 +136,9 @@ The **settings** file is a little different than the other files provided in thi
   idle_timeout: 600 # If no output is detected from ansible in this number of seconds the execution will be terminated.
   job_timeout: 3600 # The maximum amount of time to allow the job to run for, exceeding this and the execution will be terminated.
   pexpect_timeout: 10 # Number of seconds for the internal pexpect command to wait to block on input before continuuing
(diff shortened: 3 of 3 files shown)


ryanpetrello:
Is it possible that this behavior be made configurable?  https://github.com/ansible/awx relies on the usage of `poll` to work around a notable issue with `select()`:

https://github.com/pexpect/pexpect/issues/47

cubeek:
Aha, I didn't know about awx. I'll look into it. Thanks for a quick response!

cubeek:
@ryanpetrello thanks, that's actually pretty much what I came up with :D I tested that it works with networking-ansible openstack driver.

ryanpetrello:
One more thing @cubeek - you mind documenting this new option at https://github.com/ansible/ansible-runner/blob/35f9eb6b76f341d9ea70042e5d89156916858bea/docs/intro.rst#envsettings---settings-for-runner-itself ?

matburt:
If we can add some docs, we'll get this merged

cubeek:
Sure, I'll add the docs

matburt:
I probably should have run through the unit tests first:

```
    def run(self):
        '''
            Launch the Ansible task configured in self.config (A RunnerConfig object), returns once the
            invocation is complete
            '''
        self.status = "starting"
        stdout_filename = os.path.join(self.config.artifact_dir, 'stdout')
    
        try:
            os.makedirs(self.config.artifact_dir)
            os.close(os.open(stdout_filename, os.O_CREAT, stat.S_IRUSR | stat.S_IWUSR))
        except OSError as exc:
            if exc.errno == errno.EEXIST and os.path.isdir(self.config.artifact_dir):
                pass
            else:
                raise
    
        stdout_handle = codecs.open(stdout_filename, 'w', encoding='utf-8')
        stdout_handle = OutputEventFilter(stdout_handle, self.event_callback, self.config.suppress_ansible_output, output_json=self.config.json_mode)
    
        if not isinstance(self.config.expect_passwords, collections.OrderedDict):
            # We iterate over `expect_passwords.keys()` and
            # `expect_passwords.values()` separately to map matched inputs to
            # patterns and choose the proper string to send to the subprocess;
            # enforce usage of an OrderedDict so that the ordering of elements in
            # `keys()` matches `values()`.
            expect_passwords = collections.OrderedDict(self.config.expect_passwords)
        password_patterns = list(expect_passwords.keys())
        password_values = list(expect_passwords.values())
    
        # pexpect needs all env vars to be utf-8 encoded strings
        # https://github.com/pexpect/pexpect/issues/512
        for k, v in self.config.env.items():
            if k != 'PATH' and isinstance(v, six.text_type):
                self.config.env[k] = v.encode('utf-8')
    
        self.status = 'running'
        child = pexpect.spawn(
            self.config.command[0],
            self.config.command[1:],
            cwd=self.config.cwd,
            env=self.config.env,
            ignore_sighup=True,
            encoding='utf-8',
            echo=False,
>           use_poll=self.config.pexpect_use_poll,
        )
E       AttributeError: 'RunnerConfig' object has no attribute 'pexpect_use_poll'
```

cubeek:
Whoopsie, I'll send a PR asap

cubeek:
Addressed at https://github.com/ansible/ansible-runner/pull/93 - sorry for breaking it.

matburt:
it's no problem, `tox` at the top of the tree will run all tests under py2 and py3 as well as flake8

Issue #92, opened 2018-07-12
init_runner overrides all logging configuration


matburt:
When working on the jupyter kernel integration: https://github.com/ansible/ansible-jupyter-kernel

I noticed that init_runner's desire to configure all logging will wipe out any python system importing runner and using the built-in interfaces.

matburt:
This should be fixed in #94

PR #94, opened 2018-07-12
WIP: Adding external event_handler and overriding logging behavior

matburt:
This represents work items needed for my integration of Runner into the Ansible Jupyter kernel: https://github.com/ansible/ansible-jupyter-kernel

Specifically this branch: https://github.com/matburt/ansible-jupyter-kernel/tree/ansible_runner_backend

while working on this integration I also found this: https://github.com/ansible/ansible-runner/issues/92

This features likely will/would also be enjoyed by other module interface consumers:
* An event callback handler can now be passed down into the Runner to
  get callbacks related to events that Runner processes
* Adds an option to the interface helpers to completely
  override/bypass Runner's logging initialization
* A `finished` callback event option

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -145,7 +145,7 @@ def main():
                 role['vars'] = role_vars
 
             kwargs = dict(private_data_dir=args.private_data_dir,
-                          json_mode=args.json)
+                          json_mode=args.json, ignore_logging=False)
             if args.artifact_dir:
                 kwargs['artifact_dir'] = args.artifact_dir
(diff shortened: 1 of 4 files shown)


matburt:
I think logging makes sense when it's invoked as part of the cli, but for module imports this should be bypassed entirely.

matburt:
Indeed that is pretty redundant, I'm not sure how that snuck in there.

matburt:
Actually... nothing in __main__ calls into the `init_runner` interface methods... so not technically redundant.

Issue #98, opened 2018-07-24
unify role execution code path


privateip:
currently the code path when executing roles from the cli vs executing roles via the api are different.

AlanCoding:
Could somebody give a little bit of clarification on what this issue means?

I will explain where I am coming from. As a newcomer, I'm trying to read the nuts-and-bolts docs on how to lay out a basic ansible-runner project.

https://ansible-runner.readthedocs.io/en/latest/intro.html#runner-input-directory-hierarchy

```
.
├── env
│   ├── envvars
│   ├── extravars
│   ├── passwords
│   ├── cmdline
│   ├── settings
│   └── ssh_key
├── inventory
│   └── hosts
├── project
│   └── test.yml
└── roles
    └── testrole
        ├── defaults
        ├── handlers
        ├── meta
        ├── README.md
        ├── tasks
        ├── tests
        └── vars
```

That has the roles directory as a top-level dir in the runner (`--project-dir`?) directory.

But the demo project, linked directly below that tree:

https://github.com/ansible/ansible-runner/tree/master/demo/project

has a roles directory nested under the `project/` directory.

```
$ tree demo
demo
├── env
│   ├── envvars
│   ├── extravars
│   ├── passwords
│   ├── settings
│   └── ssh_key
├── inventory
│   └── hosts
└── project
    ├── roles
    │   └── testrole
    │       ├── README.md
    │       ├── defaults
    │       │   └── main.yml
    │       ├── handlers
    │       │   └── main.yml
    │       ├── meta
    │       │   └── main.yml
    │       ├── tasks
    │       │   └── main.yml
    │       ├── tests
    │       │   ├── inventory
    │       │   └── test.yml
    │       └── vars
    │           └── main.yml
    └── test.yml
```

If both are possible, that's fine, but I would like to get a preferred structure, and have all the newcomer materials reflect that.

matburt:
The code paths that execute Runner branch at an early point:

https://github.com/ansible/ansible-runner/blob/116d23702496e415b1bd625b1a9f1b65ba4219a2/ansible_runner/__main__.py#L55

This whole mechanism under the role manager and everything that leads up to this conditional execution (like this):

https://github.com/ansible/ansible-runner/blob/116d23702496e415b1bd625b1a9f1b65ba4219a2/ansible_runner/__main__.py#L301

mean that getting from entrypoint -> valid Runner object is needlessly complex.

[...]

PR #114, opened 2018-08-08
Changes role execution to use the same run call as non-roles

benthomasson:
See: #98

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -28,6 +28,7 @@ import sys
 import os
 import shlex
 import shutil
+from contextlib import contextmanager
 
 from uuid import uuid4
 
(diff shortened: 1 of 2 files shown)


benthomasson:
This is half of the solution.   @privateip asked for the API to have a unified execution path as well.

ansible_runner/__main__.py
```diff
@@ -49,6 +50,106 @@
 logger = logging.getLogger('ansible-runner')
 
 
+class Bunch(object):
```
matburt: This feels a little dirty to me... I'd almost rather just see direct dict mappings or some other way (namedtuples?) to map to properties.

I do like the usage of role_manager as a `contextmanager` though. We should keep that.

PR #115, opened 2018-08-08
Fixes handle_termination by removing unfinished wrapper handler code

benthomasson:
See: #111

### DIFF
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -137,8 +137,7 @@ class Runner(object):
                 # if isinstance(extra_update_fields, dict):
                 #     extra_update_fields['job_explanation'] = "Job terminated due to timeout"
             if self.canceled or self.timed_out or self.errored:
-                # TODO: proot_cmd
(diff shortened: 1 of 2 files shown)


ansible_runner/runner.py
```diff
@@ -259,32 +258,26 @@ def host_events(self, host):
...
-                    try:
-                        main_proc = psutil.Process(pid=pid)
-                        child_procs = main_proc.children(recursive=True)
-                        for child_proc in child_procs:
-                            os.kill(child_proc.pid, signal.SIGKILL)
-                        os.kill(main_proc.pid, signal.SIGKILL)
-                    except (TypeError, psutil.Error):
-                        os.kill(pid, signal.SIGKILL)
+            if not psutil:
+                os.kill(pid, signal.SIGKILL)
```
matburt: I don't think this psutil check is even necessary anymore. In the old AWX codebase it was originally conditionally imported because it wasn't available in a good form on py2.6. This path can be removed entirely.

benthomasson:
We should leave the bwrap code for that separate issue.   I'd like to keep this PR clean of future-work.

Issue #100, opened 2018-07-27
Using --role ROLE and start runs the job in the foreground instead of background


Ladas:
Using ```--role ROLE``` and ```start``` runs the job in the foreground instead of background.

Using ```--playbook PLAYBOOK``` and ```start``` runs the job in the background, as expected.

Looks like these might be different code paths?

Ladas:
ping @matburt

matburt:
Yeah, this has turned into a major problem that we need to fix. We're tracking the code path unification here: https://github.com/ansible/ansible-runner/issues/98

PR #120, opened 2018-08-15
Fixes paths when using start

benthomasson:
Changes private_data_dir to be an absolute path.

Daemonized programs need to use absolute paths since their
working directory is the root directory.


benthomasson:
This should fix #100

PR #124, opened 2018-08-16
Allow user to specify the binpath for ansible or ansible-playbook

jctanner:
This will be handy for support, debugging and profiling scenarios where we need to quickly choose a different version or install path for ansible or even a wrapper.


jctanner:
Awesome, thanks for the merge @matburt

Issue #132, opened 2018-08-29
Incorrect PYTHONPATH set


mattclay:
The `PYTHONPATH` environment variable is set incorrectly:

https://github.com/ansible/ansible-runner/blob/7fa71c4f9615a632513829a1c1345ba14e978667/ansible_runner/runner_config.py#L131

If we have the following values:

- `PYTHONPATH` = `/some/dir`
- `callback_dir` = `/another/dir`

This will result in a `PYTHONPATH` of `/some/dir/another/dir:`

The code should probably be something like this instead:

```python
        python_path = self.env.get('PYTHONPATH', '')
        if python_path:
            python_path += ':'
        python_path += callback_dir + ':'
```

The presence of the trailing `:` may also be incorrect, as that includes the current working directory. Hopefully someone who knows ansible-runner better will know if that is needed or not.

PR #134, opened 2018-09-04
Fix an issue where PYTHONPATH and callback_dir can't be set together

matburt:
Related: #132

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -120,6 +120,9 @@ class RunnerConfig(object):
         if callback_dir is None:
             callback_dir = os.path.join(os.path.split(os.path.abspath(__file__))[0],
                                         "callbacks")
+        python_path = os.getenv('PYTHONPATH', '')
+        if python_path:
+            python_path += ":"
         self.env['ANSIBLE_CALLBACK_PLUGINS'] = callback_dir
         if 'AD_HOC_COMMAND_ID' in self.env:
             self.env['ANSIBLE_STDOUT_CALLBACK'] = 'minimal'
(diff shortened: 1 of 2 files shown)


ansible_runner/runner_config.py
```diff
@@ -128,7 +131,8 @@ def prepare(self):
         self.env['ANSIBLE_RETRY_FILES_ENABLED'] = 'False'
         self.env['ANSIBLE_HOST_KEY_CHECKING'] = 'False'
         self.env['AWX_ISOLATED_DATA_DIR'] = self.artifact_dir
-        self.env['PYTHONPATH'] = self.env.get('PYTHONPATH', '') + callback_dir + ':'
+
+        self.env['PYTHONPATH'] = python_path + callback_dir + ':'
```
benthomasson: How about?
```
self.env['PYTHONPATH'] = ":".join(filter(None, [python_path, callback_dir]))
```

test/unit/test_runner_config.py
```diff
@@ -300,6 +300,10 @@ def test_prepare():
     assert rc.env['AWX_ISOLATED_DATA_DIR'] == '/'
     assert rc.env['PYTHONPATH'] == '/:'
 
+    os.environ['PYTHONPATH'] = "/foo/bar"
+    rc.prepare()
+    assert rc.env['PYTHONPATH'] == "/foo/bar:/:"
```
benthomasson: Is the trailing `:` required?

ansible_runner/runner_config.py
matburt: meh? It's just two static elements

test/unit/test_runner_config.py
matburt: it's not *not* required, and can be helpful during future concatenations.

benthomasson:
mergeit

Issue #131, opened 2018-08-28
event handler callback isn't passed the runner object


pcuzner:
In my project I enable the dev to query events for a playbook - but this query currently stems from a listdir against the artifacts directory - which could be 'expensive'. 

I'd like to use the event_handler callback to populate a cache, so the cache of recent events could satisfy these queries. The problem is that the callback only passes the event data not the runner object's "ident".

Does it make sense to pass 'self' to the event_handler callback, so the callback function can get access to runner.config.ident? If so, I'm happy to raise a PR

andyfeller:
@matburt : in building my own ansible-runner plugin, I asked the same question as @pcuzner as I wanted to access stats information within the status_handler to detect what changes were made whenever runner completed.  Unless there is an objective security reason not to, I think it would open up greater options for those who are developing plugins to the same extent those using it as a Python module have available.

matburt:
What resources are you looking for? We *are* passing the config object (which gives access to the ident) to the status handler plugin.

https://github.com/ansible/ansible-runner/blob/master/ansible_runner/runner.py#L84

andyfeller:
hey @matburt!  so I'm definitely interested in accessing the stats within the `status_handler` whenever a final status _(`successful`, `canceled`, `failed`, `timeout`)_ is reached so I finalize reporting.

PR #135, opened 2018-09-04
Add runner ident to event data structure

matburt:
ref #131 

If a system is attempting to track multiple simultaneous jobs this
will make it easier to distinguish events and associate them with the
actual work, rather then the system needing to hold a context to the
Runner object itself


benthomasson:
mergeit

PR #140, opened 2018-09-24
add missing handling of status_handler-parameter

zauberpony:
Hello,

While playing around with ansible-runner, I noticed the `RunnerConfig` raises a TypeError *"__init__() got an unexpected keyword argument 'status_handler'"* when passing `status_handler` as advertised in `ansible_runner.run`'s documentation. 

I think my simple fix solves the problem — it did at least for my use case. I did not find any tests covering the *init_runner*, so I hope this PR is ok without tests. Otherwise I'll rework this PR and try to add a test…


matburt:
recheck

Issue #142, opened 2018-09-24
role-vars needs to accept data types other than string


trozet:
Currently role-vars are parsed as a bunch of key=value pairs where value is always interpreted to be a string. This doesn't work when a user wants to provide role-vars which have more complex data types like a list, dictionary, boolean, etc. The role-vars argument should also be able to take in these kinds of data types and set them the same way as role-vars would be read from a yaml vars file.

PR #143, opened 2018-09-24
Ability to consume other data types for rolevars

trozet:
Currently data read in with role-vars is interpreted as string values.
This patch modifies the behavior to try to interpret the value as a
literal and as a fallback assume it is a string.

closes: #142

Signed-off-by: Tim Rozet <trozet@redhat.com>


matburt:
recheck

Issue #154, opened 2018-10-12
ansible-runner is buffering output


jmontleon:
ansible-runner is buffering output. Inside containers I see task output coming in huge chunks, rather than as they occur. Often they never return and the container just hangs forever.

Preceding the ansible-runner command with  `export PYTHONUNBUFFERED=True` causes output to come normally.

We can obviously workaround this in our entrypoint script, but thought it might be beneficial to bring it up so that it can maybe be dealt with inside ansible-runner.

matburt:
@jmontleon can you give #155 a try and see if that works for you?

dch:
@matburt are you able to do a tagged release with #155 in it? I'm porting this to FreeBSD and it's a very useful fix, as well as the other commits since 1.1.1.

matburt:
Fixed in #155

matburt:
@dch I've got one more change I need to get in in order to make a new release: https://github.com/ansible/ansible-runner/pull/157

Issue #122, opened 2018-08-15
stdout isn't available for background jobs until the job is finished


matburt:
Also... if a job is cancelled then the stdout will never be available.

Job events are available however.

matburt:
Fixed in #155

PR #155, opened 2018-10-12
Fix output buffering by forcing a flush() on write() for stdout

matburt:
Ref: https://github.com/ansible/ansible-runner/issues/154 and https://github.com/ansible/ansible-runner/issues/122


wwitzel3:
This worked for me.

jmontleon:
The output still comes in bursts with just this change, but I haven't been able to make it freeze completely thus far, so it seems to be an improvement.

matburt:
I can add the unbuffered environment in.

matburt:
Give this one a try @jmontleon

ansible_runner/utils.py
```diff
@@ -240,12 +240,14 @@ def write(self, data):
                 if not self.suppress_ansible_output:
                     sys.stdout.write(stdout_actual + "\n")
```
jmontleon: This is what is still not being flushed. Adding a line after `sys.stdout.flush()` made things work as expected.

ansible_runner/runner_config.py
```diff
@@ -129,6 +129,7 @@ def prepare(self):
             self.env['ANSIBLE_STDOUT_CALLBACK'] = 'awx_display'
         self.env['ANSIBLE_RETRY_FILES_ENABLED'] = 'False'
         self.env['ANSIBLE_HOST_KEY_CHECKING'] = 'False'
+        self.env['PYTHONUNBUFFERED'] = 'True'
```
jmontleon: This didn't work either, which surprises me, but I don't confess to be a python expert.

matburt:
regate

matburt:
regate

matburt:
regate

ansible_runner/utils.py
```diff
@@ -239,13 +239,16 @@ def write(self, data):
             if stdout_actual and stdout_actual != "{}":
                 if not self.suppress_ansible_output:
                     sys.stdout.write(stdout_actual + "\n")
+                    self._handle.flush()
```
jmontleon: Why `self._handle.flush()` here and not `sys.stdout.flush()`? It does not seem to be working.

matburt: good point

matburt:
Give that a try and hopefully we'll be in good shape.

jmontleon:
@eriknelson this is working for me now. I don't know if you're in a position to test on the operator, but it might be good input to know if it's helping you as well.

matburt:
I'm going to trigger the gate here, let me know if it's still an issue and we can reopen

Issue #152, opened 2018-10-10
Ansible Runner doesn't respect `--roles-path` arg


dymurray:
# Version
```
ansible-runner (1.1.1)
ansible-runner-http (1.0.0)
```
Running ansible-runner with `--roles-path` set seems to do nothing:

```
[dymurray@pups bitcoin-sv-operator]$ ansible-runner -vv --role Bitd --roles-path /home/dymurray/go/src/github.com/dymurray/opsdk/bitcoin-sv-operator/roles/ --hosts localhost run demo        
ansible-playbook 2.6.5                                                                                                                                                                        
  config file = /etc/ansible/ansible.cfg                                                                                                                                                      
  configured module search path = [u'/home/dymurray/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']                                                                         
  ansible python module location = /usr/lib/python2.7/site-packages/ansible                                                                                                                   
  executable location = /usr/bin/ansible-playbook                                                                                                                                             
  python version = 2.7.15 (default, May 16 2018, 17:50:09) [GCC 8.1.1 20180502 (Red Hat 8.1.1-1)]                                                                                             
Using /etc/ansible/ansible.cfg as config file                                                                                                                                                 
                                                                                                                                                                                              
 [WARNING]: Unable to parse /home/dymurray/go/src/github.com/dymurray/opsdk                                                                                                                   
/bitcoin-sv-operator/demo/inventory as an inventory source                                                                                                                                    
                                                                                                                                                                                              
                                                                                                                                                                                              
 [WARNING]: No inventory was parsed, only implicit localhost is available                                                                                                                     
                                                                                                                                                                                              
                                                                                                                                                                                              
 [WARNING]: provided hosts list is empty, only localhost is available. Note                                                                                                                   
that the implicit localhost does not match 'all'                                                                                                                                              
                                                                              
                                                       
ERROR! the role 'Bitd' was not found in /home/dymurray/go/src/github.com/dymurray/opsdk/bitcoin-sv-operator/demo/project/roles:/home/dymurray/.ansible/roles:/usr/share/ansible/roles:/etc/an$
ible/roles:/home/dymurray/go/src/github.com/dymurray/opsdk/bitcoin-sv-operator/demo/project
```

When I run it with `ANSIBLE_ROLES_PATH` set to the same path everything works:
```
[...]

PR #157, opened 2018-10-12
Actually pass role path selection as part of main invocation

matburt:
related to: https://github.com/ansible/ansible-runner/issues/152


matburt:
recheck

PR #164, opened 2018-10-27
add process isolation support

### LINKED ISSUE #117: Implement bubblewrap for process isolation
The original AWX/Tower isolated execution code supported prefixing the command to run under bubblewrap. This was removed when `Runner` was created but this should go back in.

wwitzel3:
closes #117

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -217,6 +217,24 @@ def main(sys_args=None):
     parser.add_argument("-a", "--args", dest='module_args',
                         help="Module arguments")
 
+    parser.add_argument("--process-isolation", dest='process_isolation', default=False,
+                        help="Limits what directories on the filesystem the playbook run has access to, defaults to /tmp")
+
+    parser.add_argument("--process-isolation-executable", dest='process_isolation_executable', default="bwrap",
+                        help="Path that an isolated playbook run will use for staging. Defaults to /tmp")
+
+    parser.add_argument("--process-isolation-path", dest='process_isolation_path', default="/tmp",
+                        help="Path that an isolated playbook run will use for staging. Defaults to /tmp")
+
+    parser.add_argument("--process-isolation-hide-paths", dest='process_isolation_hide_paths',
+                        help="List of paths on the system that should be hidden from the playbook run")
+
+    parser.add_argument("--process-isolation-show-paths", dest='process_isolation_show_paths',
+                        help="List of paths on the system that should be exposed to the playbook run")
+
+    parser.add_argument("--process-isolation-venv", dest='process_isolation_venv',
+                        help="Path of a virtual environment to bind in to the isolated environment")
+
     args = parser.parse_args(sys_args)
 
     output.configure()
@@ -281,8 +299,13 @@ def main(sys_args=None):
                                    ignore_logging=False,
                                    json_mode=args.json,
                                    inventory=inventory_data,
-                                   roles_path=[args.roles_path] if args.roles_path else None)
-
+                                   roles_path=[args.roles_path] if args.roles_path else None,
+                                   process_isolation=args.process_isolation,
+                                   process_isolation_executable=args.process_isolation_executable,
+                                   process_isolation_path=args.process_isolation_path,
+                                   process_isolation_hide_paths=args.process_isolation_hide_paths,
+                                   process_isolation_show_paths=args.process_isolation_show_paths,
+                                   process_isolation_venv=args.process_isolation_venv)
                 if args.cmdline:
                     run_options['cmdline'] = args.cmdline
 
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -217,11 +217,11 @@ def main(sys_args=None):
     parser.add_argument("-a", "--args", dest='module_args',
                         help="Module arguments")
 
-    parser.add_argument("--process-isolation", dest='process_isolation', default=False,
+    parser.add_argument("--process-isolation", dest='process_isolation', action="store_true",
                         help="Limits what directories on the filesystem the playbook run has access to, defaults to /tmp")
 
     parser.add_argument("--process-isolation-executable", dest='process_isolation_executable', default="bwrap",
-                        help="Path that an isolated playbook run will use for staging. Defaults to /tmp")
+                        help="Process isolation executable that will be used. Defaults to bwrap")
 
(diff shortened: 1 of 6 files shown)


wwitzel3:
recheck

ansible_runner/__main__.py
```diff
@@ -217,6 +217,24 @@ def main(sys_args=None):
...
+    parser.add_argument("--process-isolation-path", dest='process_isolation_path', default="/tmp",
+                        help="Path that an isolated playbook run will use for staging. Defaults to /tmp")
+
+    parser.add_argument("--process-isolation-hide-paths", dest='process_isolation_hide_paths',
+                        help="List of paths on the system that should be hidden from the playbook run")
+
+    parser.add_argument("--process-isolation-show-paths", dest='process_isolation_show_paths',
+                        help="List of paths on the system that should be exposed to the playbook run")
+
+    parser.add_argument("--process-isolation-venv", dest='process_isolation_venv',
```
matburt: It looks like this only controls whether a read-only path is added to the bwrap. I think we should switch this to behaving like `show_paths` above and try not to encode any awx-specific logic (not referring to it as `venv`). Instead we can just modify AWX usage to attach it as a `ro` path and adjust the path to the default python executable (which normally comes in through the `--binary` or `RUNNER_BINARY` environment variable)

wwitzel3: Good point. I will update this.

wwitzel3:
I'll add docs and make those changes requested.

ansible_runner/runner_config.py
```diff
@@ -270,6 +281,60 @@ def generate_ansible_command(self):
...
+        '''
+        Wrap existing command line with proot to restrict access to:
+         - self.process_isolation_path (generally, /tmp) (except for own /tmp files)
+        '''
+        cwd = os.path.realpath(self.cwd)
+        pi_temp_dir = self.build_process_isolation_temp_dir()
+        new_args = [self.process_isolation_executable or 'bwrap', '--unshare-pid', '--dev-bind', '/', '/', '--proc', '/proc']
+
+        for path in sorted(set(self.process_isolation_hide_paths or [])):
+            if not os.path.exists(path):
```
chrismeyersfsu: Maybe a debug message letting the user know the path doesn't exist.

ansible_runner/runner_config.py
```diff
@@ -270,6 +281,60 @@ def generate_ansible_command(self):
...
+                continue
+            path = os.path.realpath(path)
```
chrismeyersfsu: Nice, resolve symlinks :+1:

ansible_runner/runner_config.py
```diff
@@ -270,6 +281,60 @@ def generate_ansible_command(self):
...
+            show_paths = [self.private_data_dir]
+        else:
+            show_paths = [cwd]
+
+        if self.process_isolation_venv:
+            new_args.extend(['--ro-bind', self.process_isolation_venv, self.process_isolation_venv])
+
+        show_paths.extend(self.process_isolation_show_paths or [])
+        for path in sorted(set(show_paths)):
+            if not os.path.exists(path):
```
chrismeyersfsu: Another debug message.

ansible_runner/runner_config.py
```diff
@@ -270,6 +281,60 @@ def generate_ansible_command(self):
...
+                continue
+            path = os.path.realpath(path)
```
chrismeyersfsu: :+1: I tested `os.path.exists` behavior and `os.path.realpath` behavior. Both resolve symlinks.

```
lrwxrwxrwx   1 meyers meyers        4 Nov 16 08:12 dir1 -> dir2
drwxr-xr-x   2 meyers meyers     4096 Nov 16 08:12 dir2
lrwxrwxrwx   1 meyers meyers        4 Nov 16 08:12 dir3 -> dir4
```

```
>>> import os
>>> os.path.exists('/tmp/dir1')
True
>>> os.path.exists('/tmp/dir3')
False
>>> os.path.realpath('/tmp/dir3/')
'/tmp/dir4'
```

ansible_runner/runner_config.py
```diff
@@ -270,6 +281,60 @@ def generate_ansible_command(self):
...
+            new_args.extend(['--bind', '%s' % (path,), '%s' % (path,)])
+
+        if 'ansible-playbook' in args:
```
chrismeyersfsu: Users can pass `binary` in i.e. `/usr/bin/ansible-playbook`. In which case, this match would fail.

ansible_runner/runner_config.py
```diff
@@ -270,6 +281,60 @@ def generate_ansible_command(self):
...
+        os.chmod(path, stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
+        return path
+
+    def wrap_args_with_process_isolation(self, args):
+        '''
+        Wrap existing command line with proot to restrict access to:
+         - self.process_isolation_path (generally, /tmp) (except for own /tmp files)
+        '''
+        cwd = os.path.realpath(self.cwd)
+        pi_temp_dir = self.build_process_isolation_temp_dir()
```
chrismeyersfsu: Does this whole dir get cleaned up?

wwitzel3: The contents get completely removed.

wwitzel3:
@matburt ready for another view, PTAL

wwitzel3:
@chrismeyersfsu I agree with your comment and this will be addressed in separate PR as we start to plumb things through awx.

ansible_runner/runner_config.py
wwitzel3: nice catch

ansible_runner/__main__.py
```diff
@@ -217,6 +217,24 @@ def main(sys_args=None):
     parser.add_argument("-a", "--args", dest='module_args',
                         help="Module arguments")
 
+    parser.add_argument("--process-isolation", dest='process_isolation', default=False,
```
matburt: one last thing that I hadn't though of before now... I wonder if we should switch this to `store_true` so that we don't end up requiring a positional argument after the `--process_isolation` flag.

Basically, passing `--process_isolation` on its own leads to PI mode being enabled rather than needing set pass the value T/F.

wwitzel3: Yeah, I agree with that. That's a better UX.

Issue #180, opened 2019-01-07
Add --tags and --skip-tags support to runner_config.py


mtucker502:
Add --tags and --skip-tags support to runner_config.py.

Currently this can be done only via injecting env/cmdline. This is not preferrable to systematically running specific tasks in a playbook since env/cmdline will have to be updated for each run instead of supplying as variable to ansible_runner.RunnerConfig().

matburt:
Implemented in #181

PR #181, opened 2019-01-07
Add --tags and --skip-tags support

mtucker502:
Add --tags and --skip-tags support to Runner_config()
to address issue #180

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -1,4 +1,4 @@
-#
+############################
 # Licensed to the Apache Software Foundation (ASF) under one
 # or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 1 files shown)


mtucker502:
@matburt can you rerun the integration tests? I think the test_run_command_job_timeout test is a bad test. I might look into changing it to something that runs longer.

matburt:
fwiw in the future if you need to re-run it you can add a comment that is just the text `recheck`

nim-odoo:
Hello,
It might be nice to add this to the docstring in https://github.com/ansible/ansible-runner/blob/e723a547b96b0aad47abd499f96088d69517032a/ansible_runner/interface.py#L134

PR #184, opened 2019-01-14
allow PYTHONPATH to be overwritten

chrismeyersfsu:
* Allow PYTHONPATH to be specified via envvars interface and append
runner environment.


mtucker502:
This can also be done by using the automatic `env/extravars` fileor `--extravars`. Just add the variable as follows:

`ansible_python_interpreter: "/opt/conda/bin/python"`

mtucker502:
recheck

PR #185, opened 2019-01-15
Adds jsonfile fact caching

benthomasson:
Adds fact caching to runner using the `jsonfile` plugin.  This adds setter and getter methods that read and write the JSON fact cache from the `Runner` instance.   This adds a new setting to `RunnerConfig` named 'fact_cache' which can be provided using `env/settings` or setting it on the `RunnerConfig` instance.

The default behavior is to create the fact cache in the `project/fact_cache` directory. The location of the directory can be changed using the `fact_cache` setting or argument to `RunnerConfig`.

### DIFF
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -297,3 +297,27 @@ class Runner(object):
                 os.kill(pid, signal.SIGKILL)
             except (OSError):
                 pass
+
+    def get_fact_cache(self, host):
+        '''
+        Get the entire fact cache only if the fact_cache_type is 'jsonfile'
+        '''
+        if self.config.fact_cache_type != 'jsonfile':
+            raise Exception('Unsupported fact cache type.  Only "jsonfile" is supported for reading and writing facts from ansible-runner')
+        fact_cache = os.path.join(self.config.fact_cache, host)
+        if os.path.exists(fact_cache):
+            with open(fact_cache) as f:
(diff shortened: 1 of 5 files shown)


ansible_runner/runner.py
```diff
@@ -297,3 +297,17 @@ def handle_termination(cls, pid, is_cancel=True):
                 os.kill(pid, signal.SIGKILL)
             except (OSError):
                 pass
+
+    def get_fact_cache(self, host):
+        fact_cache = os.path.join(self.config.cwd, self.config.fact_cache, host)
```
matburt: I think the fact cache should go in the artifacts dir?

matburt:
I think we were going to have callbacks for facts also? Is that doable here?

benthomasson:
> I think we were going to have callbacks for facts also? Is that doable here?

I was thinking about this too.  We would need a different plugin to enable callbacks.  The jsonfile plugin will not be sufficient.    I'll investigate that some more this week.

matburt:
I don't want to slow things down *too* much... we can always add these in later.

benthomasson:
I might want a way to disable the ansible-runner fact cache so I can use a different one and there can be only one fact cache plugin.  Should the ansible-runner fact cache be enabled by default?

matburt:
I think I *would* like this on by default, but allow it to be disabled/overridden

PR #188, opened 2019-01-23
fix bug in undocumented args file feature

chrismeyersfsu:
* args, much like env/cmdline, is expected to be whitespace separated
command-line args. However, internal to the code the expected return is
an array of args. This adds the conversion of string to array of args.

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -92,6 +92,7 @@ class RunnerConfig(object):
         self.verbosity = verbosity
         self.quiet = quiet
         self.suppress_ansible_output = suppress_ansible_output
+        self.has_raw_args = False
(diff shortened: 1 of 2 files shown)


ansible_runner/runner_config.py
```diff
@@ -220,7 +222,9 @@ def prepare_command(self):
         and if not calls :py:meth:`ansible_runner.runner_config.RunnerConfig.generate_ansible_command`
         """
         try:
-            self.command = self.loader.load_file('args', string_types)
+            cmdline_args = self.loader.load_file('args', string_types)
```
chrismeyersfsu: Or maybe we should do `self.command = self.loader.load_file('args', list)` and expect the `args` file to be a json array?

matburt:
regate

PR #189, opened 2019-01-23
add output_type parameter

chrismeyersfsu:
* Default is the existing event type output
* New output_type splits stdout on newlines and creates events out of
them


matburt:
`ansible_runner/runner.py:10:1: F401 'uuid' imported but unused`

chrismeyersfsu:
All that was needed of this PR was taken out and put in https://github.com/ansible/ansible-runner/pull/205 Turned out we don't need a new event filter. We just need to allow events w/ out a partial event to flow through the callback.

PR #203, opened 2019-02-11
Fix extravar loader behavior

### LINKED ISSUE #172: then env/extravars are overwritten
Take code from the  docs and add extraverts variable

```
import ansible_runner
extravars = {
    'ansible_connection': 'local',
    'test': 'extra'
}

r = ansible_runner.run(private_data_dir='/tmp/demo', playbook='test.yml', extravars=extravars)
print("{}: {}".format(r.status, r.rc))
# successful: 0
for each_host_event in r.events:
    print(each_host_event['event'])
print("Final status:")
print(r.stats)
```

take the demo directory from the repository, where `env/extravars` is

```
---
ansible_connection: local
test: val
```

run the ansible runner and check the `env/extravars`

```
{"test": "extra", "ansible_connection": "local"}
```

According the docs in the code **https://ansible-runner.readthedocs.io/en/latest/_modules/ansible_runner/interface.html#run**

```
 :param extravars: Extra variables to be passed to Ansible at runtime using ``-e``. Extra vars will also be
                      read from ``env/extravars`` in ``private_data_dir``.
```

This not true, they are not read at all and are overwritten.

matburt:
Our documentation makes it clear that extra vars passed in via the
programmatic interface are additive. In the presence of the
env/extravars file we should load that and then apply any extra vars
given in the parameterization afterwards.

Also, we were persisting some module-loaded data to disk which would
have the effect of writing the data out to files which is unexpected.

For a group of files (env/extravars, env/envvars, env/passwords,
env/settings, ssh_key, cmdline) if these files exist already then they
will not be overridden.

Fixes #172

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -197,10 +197,6 @@ class RunnerConfig(object):
             # Still need to pass default environment to pexpect
             self.env = os.environ.copy()
 
-        # extravars dict passed in via the interface API takes precedence over on-disk
-        if not self.extra_vars and self.loader.isfile('env/extravars'):
-            self.extra_vars = self.loader.abspath('env/extravars')
-
         try:
             self.settings = self.loader.load_file('env/settings', Mapping)
         except ConfigurationError:
@@ -271,6 +267,8 @@ class RunnerConfig(object):
             exec_list.append("--limit")
             exec_list.append(self.limit)
 
+        if self.loader.isfile('env/extravars'):
+            exec_list.extend(['-e', '@{}'.format(self.loader.abspath('env/extravars'))])
         if isinstance(self.extra_vars, dict) and self.extra_vars:
             exec_list.extend(
                 [
@@ -280,8 +278,6 @@ class RunnerConfig(object):
                     )
                 ]
             )
-        elif self.extra_vars:
-            exec_list.extend(['-e', '@%s' % self.extra_vars])
         if self.verbosity:
             v = 'v' * self.verbosity
             exec_list.append('-%s' % v)
--- a/test/unit/test_runner_config.py
+++ b/test/unit/test_runner_config.py
@@ -119,14 +119,6 @@ def test_prepare_env_extra_vars_defaults():
     assert rc.extra_vars is None
 
 
-def test_prepare_env_extra_vars():
-    rc = RunnerConfig('/')
-
-    with patch.object(rc.loader, 'isfile', side_effect=lambda x: True):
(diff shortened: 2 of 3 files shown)


matburt:
recheck

matburt:
recheck

matburt:
recheck

ansible_runner/runner_config.py
```diff
@@ -197,10 +197,6 @@ def prepare_env(self):
             # Still need to pass default environment to pexpect
             self.env = os.environ.copy()
 
-        # extravars dict passed in via the interface API takes precedence over on-disk
-        if not self.extra_vars and self.loader.isfile('env/extravars'):
-            self.extra_vars = self.loader.abspath('env/extravars')
```
chrismeyersfsu: Should still keep the self.extra_vars so if somewhere else in the runner code someone wants to reference it.

matburt: in this case we're just *not* using this to populate the file if it's not given as an argument.

test/unit/test_runner_config.py
```diff
@@ -189,9 +181,16 @@ def test_generate_ansible_command():
...
 
-    rc.extra_vars = '/env/extravars'
+    rc.extra_vars = dict(test="key")
     cmd = rc.generate_ansible_command()
-    assert cmd == ['ansible-playbook', '-i', '/inventory', '-e', '@/env/extravars', 'main.yaml']
+    assert cmd == ['ansible-playbook', '-i', '/inventory', '-e', '\'test="key"\'', 'main.yaml']
+
+    with patch.object(rc.loader, 'isfile', side_effect=lambda x: True):
+        cmd = rc.generate_ansible_command()
+        assert cmd == ['ansible-playbook', '-i', '/inventory', '-e', '@/env/extravars', '-e', '\'test="key"\'', 'main.yaml']
```
benthomasson: I like the use of `@file_name` for passing extra vars to ansible.   It might be nice to make another temporary file for passing the API provided extra vars too.

benthomasson: Is there a limit on the number of arguments that can be passed to ansible?  That might limit the number of extra vars that can be passed via the API.

matburt: When the user provides `self.extra_vars` we treat it as one big blob, but yeah... we could run into problems there. The idea if that's the case would be for the user to put that information *into* `env/extravars`

benthomasson: That makes sense.

PR #204, opened 2019-02-12
Fix up and expand inventory inputs

### LINKED ISSUE #179: No option to use multiple inventory files (and thus constructed inventories)
a singular parameter:

https://github.com/ansible/ansible-runner/blob/94f58c7c3667102dbafbf3b2349322c594deeccc/ansible_runner/runner_config.py#L249-L250

To be consistent with how Ansible core has operated since 2.4, it needs some way to process a list of inventory sources. The use case of constructed inventories need multiple `-i` flags in order to be usable. See the full documentation here:

https://docs.ansible.com/ansible/latest/plugins/inventory.html

A faster example here:

https://gist.github.com/halberom/e7791aa22c7dd1ef3f1ece3c815b4af8

```
ansible host5 -i hosts -i dynamic.config -m debug -a 'var=groups'
```

Any software that integrates Ansible via runner will be blocked from using any of these features unless runner changes its API to take multiple inventories somehow.

matburt:
* Fix an issue where we would always attempt to load inventory as a
  file if passed as an argument to Runner cli
* Add the ability to pass a list as a module argument to generate
  multiple -i ansible arguments
* Passing an empty list [] will disable passing the -i argument to
  ansible entirely

Fixes #179

### DIFF
--- a/ansible_runner/interface.py
+++ b/ansible_runner/interface.py
@@ -90,6 +90,7 @@ def run(**kwargs):
       - Path to the inventory file in the ``private_data_dir``
       - Native python dict supporting the YAML/json inventory structure
       - A text INI formatted string
+      - A list of inventory sources, or an empty list to disable passing inventory
     :param roles_path: Directory or list of directories to assign to ANSIBLE_ROLES_PATH
     :param envvars: Environment variables to be used when running Ansible. Environment variables will also be
                     read from ``env/envvars`` in ``private_data_dir``
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -281,11 +281,6 @@ def main(sys_args=None):
 
         with context:
             with role_manager(args) as args:
-                if args.inventory:
-                    with open(args.inventory) as f:
-                        inventory_data = f.read()
-                else:
-                    inventory_data = None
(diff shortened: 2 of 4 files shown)


ansible_runner/interface.py
```diff
@@ -90,6 +90,7 @@ def run(**kwargs):
       - Path to the inventory file in the ``private_data_dir``
       - Native python dict supporting the YAML/json inventory structure
       - A text INI formatted string
+      - A list of inventory sources, or an empty list to disable passing inventory
```
AlanCoding: I tried to read a lot of the docs to make sense of this.... so I think the bottom line is that this removes extra processing in the CLI interface with the result being that what you pass the the ansible-runner CLI as `--inventory` becomes the actual thing passed to Ansible via `-i`. There's also the additional detail of allowing use of lists, but this can't quite be utilized via the runner CLI, which is certainly fine.

matburt: You're right, most of the extra focus is on module interaction with the thought that *that* is usually the point where you want *more* flexibility beyond the `private_data_dir`.

In the future we can expand the functionality of the cli if that's needed.

benthomasson: If you pass `None` to inventory does it also disable passing inventory?

Issue #206, opened 2019-02-14
Fact cache doesn't respect artifact_dir param


chrismeyersfsu:
using the interface `ansible_runner.interface.run(artifact_dir='artifacts_explicit', ident=253)`
```bash-4.2# tree
.
├── artifacts
│   └── 253
│       └── fact_cache
│           └── HostChainLengthlhz
├── artifacts_explicit
│   └── artifacts
│       └── 253
│           ├── fact_cache
│           │   └── HostChainLengthlhz
│           ├── job_events
│           │   ├── 10-0242ac12-0005-ec2d-b714-000000000009.json
│           │   ├── 11-0a954587-d622-4108-b246-f861ce0a5001.json
│           │   ├── 12-0242ac12-0005-ec2d-b714-00000000000a.json
│           │   ├── 1-2b097bfb-fd8d-406d-9bcd-7cea8629ecd5.json
│           │   ├── 13-7b3b2e92-2937-42fb-855a-76678357133c.json
│           │   ├── 14-0242ac12-0005-ec2d-b714-00000000000b.json
│           │   ├── 15-211d4b09-6d8b-4cfa-a2bf-5505f8c84f04.json
│           │   ├── 16-0242ac12-0005-ec2d-b714-00000000000c.json
│           │   ├── 17-fb9b8e7d-d01e-4068-9e11-49d34ada56c5.json
│           │   ├── 18-0242ac12-0005-ec2d-b714-00000000000d.json
│           │   ├── 19-c5ad45cf-cf24-4d80-9966-ce234b079089.json
│           │   ├── 20-0242ac12-0005-ec2d-b714-00000000000e.json
│           │   ├── 21-7f6bcdbf-09bb-4513-80de-db735184929e.json
│           │   ├── 22-503ef002-2112-45ba-bb6d-e9953162f025.json
│           │   ├── 2-b3aa520b-6078-45e3-93cf-a619d885cc18.json
│           │   ├── 3-0bc550aa-bb52-488f-a3b5-341e8c64ace2.json
│           │   ├── 4-841cbc64-5dc9-4878-a9f2-b682ced8be03.json
│           │   ├── 5-0242ac12-0005-ec2d-b714-000000000006.json
│           │   ├── 6-0242ac12-0005-ec2d-b714-000000000012.json
│           │   ├── 7-280d22ab-2b75-43ec-a1f8-f3d6cbb75c07.json
[...]

PR #209, opened 2019-02-18
Converts a relative path in artifact_dir to an absolute one

benthomasson:
This might address #206 where the fact_cache was in a different location than expected.


chrismeyersfsu:
Does not fix my problem.

PR #210, opened 2019-02-19
Implement directory isolation during Runner ansible execution

### LINKED ISSUE #112: implement project directory execution isolation
The idea behind this is that when Runner starts execution it will copy the project directory and execute the runner invocation in that.

The issue that leads to this arises when multiple invocations of Runner are sharing the project directory... if they are managing the directory (such as checking out galaxy roles, etc) then they can end up stepping on each other's toes

matburt:
This helps prevent two simultaneously running tasks from clobbering
each other in the same working directory. This also allows any local
changes made by playbooks to be discarded preventing rogue playbooks
from making changes to the underlying automation.

This also adds an optional `project dir` parameter to allow localizing
the playbook content somewhere other than the private data dir.

Fixes #112

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -72,7 +72,10 @@ def role_manager(args):
         if args.artifact_dir:
             kwargs.artifact_dir = args.artifact_dir
 
-        project_path = os.path.join(args.private_data_dir, 'project')
+        if args.project_dir:
+            project_path = kwargs.project_dir = args.project_dir
+        else:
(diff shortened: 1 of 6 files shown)


chrismeyersfsu:
I can't seem to get the feature to work.

```
tree /tmp/my/                                                             /tmp/my/
├── isolated_dir
├── private_data_dir
└── project_dir
    └── main.yml
```

`ansible-runner --directory-isolation-base-path /tmp/my/isolated_dir --project-dir /tmp/my/project_dir -p main.yml run /tmp/my/private_data_dir`

PR #215, opened 2019-02-27
Add new playbook stats fields (ignored, rescued)

jladdjr:
Ports changes made in https://github.com/ansible/awx/pull/3272 to runner

New Ansible Playbook fields:
- https://github.com/ansible/ansible/pull/48418

Verified that test passes with both devel ansible and 2.7.8.

### DIFF
--- a/ansible_runner/loader.py
+++ b/ansible_runner/loader.py
@@ -166,7 +166,7 @@ class ArtifactLoader(object):
         except ConfigurationError as exc:
             debug(exc)
             raise
-        except UnicodeEncodeError as exc:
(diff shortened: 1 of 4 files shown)


ansible_runner/loader.py
```diff
@@ -166,7 +166,7 @@ def load_file(self, path, objtype=None, encoding='utf-8'):
         except ConfigurationError as exc:
             debug(exc)
             raise
-        except UnicodeEncodeError as exc:
```
jladdjr: There are a few flake8 fixes not related to playbook stats

PR #216, opened 2019-02-27
hide, show, ro all use realpath so should chdir

chrismeyersfsu:
* process isolation chdir directory should be resolved before pasing to
bwrap


chrismeyersfsu:
recheck

chrismeyersfsu:
recheck

PR #220, opened 2019-03-05
Debian/Ubuntu packaging

jbradberry:
Builds reasonable Debian/Ubuntu packages for ansible-runner
- ~~python-ansible-runner for Python 2~~ (**edit**: dropped for simplicity)
- ansible-runner for Python 3

Due to a hard dependency on pexpect >= 4.5, a backported deb package for python*-pexpect will be required for installation on any stock Debian 9, Ubuntu 16.04, or Ubuntu 18.04 system.  Similarly, Ansible >= 2.1 is required in order to run correctly, so a backported package or a pip installed version will be required to work on Ubuntu 16.04.


Makefile
```diff
@@ -28,11 +28,25 @@ ifeq ($(RPM_ARCH),)
     RPM_ARCH = $(shell uname -m)
 endif
 
+# DEB build variables
+DEB_DIST ?=
+DEB_ARCH ?= amd64
+DEB_NVR = $(NAME)_$(VERSION)-$(RELEASE)~$(DEB_DIST)
+DEB_NVRA = $(DEB_NVR)_$(DEB_ARCH)
+DEB_NVRS = $(DEB_NVR)_source
```
shanemcd: Is this variable ever used anywhere?

jbradberry: heh.  Probably not.

packaging/debian/control
```diff
@@ -0,0 +1,38 @@
...
+Standards-Version: 4.1.3
+Homepage: https://www.ansible.com
+X-Python-Version: >= 2.7
+X-Python3-Version: >= 3.5
+Vcs-Git: https://github.com/ansible/ansible-runner.git
+
+Package: python-ansible-runner
+Architecture: all
+Depends: ${python:Depends}, ${misc:Depends}
+Recommends: ansible (>= 2.1)
```
ghjm: Will Ansible versions back to 2.1 actually work with ansible-runner?  Maybe we should be looking for a newer Ansible version.

packaging/debian/control
```diff
@@ -0,0 +1,38 @@
...
+Priority: optional
+Maintainer: Ansible <info@ansible.com>
+Build-Depends: debhelper (>= 11), dh-python, python-all, python-setuptools, python3-all, python3-setuptools, python-pytest, python3-pytest, python-pexpect (>= 4.5), python3-pexpect (>= 4.5), python-mock, python3-mock, python-psutil, python3-psutil, python3-yaml
+Standards-Version: 4.1.3
+Homepage: https://www.ansible.com
+X-Python-Version: >= 2.7
+X-Python3-Version: >= 3.5
+Vcs-Git: https://github.com/ansible/ansible-runner.git
+
+Package: python-ansible-runner
```
ghjm: Will users be able to `apt-get install ansible-runner` and get the correct version for whatever Python they have installed?  Or will they have to `apt-get install python-ansible-runner`?  Would aesthetically prefer the former.

shanemcd: I'd also prefer this to be just `ansible-runner`, to match our RPM packaging. Runner is a more than just a python module, which is what usually follows this `python-` naming convention.

jbradberry: @ghjm @shanemcd the most straightforward way of doing so is to ignore Python 3 support.  Since 2.7 is still the default version for Debian 9 and both Ubuntu LTS versions we're talking about, I suppose that's not too big a deal.

shanemcd: For the upcoming changes to the RPM, I have created a metapackage called `ansible-runner`, that `Requires: python2-ansible-runner` on on EL7, and `Requires: python3-ansible-runner` on platforms where that is the default. I'm not sure if something like this would make sense for Debian.

wenottingham: *both* Ubuntu LTS?

jbradberry: @wenottingham 16.04 and 18.04 were specified.  14.04 is End of Life as of next month.

packaging/debian/changelog
```diff
@@ -0,0 +1,5 @@
+ansible-runner (1.2.0-1) unstable; urgency=low
+
+  * 1.2.0 release
```
matburt: This will probably go out on 1.3.0 first.

packaging/debian/control
matburt: I've only personally tested this with 2.5+ would be interested to see how this works going that far back.

packaging/debian/control
```diff
@@ -0,0 +1,38 @@
...
+Architecture: all
+Depends: ${python:Depends}, ${misc:Depends}
+Recommends: ansible (>= 2.1)
+Conflicts: python3-ansible-runner
+Description: interfaces with Ansible from other systems (Python 2)
```
matburt: This first line of the description seems weird and not a complete sentence... is this a short description?

jbradberry: Yeah.  The linting that the package builder does wants a _really_ short length.

jbradberry: https://lintian.debian.org/tags/description-too-long.html

packaging/debian/control
jbradberry: My "hello world"-style debug message example worked on Ansible 2.2.1.0, but failed on Ubuntu 16.04's Ansible version (2.0.0.2, if I remember correctly).  Runner is making use of a setting that was only made available starting in 2.1.

packaging/debian/control
jbradberry: There was also another complaint, about starting the short description with an article, of all things.  :moyai:

packaging/debian/control
jbradberry: @shanemcd Debian/Ubuntu has metapackages, but none of these targets are Python 3-first so creating one in this case seems pointless.  Additionally, it's easy on Debian-based systems to have multiple versions of Python installed at the same time, so I had thought that it made sense (and was certainly the direction that the build tools pushed me in) to have explicit per-Python packages like this so that the user can decide which makes the most sense for them.

With all that said, how strongly do we still feel that there should be an `ansible-runner` metapackage that only points to the Python 2 package?  cc @ghjm

packaging/debian/control
jbradberry: Do we have a set of versions that runner explicitly supports?

packaging/debian/control
matburt: Lets get @tvo318 to rewrite this into something better.

packaging/debian/changelog
jbradberry: I've bumped it now to 1.3.0.

packaging/debian/control
ghjm: I don't think we have an explicit list.  Leaving this at >= 2.1 is probably fine.

PR #225, opened 2019-03-19
add process isolation to settings file interface

wwitzel3:
Requested by @chrismeyersfsu for easier creation of runner instances without having to pass in long strings of command line arguments.

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -242,6 +242,13 @@ class RunnerConfig(object):
         self.job_timeout = self.settings.get('job_timeout', None)
         self.pexpect_timeout = self.settings.get('pexpect_timeout', 5)
 
+        self.process_isolation = self.settings.get('process_isolation', self.process_isolation)
+        self.process_isolation_executable = self.settings.get('process_isolation_executable', self.process_isolation_executable)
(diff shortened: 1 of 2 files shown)


ansible_runner/runner_config.py
```diff
@@ -242,6 +242,13 @@ def prepare_env(self):
         self.job_timeout = self.settings.get('job_timeout', None)
         self.pexpect_timeout = self.settings.get('pexpect_timeout', 5)
 
+        self.process_isolation = self.settings.get('process_isolation', False)
```
wwitzel3: The defaults in the documentation do no match the actual defaults we set in the code because those defaults are set by runner itself.

ansible_runner/runner_config.py
```diff
@@ -242,6 +242,13 @@ def prepare_env(self):
...
+        self.process_isolation_executable = self.settings.get('process_isolation_executable', None)
+        self.process_isolation_path = self.settings.get('process_isolation_path', None)
```
wwitzel3: We have to default this at runtime, see https://github.com/ansible/ansible-runner/blob/c578377e12ee7fb6f788e493c8a96faa8d922cb4/ansible_runner/runner_config.py#L372

This is why it is None here, but documented as `bwrap`

PR #230, opened 2019-03-20
[WIP] Add signing to deb builds

ghjm:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,18 +1,20 @@
 PYTHON ?= python
-ifeq ($(origin VIRTUAL_ENV),undefined)
-	DIST_PYTHON ?= pipenv run $(PYTHON)
+ifeq ($(origin VIRTUAL_ENV), undefined)
+    DIST_PYTHON ?= pipenv run $(PYTHON)
 else
-	DIST_PYTHON ?= $(PYTHON)
+    DIST_PYTHON ?= $(PYTHON)
 endif
 
 NAME = ansible-runner
 IMAGE_NAME ?= $(NAME)
 PIP_NAME = ansible_runner
-VERSION = $(shell $(DIST_PYTHON) setup.py --version)
+VERSION := $(shell $(DIST_PYTHON) setup.py --version)
 ifeq ($(OFFICIAL),yes)
     RELEASE ?= 1
 else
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,18 +1,20 @@
...
 else
-	DIST_PYTHON ?= $(PYTHON)
+    DIST_PYTHON ?= $(PYTHON)
 endif
 
 NAME = ansible-runner
 IMAGE_NAME ?= $(NAME)
 PIP_NAME = ansible_runner
-VERSION = $(shell $(DIST_PYTHON) setup.py --version)
+VERSION := $(shell $(DIST_PYTHON) setup.py --version)
```
ghjm: Using `:=` here lets us run pipenv once, instead of a dozen times.

Makefile
```diff
@@ -28,31 +30,49 @@ ifeq ($(RPM_ARCH),)
...
 	rm -rf dist
+	rm -rf ansible-runner.egg-info
 	rm -rf rpm-build
 	rm -rf deb-build
 
 dist:
 	$(DIST_PYTHON) setup.py bdist_wheel --universal
 
-sdist: dist/$(PIP_NAME)-$(VERSION).tar.gz
+sdist: dist/$(NAME)-$(VERSION).tar.gz
```
ghjm: It seems like setup.py is creating this file as `ansible-runner`, not `ansible_runner`, so we have to use `$(NAME)` here, not `$(PIP_NAME)`.  But if that's true, how did it ever work?  So there's clearly something I don't understand going on here.

jbradberry: My vague understanding of Make suggests that it doesn't care much what you call a rule, but if a file of the same name doesn't exist it'll always run instead of skipping when the artifact of that rule is up to date.

packaging/debian/changelog
```diff
@@ -1,5 +1,6 @@
-ansible-runner (1.2.0-1) unstable; urgency=low
+ansible-runner (%VERSION%-%RELEASE%~%DEB_DIST%) %DEB_DIST%; urgency=low
```
jbradberry: Huh, nice.

Makefile
ghjm: I think you're exactly right.  I think it was working before because the `PIP_NAME` version of the filename was just a rule name, that didn't actually match the files being generated, but who cares because Make will still chain to the next rule.

PR #238, opened 2019-03-27
fix a bug that causes ansible-runner to crash w/ playbooks w/ unicode

ryanpetrello:


matburt:
This looks good, and it's similar to what I did but set aside.

We need some tests around this. Specifically around calling `write(...)` that's sensitive to the handle provided to `OutputEventStream()`

Spredzy:
recheck

PR #244, opened 2019-04-02
Capture the command being executed to an artifact file

jbradberry:
So that programs that invoke ansible-runner can make use of the information.

### DIFF
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -15,7 +15,7 @@ import psutil
 
 import ansible_runner.plugins
 
-from .utils import OutputEventFilter, cleanup_artifact_dir
+from .utils import OutputEventFilter, cleanup_artifact_dir, build_safe_env
 from .exceptions import CallbackError, AnsibleRunnerException
 from ansible_runner.output import debug
 
@@ -102,7 +102,11 @@ class Runner(object):
 
         with codecs.open(command_filename, 'w', encoding='utf-8') as f:
             os.chmod(command_filename, stat.S_IRUSR | stat.S_IWUSR)
-            json.dump({'command': self.config.command, 'cwd': self.config.cwd}, f)
+            json.dump(
+                {'command': self.config.command,
+                 'cwd': self.config.cwd,
+                 'env': build_safe_env(self.config.env)}, f
+            )
 
         if self.config.ident is not None:
             cleanup_artifact_dir(os.path.join(self.config.artifact_dir, ".."), self.config.rotate_artifacts)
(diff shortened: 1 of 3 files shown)


ansible_runner/runner.py
```diff
@@ -99,6 +100,10 @@ def run(self):
                 raise
         os.close(os.open(stdout_filename, os.O_CREAT, stat.S_IRUSR | stat.S_IWUSR))
 
+        with codecs.open(command_filename, 'w', encoding='utf-8') as f:
+            os.chmod(command_filename, stat.S_IRUSR | stat.S_IWUSR)
+            json.dump({'command': self.config.command, 'cwd': self.config.cwd}, f)
```
chrismeyersfsu: I don't love stuffing a json in here. I'd rather this be two separate files.

test/integration/test_runner.py
```diff
@@ -26,6 +26,11 @@ def test_run_command(rc):
     status, exitcode = Runner(config=rc).run()
     assert status == 'successful'
     assert exitcode == 0
+    with open(os.path.join(rc.artifact_dir, 'command')) as f:
+        data = f.read()
```
ryanpetrello: Seems easier to `json.load()` ?

jbradberry: yeah, I wasn't confident that I'd be able to get a reasonable pattern out of the tmp directory

ansible_runner/utils.py
```diff
@@ -78,6 +78,27 @@ def check_isolation_executable_installed(isolation_executable):
...
+    '''
+    Build environment dictionary, hiding potentially sensitive information
+    such as passwords or keys.
+    '''
+    HIDDEN_PASSWORD = '**********'
+    hidden_re = re.compile(r'API|TOKEN|KEY|SECRET|PASS', re.I)
+    urlpass_re = re.compile(r'^.*?://[^:]+:(.*?)@.*?$')
+    safe_env = dict(env)
+    for k, v in safe_env.items():
+        if k == 'AWS_ACCESS_KEY_ID':
```
matburt: This is too specific to have in a generalized tool.

jbradberry: This function is ported over almost directly as is from AWX.  We _could_ drop the unredacted env into the artifact file and then clean it up on the other side, if we are ok with that.

jbradberry: I've gone ahead and done it that way now.  It seems like this isn't any worse than the existing situation, since we are dropping an envvars file with this stuff unredacted anyway.

PR #246, opened 2019-04-03
Collect performance data during playback run

jladdjr:


ansible_runner/display_callback/module.py
```diff
@@ -20,6 +20,8 @@
 # Python
 import collections
 import contextlib
+import json
```
ryanpetrello: Are these new imports necessary?

ansible_runner/runner.py
```diff
@@ -186,6 +229,10 @@ def run(self):
                 f.write(str(data))
         if self.config.directory_isolation_path and self.config.directory_isolation_cleanup:
             shutil.rmtree(self.config.directory_isolation_path)
+        if self.config.resource_profiling:
+            cmd = ['cgdelete', '-g', 'cpuacct,memory,pids:ansible-runner-{}'.format(self.config.ident)]
```
ryanpetrello: Do you want to use `resource_profiling_base_cgroup`, not hard-coded `ansible-runner-`?

ryanpetrello: Instead of making this string again, could you just reuse `cgroup_path` from above?

ansible_runner/runner.py
```diff
@@ -186,6 +229,10 @@ def run(self):
...
+            proc = subprocess.Popen(cmd, shell=True)
```
ryanpetrello: If you don't actually need stdout or care how this exited, using `subprocess.call` as above is simpler.

ansible_runner/runner_config.py
```diff
@@ -178,6 +180,18 @@ def prepare(self):
             self.env['ANSIBLE_STDOUT_CALLBACK'] = 'minimal'
         else:
             self.env['ANSIBLE_STDOUT_CALLBACK'] = 'awx_display'
+
+            if self.resource_profiling:
+                callback_whitelist = os.environ.get('ANSIBLE_CALLBACK_WHITELIST', '').strip()
+                self.env['ANSIBLE_CALLBACK_WHITELIST'] = callback_whitelist + (',' if len(callback_whitelist) else '') + 'cgroup_perf_recap'
```
ryanpetrello:
```python
callback_whitelist = ''
','.join(filter(None, [callback_whitelist, 'cgroup_perf_recap']))
callback_whitelist = 'foo'
','.join(filter(None, [callback_whitelist, 'cgroup_perf_recap']))
```

ansible_runner/runner_config.py
```diff
@@ -178,6 +180,18 @@ def prepare(self):
...
+                self.env['CGROUP_CONTROL_GROUP'] = '{}/{}'.format(self.resource_profiling_base_cgroup, self.ident)
+                self.env['CGROUP_OUTPUT_DIR'] = self.private_data_dir + '/profiling_data'
```
ryanpetrello: might be nice to wrap this in `os.path.normpath()` just in case an extra trailing `/` sneaks into `private_data_dir` somehow.

ansible_runner/runner_config.py
```diff
@@ -178,6 +180,18 @@ def prepare(self):
...
+                self.env['CGROUP_FILE_NAME_FORMAT'] = '%(task_uuid)s-%(feature)s.%(ext)s'
+                self.env['CGROUP_OUTPUT_FORMAT'] = 'json'
+                self.env['CGROUP_CPU_POLL_INTERVAL'] = '0.25'  # TODO
```
ryanpetrello: Do we intend to make this configurable on a first pass?

ansible_runner/runner_config.py
```diff
@@ -178,6 +180,18 @@ def prepare(self):
...
+                self.env['CGROUP_FILE_PER_TASK'] = 'True'
+                self.env['CGROUP_WRITE_FILES'] = 'True'
+                self.env['CGROUP_DISPLAY_RECAP'] = 'False'  # TODO
```
ryanpetrello: Does this just amount to an extra verbose event?  How does this materialize if you set it to `True`?

ansible_runner/runner_config.py
```diff
@@ -369,6 +386,15 @@ def build_process_isolation_temp_dir(self):
         os.chmod(path, stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
         return path
 
+    def wrap_args_with_cgexec(self, args):
+        '''
+        Wrap existing command line with cgexec in order to profile resource usage
+        '''
+        # TODO: Should only wrap if `self.execution_mode == ExecutionMode.ANSIBLE_PLAYBOOK` (see example below)
```
ryanpetrello: Does the callback plugin not work for adhoc commands?  Maybe we just don't care about those, considering their not likely critical from a performance tracking standpoint.

matburt:
fyi... you can run `tox` which will run flake8 and the unit and integration tests against python 2 and 3.

ansible_runner/runner.py
```diff
@@ -121,6 +155,15 @@ def run(self):
             if k != 'PATH' and isinstance(v, six.text_type):
                 self.config.env[k] = v.encode('utf-8')
 
+        # Prepare to collect performance data
+        if self.config.resource_profiling:
+            cgroup_path = '{0}/{1}'.format(self.config.resource_profiling_base_cgroup, self.config.ident)
+            cmd = 'cgcreate -g cpuacct,memory,pids:{}'.format(cgroup_path)
+            rc = subprocess.call(cmd, shell=True)
+            if rc:
+                # Unable to create cgroup, disable profiling
```
matburt: We should emit a `warn` logging statement here.

ansible_runner/runner_config.py
```diff
@@ -369,6 +386,15 @@ def build_process_isolation_temp_dir(self):
...
+        new_args = ['cgexec', '--sticky', '-g', 'cpuacct,memory,pids:{}/{}'.format(self.resource_profiling_base_cgroup, self.ident)]  # Is sticky needed?
```
matburt: I dunno... is sticky needed? 😄

setup.py
```diff
@@ -10,7 +10,7 @@
 
 setup(
     name="ansible-runner",
-    version="1.3.1",
+    version="1.5.0",  # .. so I can tell if SLOWYO installed correct version
```
matburt: This will almost certainly need to be removed before final acceptance.

matburt:
Some tests to validate some assertions would be good also.

ansible_runner/display_callback/module.py
jladdjr: No, think this was cruft from changes that used to be here. Removed.

ansible_runner/runner.py
```diff
@@ -36,6 +37,34 @@ def __init__(self, config, cancel_callback=None, remove_partials=True,
...
+                with open(path) as f:
+                    lines = []
+                    line = ''
+                    try:
+                        for line in f.readlines():
+                            debug('Added profiling datapoint')
+                            line = line[line.index('{'):line.index('}') + 1]  # remove cruft
+                            line = json.loads(line)
+                            lines.append(line)
+                    except:
```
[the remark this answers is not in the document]
jladdjr: Updated

ansible_runner/runner.py
```diff
@@ -57,6 +86,11 @@ def event_callback(self, event_data):
                     with codecs.open(partial_filename, 'r', encoding='utf-8') as read_file:
                         partial_event_data = json.load(read_file)
                     event_data.update(partial_event_data)
+
+                    event = event_data.get('event', '')
+                    if event == 'runner_on_ok' or event == 'runner_on_failed':
```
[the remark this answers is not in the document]
jladdjr: done

ansible_runner/runner.py
jladdjr: done

ansible_runner/runner_config.py
jladdjr: I _knew_ there was a better way - thanks!

ansible_runner/runner_config.py
jladdjr: I was thinking this might be nice to expose, too - will make the change

ansible_runner/runner_config.py
jladdjr: You get a summary at the end, like this:
Screenshot 2019-03-07 09 33 36

Seems like we would want to make this configureable, too - what do you think?

ansible_runner/runner_config.py
jladdjr: done

ansible_runner/runner_config.py
matburt: Is the summary part of the `playbook_on_stats` event or a verbose event at the end?

ansible_runner/runner_config.py
```diff
@@ -104,6 +105,9 @@ def __init__(self,
         self.process_isolation_hide_paths = process_isolation_hide_paths
         self.process_isolation_show_paths = process_isolation_show_paths
         self.process_isolation_ro_paths = process_isolation_ro_paths
+        self.resource_profiling = resource_profiling
```
matburt: These values also need to be able to come in from settings:

https://github.com/ansible/ansible-runner/blob/c70a02670882d94ae189a32b77a5f07df205dab9/ansible_runner/runner_config.py#L251

Since these values also can come in from the module interface they also need to be documented here:

https://github.com/ansible/ansible-runner/blob/c70a02670882d94ae189a32b77a5f07df205dab9/ansible_runner/interface.py#L116

(along with supporting `type` information)

ansible_runner/runner_config.py
jladdjr: @ryanpetrello - the plugin doesn't work with ad hoc commands. It only collects metrics when `v2_playbook_on_task_start` and `v2_playbook_on_stats` are called.

https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/callback/cgroup_perf_recap.py#L402-L405

ansible_runner/runner_config.py
jladdjr: Yep! This will ensure that any subprocesses that get spawned will run under the same cgroup as the parent.

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/resource_management_guide/starting_a_process

> You can also add the --sticky option before the command to keep any child processes in the same cgroup.

matburt: Then I think we can take that comment off.

jladdjr: Should be removed now

setup.py
jladdjr: Reverted to original version

ansible_runner/runner_config.py
jladdjr: The information you're seeing is just [printed](https://github.com/ansible/ansible/blob/f0480ab132bc2beb1feea42c7fcd714429114f38/lib/ansible/plugins/callback/cgroup_perf_recap.py#L414-L415) by the plugin:

image

If we wanted to collect this information from the plugin, we'd need to scrape / parse the information that gets printed. (The actual profiling data gets written, as json, to a set of files which we collect at the end of each task)

jladdjr:
> fyi... you can run `tox` which will run flake8 and the unit and integration tests against python 2 and 3.

Doh, didn't realize I was spamming the PR with Zuul checks - thanks, fixed the lint issues

ansible_runner/runner_config.py
matburt: If it gets printed (i.e... emitted to `stdout`) then we'll pick them up as `verbose` events. We should check to see if these are coming in as verbose events (would be good to know)

matburt:
Rally, it's fine to have the zuul checks in there. That's what it's there for.

packaging/rpm/ansible-runner.spec.j2
```diff
@@ -17,6 +17,7 @@ Requires: openssh-clients
 Requires: python-daemon
 Requires: python-pexpect
 Requires: python-psutil
+Requires: libcgroup-tools
```
matburt: Is there an Ubuntu analog?

ansible_runner/runner_config.py
ryanpetrello: 👍

matburt:
Lets also rewrite / squash these commits down before merge.

packaging/rpm/ansible-runner.spec.j2
ryanpetrello: Looks like `cgroup-bin`

pabelanger:
recheck

jladdjr:
We've rescoped the perf stats feature recently and it's been a while since this branch was touched. Will be opening a fresh PR.

Issue #250, opened 2019-04-05
Runner exits without artifacts if `cmdline` is not on the path


ryanpetrello:
When `ansible-playbook` (or `bwrap`, is using process isolation) isn't on the `PATH`, runner doesn't provide any clues in its artifacts (and some critical ones are missing).

```shell
~ which ansible-playbook
/usr/bin/which: no ansible-playbook in ...
~/dev/ansible-runner ansible-runner start ./tmpdir -p `pwd`/hello_world.yml  -i testing123
wc ./tmpdir/artifacts/testing123/stdout
0 0 0 ./tmpdir/artifacts/testing123/stdout
```

Notice that there are no `rc`, `status`, or `job_events` artifacts _at all_:
```
ls ./tmpdir/artifacts/testing123
command  stdout
```

PR #251, opened 2019-04-05
fail more gracefully when ansible-playbook or bwrap aren't in PATH

ryanpetrello:
addresses https://github.com/ansible/ansible-runner/issues/250

### DIFF
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -134,18 +134,37 @@ class Runner(object):
         }
 
         self.status_callback('running')
-        child = pexpect.spawn(
-            self.config.command[0],
-            self.config.command[1:],
-            cwd=self.config.cwd,
-            env=env,
-            ignore_sighup=True,
-            encoding='utf-8',
-            echo=False,
-            use_poll=self.config.pexpect_use_poll,
-        )
-        child.logfile_read = stdout_handle
         self.last_stdout_update = time.time()
+        try:
+            child = pexpect.spawn(
+                self.config.command[0],
+                self.config.command[1:],
+                cwd=self.config.cwd,
+                env=env,
+                ignore_sighup=True,
+                encoding='utf-8',
+                echo=False,
+                use_poll=self.config.pexpect_use_poll,
+            )
+            child.logfile_read = stdout_handle
+        except pexpect.exceptions.ExceptionPexpect as e:
+            child = collections.namedtuple(
+                'MissingProcess', 'exitstatus isalive'
+            )(
+                exitstatus=127,
+                isalive=lambda: False
+            )
+
+            def _decode(x):
+                return x.decode('utf-8') if six.PY2 else x
+
+            # create the events directory (the callback plugin won't run, so it
+            # won't get created)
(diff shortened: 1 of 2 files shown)


ansible_runner/runner.py
```diff
@@ -44,11 +44,11 @@ def event_callback(self, event_data):
...
         if 'uuid' in event_data:
             filename = '{}-partial.json'.format(event_data['uuid'])
-            partial_filename = os.path.join(self.config.artifact_dir,
-                                            'job_events',
-                                            filename)
-            full_filename = os.path.join(self.config.artifact_dir,
-                                         'job_events',
+            events_directory = os.path.join(self.config.artifact_dir, 'job_events')
+            if not os.path.exists(events_directory):
+                os.mkdir(events_directory, 0o700)
```
ryanpetrello: If `ansible-playbook` doesn't actually _run_, then stdout verbose events don't recorded because this directory doesn't get created (the callback plugin has code that does this directory creation later).

ansible_runner/runner.py
```diff
@@ -134,18 +134,29 @@ def run(self):
...
+                cwd=self.config.cwd,
+                env=env,
+                ignore_sighup=True,
+                encoding='utf-8',
+                echo=False,
+                use_poll=self.config.pexpect_use_poll,
+            )
+            child.logfile_read = stdout_handle
+        except pexpect.exceptions.ExceptionPexpect as e:
+            child = collections.namedtuple(
```
ryanpetrello: `pexpect` looks for `self.config.command[0]` on the system path, and if it's not, found this error is thrown (and the fork fails).

ryanpetrello: This is what things like after my patch:

```shell
[root@isolated 47]# pwd
/tmp/awx_47_17x1e6qx/artifacts/47
[root@isolated 47]# ls -la
total 28
drwx------. 3 root root 4096 Apr  5 02:04 .
drwx------. 3 root root 4096 Apr  5 02:04 ..
-rw-------. 1 root root 2662 Apr  5 02:04 command
drwx------. 2 root root 4096 Apr  5 02:04 job_events
-rw-------. 1 root root    3 Apr  5 02:04 rc
-rw-------. 1 root root    6 Apr  5 02:04 status
-rw-------. 1 root root   58 Apr  5 02:04 stdout
[root@isolated 47]# cat stdout
The command was not found or was not executable: bwrap.
[root@isolated 47]# cat rc
127
[root@isolated 47]# cat status
failed
[root@isolated 47]# cat job_events/* | python -m json.tool
{
    "counter": 1,
    "end_line": 1,
    "event": "verbose",
    "runner_ident": "47",
    "start_line": 0,
    "stdout": "The command was not found or was not executable: bwrap",
    "uuid": "eedfa39c-cd08-408f-bf76-0629c5d0295d"
}
```

Here's what it looks like in AWX:

ryanpetrello: image

ryanpetrello: With process isolation disabled:

image

PR #258, opened 2019-04-10
Use command that will wait longer to trigger the timeout tests

matburt:
Depends-on: #255

PR #261, opened 2019-04-11
Fix up timeout and cancel tests to be less flaky

matburt:
Depends on #258


matburt:
I merged this into #258

PR #264, opened 2019-04-12
Don't build sdist using tox

pabelanger:
We are able to do this now, as we have a job that will validate we can
build an sdist / wheel. We also set usedevelop so we can modify the
local source tree without doing pip install again.


pabelanger:
recheck

PR #266, opened 2019-04-15
fix a variety of py2 unicode handling bugs in artifacts/command serialization

ryanpetrello:


matburt:
regate

ansible_runner/runner.py
```diff
@@ -137,8 +138,8 @@ def run(self):
         self.last_stdout_update = time.time()
         try:
             child = pexpect.spawn(
-                self.config.command[0],
-                self.config.command[1:],
+                command[0],
```
jbradberry: Are we sure that the thing with unicode characters in the environment causing `spawn()` to hang doesn't also apply to the command?

PR #268, opened 2019-04-15
fix a bug in handling env/cmdline that contains unicode

ryanpetrello:


ansible_runner/loader.py
```diff
@@ -161,15 +161,15 @@ def load_file(self, path, objtype=None, encoding='utf-8'):
...
-            parsed_data = contents.encode(encoding)
+            contents = parsed_data = self.get_contents(path)
         except ConfigurationError as exc:
             debug(exc)
             raise
         except UnicodeEncodeError:
             raise ConfigurationError('unable to encode file contents')
 
         if objtype is not string_types:
+            parsed_data = contents.encode(encoding)
```
ryanpetrello: We only need to bother with this encoding for deserialization e.g., `load_json` and `load_yaml`, *not* simple strings like `env/cmdline`

PR #269, opened 2019-04-16
WIP: embed base64 encoded event data in the pty itself

ryanpetrello:
this removes the `-partial.json` IPC mechanism for passing event data
from ansible to the runner process

### DIFF
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -22,8 +22,8 @@ from ansible_runner.output import debug
 
 class Runner(object):
 
-    def __init__(self, config, cancel_callback=None, remove_partials=True,
-                 event_handler=None, finished_callback=None, status_handler=None):
+    def __init__(self, config, cancel_callback=None, event_handler=None,
+                 finished_callback=None, status_handler=None):
         self.config = config
         self.cancel_callback = cancel_callback
         self.event_handler = event_handler
@@ -34,7 +34,6 @@ class Runner(object):
         self.errored = False
         self.status = "unstarted"
         self.rc = None
-        self.remove_partials = remove_partials
 
     def event_callback(self, event_data):
         '''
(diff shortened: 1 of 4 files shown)


ryanpetrello:
This doesn't pass tests yet, but I think it should work.

ansible_runner/utils.py
```diff
@@ -326,6 +306,19 @@ def _emit_event(self, buffered_stdout, next_event_data=None):
             event_data['end_line'] = self._start_line + n_lines
             self._start_line += n_lines
             if self._event_callback:
+                if event_data.get('stdout'):
```
matburt: Even *if* this PR doesn't work out for replacing partials we should probably still do this. Trying to emit stdout down at the callback plugin gives us all kinds of grief.

ansible_runner/runner.py
```diff
@@ -22,8 +22,8 @@
 
 class Runner(object):
 
-    def __init__(self, config, cancel_callback=None, remove_partials=True,
-                 event_handler=None, finished_callback=None, status_handler=None):
+    def __init__(self, config, cancel_callback=None, event_handler=None,
```
matburt: Partial deletion control is also a parameterization of the module that we'll need to remove also.

ryanpetrello:
I tested this playbook:

```yaml
- name: run setup
  hosts: all
  tasks:
    - setup:
        gather_subset: all
```

...with 10 hosts and `--forks=10`, and the stdout and number of generated job events matched before _and_ after this PR.

The only concern I have with this PR is one that @matburt expressed - if there's any potential for parallel processes to interleave this stdout.

ansible_runner/runner.py
ryanpetrello: Where?

ansible_runner/utils.py
ryanpetrello: Agreed, this is a much saner place to handle `sys.stdout`

ansible_runner/runner.py
matburt: hmm, I guess I'm wrong about this one.

ryanpetrello:

PR #270, opened 2019-04-16
Revert "fix a bug in handling env/cmdline that contains unicode"

ryanpetrello:
This reverts commit e97434cd42fd9d9db1a241bf594a1d0edd68a84a.


ryanpetrello:
This breaks SSH key loading, and needs to be fixed a different way.

PR #272, opened 2019-04-17
Even more unicode fixes

ryanpetrello:


ansible_runner/runner_config.py
```diff
@@ -306,8 +306,8 @@ def generate_ansible_command(self):
         exec_list = [base_command]
 
         try:
-            cmdline_args = self.loader.load_file('env/cmdline', string_types)
-            args = shlex.split(cmdline_args.decode('utf-8'))
+            cmdline_args = self.loader.load_file('env/cmdline', string_types, encoding=None)
```
ryanpetrello: The way this worked before is kind of silly - for these `string_types`, we're just basically doing `.encode().decode()`

Issue #239, opened 2019-03-27
Improve callback plugin testing


ryanpetrello:
The callback plugin could use more tests.  These tests existed in AWX for some time, and it would be great to port them over:

https://github.com/ansible/awx/blob/e6abd77c963fe7325ef1aaa45d50f51246e871f6/awx/lib/tests/test_display_callback.py

AlanCoding:
done

PR #275, opened 2019-04-17
Port old callback tests to integration folder

AlanCoding:
Connect https://github.com/ansible/ansible-runner/issues/239

Here is a diff of what changed:

https://gist.github.com/AlanCoding/6efaed386ee65b7ac92f117178323f18/revisions

### DIFF
--- a/test/integration/test_display_callback.py
+++ b/test/integration/test_display_callback.py
@@ -1,0 +1,301 @@
+from __future__ import absolute_import
+
+import json
+import os
+import yaml
+import six
+
+from ansible import __version__ as ANSIBLE_VERSION
+
+from ansible_runner.interface import init_runner
+
+import pytest
+
+
+@pytest.fixture()
+def executor(tmpdir, request):
+    private_data_dir = six.text_type(tmpdir.mkdir('foo'))
+
+    playbooks = request.node.callspec.params.get('playbook')
+    playbook = list(playbooks.values())[0]
+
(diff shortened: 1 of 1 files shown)


AlanCoding:
@ryanpetrello looking at the python 2 results, I'm starting to wonder if there might actually be impact from https://github.com/ansible/ansible-runner/pull/272. I wonder if that might even fix the failures.

test/integration/test_display_callback.py
```diff
@@ -0,0 +1,308 @@
...
+    - pw: SENSITIVE
+  tasks:
+    - uri:
+        url: https://example.org
+        user: john-jacob-jingleheimer-schmidt
+        password: "{{ pw }}"
+'''},  # noqa
+])
+def test_module_level_no_log(executor, playbook):
+    # https://github.com/ansible/tower/issues/1101
```
ryanpetrello: We should scrub this URL.

test/integration/test_display_callback.py
```diff
@@ -0,0 +1,307 @@
...
+
+import pytest
+
+
+@pytest.fixture()
+def executor(tmpdir, request):
+    private_data_dir = six.text_type(tmpdir.mkdir('foo'))
+    print('private data dir')
+    print(private_data_dir)
+    print(type(private_data_dir))
```
AlanCoding: whoops

Issue #182, opened 2019-01-09
Duplicate output


Nani-o:
Using the latest 1.2.0 version, I have some problems trying to use ansible-runner.

I can only reproduce this problem on my Ubuntu server, on my OS X machine this works fine.

python : 3
ansible-runner : 1.2.0
ansible : 2.7.5

Given this simple playbook :
```YAML
---
- hosts: localhost
  gather_facts: no
  tasks:
    - debug:
        msg: ok
    - debug:
        msg: ko
```

I expect to see this (output from ansible-playbook command) : 
```SHELL
$ ansible-playbook toto.yml
 
PLAY [localhost] ***************************************************************************************************************************************

TASK [debug] *******************************************************************************************************************************************
ok: [localhost] => {
    "msg": "ok"
}

TASK [debug] *******************************************************************************************************************************************
ok: [localhost] => {
    "msg": "ko"
}

PLAY RECAP *********************************************************************************************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=0

```

Using ansible-runner this is what I get using it as a library : 
```SHELL
$ python3 toto.py

PLAY [localhost] ***************************************************************
ok: [localhost] => {
    "msg": "ok"
}


TASK [debug] *******************************************************************
ok: [localhost] => {
    "msg": "ok"
}
ok: [localhost] => {
    "msg": "ko"
}


TASK [debug] *******************************************************************
ok: [localhost] => {
    "msg": "ko"
}

PLAY RECAP *********************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=0
```

And using it as a command line : 
```SHELL
$ ansible-runner run . -p toto.yml
 
PLAY [localhost] ***************************************************************

TASK [debug] *******************************************************************
ok: [localhost] => {
    "msg": "ok"
}

TASK [debug] *******************************************************************
ok: [localhost] => {
    "msg": "ko"
}
localhost                  : ok=2    changed=0    unreachable=0    failed=0



PLAY RECAP *********************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=0

```

When doing the same from my mac, everything works fine (python3 also on my mac).

Nani-o:
I don't fully understand everything in the source code, but I came to the conclusion that removing [thoses lines](https://github.com/ansible/ansible-runner/blob/4a5a1d79104e3b14168b2d6aef1642f07a053ce0/ansible_runner/utils.py#L247-L251) solves my problem.

[...]

PR #283, opened 2019-04-24
write event stdout in the event emit callback

ryanpetrello:


ryanpetrello:
attempts to address: https://github.com/ansible/ansible-runner/issues/182

ansible_runner/utils.py
```diff
@@ -326,6 +306,19 @@ def _emit_event(self, buffered_stdout, next_event_data=None):
             event_data['end_line'] = self._start_line + n_lines
             self._start_line += n_lines
             if self._event_callback:
+                if event_data.get('stdout'):
+                    if self.output_json:
+                        stdout_actual = json.dumps(event_data)
```
AlanCoding: In the case that `output_json` is true, do you need to skip the case of `stdout_actual != "{}"`? It seems that that's a bit of logic that may have been missed. But then again, events with blank lines are extremely common, and I have no idea what the expectations could possibly be for those with that setting.

matburt:
recheck

pabelanger:
recheck

pabelanger:
recheck tarball jobs fixed

pabelanger:
recheck

pabelanger:
recheck

matburt:
recheck

ryanpetrello:

PR #284, opened 2019-04-24
bundle six.ensure_str to avoid pinning to a new six version

ryanpetrello:


Spredzy:
Just to add. The version required here is not available in el7. `1.9` is.

PR #290, opened 2019-04-27
Use the correct python executable for tests

toabctl:
Instead of using "python" which might point to python2, use the python
that is used to run the tests itself (which is sys.executable).

This fixes a build problem on openSUSE where ansible-runner is only
build for python3. Due to that, no "python" executable is there.


matburt:
regate

matburt:
recheck

PR #291, opened 2019-04-29
Add changed count to stats

dominikholler:
The successfully changed events per host count is added to
the stats. This is useful to check idempotency.


matburt:
What we should probably *actually* do with these instead is call `.get(val, [])`

If you can make this change I'll merge it.

dominikholler:
> What we should probably *actually* do with these instead is call `.get(val, [])`
>

Like this: `changed=last_event.get('changed',{})`?


> If you can make this change I'll merge it.
>

Thank you very much for your attention!

PR #297, opened 2019-05-09
Added --display-callback and scrubber

john-westcott-iv:
**Purpose**
This PR allows the specification of a display callback to be used for output via the cmd line parameter --display-callback.

In addition, it adds a scrubber display callback. This callback attempts to mask strings in the output from Ansible with '******'.

**Scrubber details**
It will mask any values from the following environment variables:
- ANSIBLE_NET_AUTHORIZE
- ANSIBLE_NET_AUTH_PASS
- ANSIBLE_NET_PASSWORD
- ANSIBLE_NET_SSH_KEYFILE
- AWS_ACCESS_KEY_ID
- AWS_SECURITY_TOKEN
- AZURE_PASSWORD
- AZURE_SECRET
- CLOUDSCALE_API_TOKEN
- DO_API_KEY
- DO_API_TOKEN
- DO_OAUTH_TOKEN
- F5_PASSWORD
- HEROKU_API_KEY
- MERAKI_KEY
- NETSCALER_NITRO_PASS
- OAUTH_TOKEN
- ONLINE_API_KEY
- ONLINE_API_TOKEN
- ONLINE_OAUTH_TOKEN
- ONLINE_TOKEN
- SCW_API_KEY
- SCW_API_TOKEN
- SCW_OAUTH_TOKEN
- SCW_TOKEN
- TF_VAR_HEROKU_API_KEY
- VDIRECT_PASSWORD
- VMWARE_PASSWORD

Additional environment variables can be added to the list of the environment variables via the ansible variable ar_scrub_additional_env in the env/extravars. For example, if you had a custom credential type from Tower that created a env variable MY_CUSTOM_PASS you could capture the value of MY_CUSTOM_PASS with this in env/extravars:
```
ar_scrub_additional_env:
  - MY_CUSTOM_PASS
```

Any strings in the array ar_scrub_terms in env/extravars will also be scrubbed. For example, if you never wanted to see the term "root" in the output you could set this in env/extravars:
```
ar_scrub_terms:
  - "root"
```

Any strings listed in ar_scrub_additional_vars in env/extravars will be searched for as variables in env/extravars and, if found, their value will be added to the list of items to be scrubbed. For example with this env/extravars:
```
ar_scrub_additional_vars:
  - "my_var"
my_var: "Something"
```
Any instance of the term "Something" in the output will be masked.

Any value in env/passwords file will be added to the list to scrub.

Finally, if it finds an add_host task and there is an ansible_ssh_password in the output of add host that string will be added to the items to scrub.

**Limitations**
The callback does not know about vaulted passwords (other than the vault password itself) and it does not do variable expansion inside the variables. i.e. if env/extravars contains:
```
ar_scrb_additional_vars:
  - my_var
my_var: "{{ another_var }}"
```
The scrubber call back would end up looking for the string "{{ another_var }}".

Attached is a .tar.gz of a working directory with some example usage.
[work_dir.tar.gz](https://github.com/ansible/ansible-runner/files/3163756/work_dir.tar.gz)

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -40,6 +40,7 @@ from ansible_runner import output
 from ansible_runner.utils import dump_artifact, Bunch
 from ansible_runner.runner import Runner
 from ansible_runner.exceptions import AnsibleRunnerException
+from ansible_runner import display_callback
 
 VERSION = pkg_resources.require("ansible_runner")[0].version
 
@@ -255,6 +256,13 @@ def main(sys_args=None):
     parser.add_argument("--limit",
                         help="Matches ansible's ``--limit`` parameter to further constrain the inventory to be used")
 
+    valid_callbacks = []
+    for cls_name in display_callback.__all__:
+        valid_callbacks.append(getattr(display_callback, cls_name).CALLBACK_NAME)
+    parser.add_argument('--display-callback', dest='display_callback',
+                        choices=valid_callbacks,
+                        help="Specify a display callback to use. Defaults to minimal for ad-hoc or default for anything else")
+
     args = parser.parse_args(sys_args)
 
     output.configure()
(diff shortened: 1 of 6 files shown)


matburt:
I think this is probably not something we'll want to go with for a few reasons:

* The cost of doing event post-processing is just too high, even if it's optional
* Hard coded lists of filtered environment variables just doesn't scale well with development.

Generally we would want these things solved on the core side, or the downstream Runner consumer.

PR #298, opened 2019-05-10
Implicit load as opposed to path modification

AlanCoding:


matburt:
recheck

matburt:
recheck

AlanCoding:
There's another cleaner more elegant approach we can take for this, and I'm closing this with the idea of looking into that.

PR #304, opened 2019-05-27
Keep the callback provided in the environment variable

pguermo:
When the environment variable `ANSIBLE_STDOUT_CALLBACK` is set, it was ignored when calling ansible as only the callback_dir was passed.

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -214,9 +214,6 @@ def main(sys_args=None):
     parser.add_argument("-j", "--json", action="store_true",
                         help="Output the json event structure to stdout instead of Ansible output")
 
-    parser.add_argument("-k", "--ask-pass", action="store_true",
-                        help="Ask for connection password")
-
     parser.add_argument("-v", action="count",
                         help="Increase the verbosity with multiple v's (up to 5) of the ansible-playbook output")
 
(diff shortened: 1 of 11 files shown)


matburt:
I think it's probably worth adding some tests here to specifically verify that this won't affect Runner's internal callback.

pguermo:
I've started to investigate to add a test to ensuire there is no regression with the ansible_runner callback but I'm facing the following problem: I need to define somewere an additional callback but in the test I need to know the absolute path to the folder containing the defintion of the callback as I expected to have the callback folder next to the test script.

I were thinking to use the parametrization of the arguments of the test function. But I don't find any way to provide the absolute path to the test script. 

```
@pytest.mark.parametrize('envvars', [
     {pytest.param('ANSIBLE_CALLBACK_PLUGINS', '???/test/integration/callback')}
])
```

Any idea ?

matburt:
You can see here that the roles kept as subdirectories of the integration test directory are respected for relative paths, will that not work for you?

https://github.com/ansible/ansible-runner/blob/ca7a1b92a47513e8ac447797c64f1c9ad7a13988/test/integration/test_main.py#L106

pguermo:
I've added an other callback "other_callback" in test/integration/callback.
I've added the parametrized fixtures "envvars" to the test test_display_callback::test_callback_plugin_receives_events to use or not this additional callback.

matburt:
I don't know what you've done but you've pushed a bunch of old commits into your branch... it looks like you didn't pull-rebase correctly? You should close this PR and open a new one with just your changes.

Issue #305, opened 2019-05-28
--ask-pass is not handled


pguermo:
ansible-runner --help shows that --ask-pass should be an option, but it is not supported by RunnerConfig

pguermo:
Use --cmdline='--ask-pass' to start ansible with this option

PR #306, opened 2019-05-28
Add support for parameter --ask-pass

pguermo:
ansible-runner --help shows that --ask-pass should be an option, but it is not supported by RunnerConfig

Fix for bug: https://github.com/ansible/ansible-runner/issues/305

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -214,9 +214,6 @@ def main(sys_args=None):
     parser.add_argument("-j", "--json", action="store_true",
                         help="Output the json event structure to stdout instead of Ansible output")
 
-    parser.add_argument("-k", "--ask-pass", action="store_true",
-                        help="Ask for connection password")
-
     parser.add_argument("-v", action="count",
                         help="Increase the verbosity with multiple v's (up to 5) of the ansible-playbook output")
 
@@ -315,7 +312,6 @@ def main(sys_args=None):
                                    json_mode=args.json,
                                    inventory=args.inventory,
                                    forks=args.forks,
-                                   ask_pass=args.ask_pass,
                                    project_dir=args.project_dir,
                                    roles_path=[args.roles_path] if args.roles_path else None,
                                    process_isolation=args.process_isolation,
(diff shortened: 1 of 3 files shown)


matburt:
Prompting at the pre-run level isn't something I think we want to support. Runner is intended to be given the password via the run(...) interface or via the private data directory.

It might be that this needs to be added if there are passwords provided via the private data dir OR with credentials on the `run(...)` interface.

pguermo:
Indeed, I don't expect ansible-runner prompts for the passwords but ansible must prompt for the password. (When launching ansible manually, I started it with the option --ask-pass and provide the requested password.)

If ansible prompts for the passwords, ansible runner may provide them if they are defined in env/passwords. That's the purpose of the fix.

matburt:
I think the better implementation here would be to detect whether a password prompt is needed by checking to see if `env/passwords` exists and has content. If it does then emit the `-k` and/or `-K` (sudo password prompt).

There's a whole other side to this I think that is vaulted password prompts.

pguermo:
The fix has been reworked. The option --ask-pass is not used anymore at ansible-runner level but the option -k is passed to ansible when the file env/passwords contains passwords.

matburt:
I think this might need to be expanded, there are two other password prompts that could be emitted as part of this chain: `-K` which is prompting for `BECOME` passwords. and `--ask-vault-pass` which is used if vault password is prompted for. The challenge here is the format of the password prompt file, see this part of the documentation:

https://ansible-runner.readthedocs.io/en/latest/intro.html#env-passwords

What I would *really* like to do (and allude to that in the comment there in the docs) is change this away from the regex matching and move towards stable matching on known ansible prompts, which I think is achievable since Ansible seems to have standardized on it.

as-is, your PR assumes (possibly incorrectly) that the password is a normal user password... but what happens if the input is only for become password (necessitating `-K` and *not* `-k`)?

pguermo:
Indeed, the case of the -K or --ask-vault-pass was not covered by my change. Thanks to broaden my vision.

Passing --ask-pass/-k, --ask-become-pass/-K and --ask-vault-pass to ansible-runner would give you more control. Ansible-runner could then forward those parameters to ansible. 

I think the list of parameters used by ansible will be more stable than relying on the output when requesting some type of password.

matburt:
You may need to rebase your branch with master

pguermo:
I noticed I were implementing again the option cmdline.
The options -ask-pass/-k, --ask-become-pass/-K and --ask-vault-pass may be used to start ansible using the option --cmdline.

The pull request may be closed.

Issue #310, opened 2019-06-05
atexit()-based directory cleanup insufficient when runner is called via library interface


wenottingham:
Related to https://github.com/ansible/awx/issues/4010 - when runner is called as a python library, atexit() won't necessarily run at a point that it should.

matburt:
I think the solution to this is to remove the `atexit()` call and use the system we use for directory isolation instead:

https://github.com/ansible/ansible-runner/blob/8691932005d78329383efe59357a397c7078807a/ansible_runner/runner.py#L221-L222

ryanpetrello:
@matburt I don't think this is quite solved yet, seeing a lot of these types of errors in AWX integration:

![image](https://user-images.githubusercontent.com/214912/59108077-182f2f80-8908-11e9-8335-f197ac4deb6f.png)

PR #319, opened 2019-06-28
release underlying fd when asked

chrismeyersfsu:
* Process isolation cleanup was failing because runner was holding on to
the stdout file descriptor. Runner, correctly, trys to close the file
descriptor (that is wrapped in the OutputEventFilter), but alas, that
method is NOOP. This commit closes the underlying file descriptor as
well as flushes the underlying file descriptor when asked.

related to https://github.com/ansible/awx/issues/4010 and https://github.com/ansible/ansible-runner/issues/310

### DIFF
--- a/ansible_runner/utils.py
+++ b/ansible_runner/utils.py
@@ -225,11 +225,7 @@ class OutputEventFilter(object):
         self.suppress_ansible_output = suppress_ansible_output
 
     def flush(self):
-        # pexpect wants to flush the file it writes to, but we're not
-        # actually capturing stdout to a raw file; we're just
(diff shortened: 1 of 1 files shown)


ansible_runner/utils.py
```diff
@@ -229,7 +229,7 @@ def flush(self):
         # actually capturing stdout to a raw file; we're just
         # implementing a custom `write` method to discover and emit events from
         # the stdout stream
-        pass
+        self._handle.flush()
```
matburt: If we're *actually* flushing then lets remove the comment above this.

PR #328, opened 2019-07-29
Fix event generation on run_async

irosenzw:
During the playbook execution, Runner watches
the events directory for new events and return
(yield) them.
Runner stop watching the directory when the
playbook finishes.

The old logic worked only on a simple run -
first run the playbook and then (after it finished) 
yield all its events, while on the async run it yielded only
the 4-5 first events.

Signed-off-by: Ido Rosenzwig <irosenzw@redhat.com>

### DIFF
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -1,5 +1,4 @@
 import os
-import re
 import stat
 import time
 import json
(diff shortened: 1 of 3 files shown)


matburt:
This looks good, I wonder if you could add some test coverage for this?

irosenzw:
@matburt, I added the test coverage that you requested, please review.

Issue #330, opened 2019-08-14
Errors thrown serializing vault data


AlanCoding:
See https://github.com/ansible/awx/issues/4478

reproduced locally with runner

```
tree alan/
alan/
├── env
│   ├── cmdline
│   ├── extravars
│   └── passwords
├── inventory
│   └── hosts
└── project
    └── test.yml

3 directories, 5 files
```

```
(runner) arominge-OSX:ansible-runner alancoding$ cat alan/env/cmdline 
--ask-vault-pass
(runner) arominge-OSX:ansible-runner alancoding$ cat alan/env/passwords 
---
"password:\\s*?$": "password"
(runner) arominge-OSX:ansible-runner alancoding$ cat alan/project/test.yml 
# this secret was produced by the command
# ansible-vault encrypt_string --ask-vault-pass "hell workld111"
# given password is "password"
# there is no id for this vault secret
#
# run this locally:
# ansible-playbook -i localhost, set_vault_var.yml --ask-vault-pass --connection=local
- hosts: localhost
  gather_facts: no
  connection: local
  tasks:
  - set_fact:
      password: !vault |
          $ANSIBLE_VAULT;1.1;AES256
          34623238666437386233333863363534313333363763376335343464373337376665626433376662
          6431326264333965383565386132656461663131373261630a666538323133656564306634653137
          30616365366466353332313238626634646235376231326538346635646437306638396162613762
          3638336436633037330a393332306632646236626231316134366431343239303762376631653837
          3362

  - debug: msg="My password value is {{ password }}"
```

```
ansible-runner --playbook test.yml run alan
Vault password: 

PLAY [localhost] ***************************************************************

TASK [set_fact] ****************************************************************
 [WARNING]: Failure using method (v2_runner_on_ok) in callback plugin
(<ansible.plugins.callback.awx_display.CallbackModule object at 0x1093890b8>):
Object of type 'AnsibleVaultEncryptedUnicode' is not JSON serializable


TASK [debug] *******************************************************************
ok: [localhost] => {
    "msg": "My password value is hell workld111"
}

PLAY RECAP *********************************************************************
localhost                  : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   


```

AlanCoding:
Expected to be resolved by https://github.com/ansible/ansible-runner/pull/331

PR #331, opened 2019-08-14
Use Ansible encoder for dumping event data

AlanCoding:
Connect https://github.com/ansible/ansible-runner/issues/330

See: https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/callback/json.py#L38

I think we can import Ansible here???

```
ansible-runner --playbook test.yml run alan 
Vault password: 

PLAY [localhost] ***************************************************************

TASK [set_fact] ****************************************************************
ok: [localhost]

TASK [debug] *******************************************************************
ok: [localhost] => {
    "msg": "My password value is hell workld111"
}

localhost                  : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   


```

### DIFF
--- a/ansible_runner/display_callback/events.py
+++ b/ansible_runner/display_callback/events.py
@@ -43,6 +43,8 @@ class AnsibleJSONEncoderLocal(json.JSONEncoder):
     def default(self, o):
         if getattr(o, 'yaml_tag', None) == '!vault':
             return o.data
+        elif isinstance(o, (datetime.date, datetime.datetime)):
+            return o.isoformat()
         return super(AnsibleJSONEncoderLocal, self).default(o)
 
 
--- a/ansible_runner/display_callback/events.py
+++ b/ansible_runner/display_callback/events.py
@@ -28,6 +28,9 @@ import stat
 import threading
 import uuid
 
+# Ansible
+from ansible.parsing.ajson import AnsibleJSONEncoder
+
 __all__ = ['event_context']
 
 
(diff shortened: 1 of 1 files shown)


ansible_runner/display_callback/events.py
```diff
@@ -28,6 +28,9 @@
 import threading
 import uuid
 
+# Ansible
+from ansible.parsing.ajson import AnsibleJSONEncoder
```
matburt: This is heading into some dangerous territory as Runner doesn't import Ansible directly right now. How portable is this? Do we *need* to do this?

AlanCoding: This bug throws errors in the callback plugin, I would say it's a must-fix.

The alternative is to copy over @ryanpetrello's

https://github.com/ansible/awx/blob/devel/awx/main/queue.py#L23

We could switch AWX to import runner's version of the encoder, eventually. That could move us in the direction of making ansible-runner the de-facto python library for Ansible variable encoder / decoders. I don't necessarily think that's a bad idea either.

Keep in mind, AWX already has _other_ encoders and decoders for unsafe variables. We may have yet un-discovered bugs for other types.

matburt: I think I'd definitely like to pull in *that* old encoder if it gets us there. I'd like to keep from importing Ansible *at all* if there is literally any other way around an issue like this.

AlanCoding: here? https://github.com/ansible/ansible-runner/blob/master/ansible_runner/utils.py

AlanCoding: No, wait, I have even more questions. If I put this in utils, it still wouldn't be a simple import. In cases where these files import from other parts of ansible_runner, it doesn't import from the ansible_runner module.

https://github.com/ansible/ansible-runner/blob/master/ansible_runner/display_callback/module.py#L33

There's a lot of baggage with how the imports are done. I don't necessarily feel more confident importing from another file (wherever that may be) in ansible_runner. So I guess the option on the table is to just define the class right in that file.

matburt: If we think we might expand on this in the future, which seems likely, then I feel like we should possibly create a `ansible_runner/encoders.py`

AlanCoding:
working with new version

```
ansible-runner --playbook test.yml run alan 
Vault password: 

PLAY [localhost] ***************************************************************

TASK [set_fact] ****************************************************************
ok: [localhost]

TASK [debug] *******************************************************************
ok: [localhost] => {
    "msg": "My password value is hell workld111"
}

localhost                  : ok=2    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   

```

AlanCoding:
Handling of unsafe vars is left out of this, but I realize that may need addressing if

https://github.com/ansible/ansible/pull/60602

is merged.

Issue #29, opened 2018-04-29
How to debug?


cdvv7788:
I am running the demo on this project. Cloned the repo, copied it to /tmp and ran the following commands:

![image](https://user-images.githubusercontent.com/5531776/39412022-ff8e4a80-4bda-11e8-9810-41f717e3d1da.png)
Nothing outputs.
r.status is 'failed' but there is no clue about what is failing.

matburt:
What do you see in the artifacts directory under the listed in the private data dir?

cdvv7788:
I see the following:
![image](https://user-images.githubusercontent.com/5531776/39437247-e82ffd12-4c65-11e8-8f77-e886ed868101.png)
Inside each folder, there are 3 files:
rc: 250
status: failed
stdout:

matburt:
I know it's been a couple of weeks, I'm sorry it's taken me a while to get back to you... can you try this with the latest HEAD?

cdvv7788:
Thanks for checking this. After using the latest version, I get the following:
![image](https://user-images.githubusercontent.com/5531776/40189524-49b19cc6-59c2-11e8-8aff-5a7a246768bf.png)

matburt:
Fixed in #34

PR #256, opened 2019-04-09
Port AWX changes to stream verbose events

AlanCoding:
I did not write integration tests for this, although you can find my manual testing at:

https://github.com/AlanCoding/ansible-runner-examples/tree/master/non_playbook/sleep_with_writes

I did test this fix with that example, and it produces:

```
$ ansible-runner run ~/Documents/repos/ansible-runner-examples/non_playbook/sleep_with_writes/
hi world
goodby world
```

(5 second delay between the two)

This is desirable, because it obtains (as far as the eye can see) the same as:

```
$ ~/Documents/repos/ansible-runner-examples/non_playbook/sleep_with_writes/sleep_and_write.sh 
hi world
goodby world
```

PR #335, opened 2019-08-22
[utils.py] Fix currupt json stdout with large events

NickLaMuro:
When using the `--json` output with `ansible-runner`, it is possible to over fill `pexpect`'s `maxread` default of 2000, which ends up causing the `else` of the `while` loop in the `OutputEventFilter.write` method to get triggered, and that will indiscriminately push raw output to the stdout handle.  This output is also a partial event at this point as well, and shouldn't be pushed to STDOUT until a full event has been processed and breaks the expectation of 1 line of JSON per line.

The intent of this section of the `else` is to provide some feedback to the user if something goes wrong (for example, you provide a playbook path that does not exist), as shown in the issue and pull request (fix) where they originate from:

- https://github.com/ansible/ansible-runner/issues/29
- https://github.com/ansible/ansible-runner/pull/34

This tries to keep with the spirit of those fixes, but does so after the main spawned process loop has completed, and not while still in progress.

### DIFF
--- a/test/integration/test_display_callback.py
+++ b/test/integration/test_display_callback.py
@@ -338,3 +338,46 @@ def test_output_when_given_invalid_playbook(private_data_dir):
     stdout = executor.stdout.read()
     assert "ERROR! the playbook:" in stdout
     assert "could not be found" in stdout
+
+
+def test_output_when_given_non_playbook_script(private_data_dir):
+    # As shown in the following pull request:
+    #
+    #   https://github.com/ansible/ansible-runner/pull/256
+    #
+    # This ports some functionality that previously lived in awx and allows raw
+    # lines of stdout to be treated as event lines.
+    #
+    # As mentioned in the pull request as well, there were no specs added, and
+    # this is a retro-active test based on the sample repo provided in the PR:
+    #
+    #   https://github.com/AlanCoding/ansible-runner-examples/tree/master/non_playbook/sleep_with_writes
+
+    with open(os.path.join(private_data_dir, "args"), 'w') as args_file:
+        args_file.write("bash sleep_and_write.sh\n")
+    with open(os.path.join(private_data_dir, "sleep_and_write.sh"), 'w') as script_file:
+        script_file.write("echo 'hi world'\nsleep 0.5\necho 'goodbye world'\n")
+
+    # Update the settings to make this test a bit faster :)
+    os.mkdir(os.path.join(private_data_dir, "env"))
+    with open(os.path.join(private_data_dir, "env", "settings"), 'w') as settings_file:
+        settings_file.write("pexpect_timeout: 0.2")
+
+    executor = init_runner(
+        private_data_dir=private_data_dir,
+        inventory="localhost ansible_connection=local",
+        envvars={"ANSIBLE_DEPRECATION_WARNINGS": "False"}
+    )
+
+    executor.run()
+    stdout = executor.stdout.readlines()
+    assert stdout[0].strip() == "hi world"
+    assert stdout[1].strip() == "goodbye world"
+
+    events = list(executor.events)
+
+    assert len(events) == 2
+    assert events[0]['event'] == 'verbose'
+    assert events[0]['stdout'] == 'hi world'
+    assert events[1]['event'] == 'verbose'
+    assert events[1]['stdout'] == 'goodbye world'
--- a/test/integration/test_display_callback.py
+++ b/test/integration/test_display_callback.py
@@ -25,12 +25,15 @@ def executor(request, private_data_dir):
     playbook = list(playbooks.values())[0]
     envvars = request.node.callspec.params.get('envvars')
     envvars = envvars.update({"ANSIBLE_DEPRECATION_WARNINGS": "False"}) if envvars is not None else {"ANSIBLE_DEPRECATION_WARNINGS": "False"}
+    json_mode = request.node.callspec.params.get('json_mode')
+    json_mode = json_mode if json_mode is not None else False
 
     r = init_runner(
         private_data_dir=private_data_dir,
         inventory="localhost ansible_connection=local",
         envvars=envvars,
-        playbook=yaml.safe_load(playbook)
+        playbook=yaml.safe_load(playbook),
+        json_mode=json_mode
     )
 
     return r
@@ -381,3 +384,28 @@ def test_output_when_given_non_playbook_script(private_data_dir):
     assert events[0]['stdout'] == 'hi world'
     assert events[1]['event'] == 'verbose'
     assert events[1]['stdout'] == 'goodbye world'
+
+
+@pytest.mark.parametrize('json_mode', [True])
+@pytest.mark.parametrize('playbook', [
+{'listvars.yml': '''
+- name: List Variables
+  connection: local
+  hosts: all
+  tasks:
+    - name: Display all variables/facts known for a host
+      debug:
+        var: hostvars[inventory_hostname]
+'''},  # noqa
+])
+def test_large_stdout_parsing_when_using_json_output(executor, playbook, json_mode):
+    # When the json flag is used, it is possible to output more data than
+    # pexpect's maxread default of 2000 characters.  As a result, if not
+    # handled properly, the stdout can end up being corrupted with partial
+    # non-event matches with raw "non-json" lines being intermixed with json
+    # ones.
+    #
+    # This tests to confirm we don't polute the stdout output with non-json
+    # lines when a single event has a lot of output.
+    executor.run()
+    assert all(line[0] == "{" for line in executor.stdout.readlines())
--- a/test/integration/test_display_callback.py
+++ b/test/integration/test_display_callback.py
@@ -15,9 +15,12 @@ HERE = os.path.abspath(os.path.dirname(__file__))
 
 
 @pytest.fixture()
-def executor(tmpdir, request):
-    private_data_dir = six.text_type(tmpdir.mkdir('foo'))
+def private_data_dir(tmpdir):
+    return six.text_type(tmpdir.mkdir('foo'))
 
+
+@pytest.fixture()
+def executor(request, private_data_dir):
     playbooks = request.node.callspec.params.get('playbook')
     playbook = list(playbooks.values())[0]
     envvars = request.node.callspec.params.get('envvars')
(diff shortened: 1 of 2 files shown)


matburt:
Looks like some CI failures ^^

NickLaMuro:
So the one of the test failures is from one I added, but it seems to be OS/`ansible` dependent.

As far as I can tell right now, on a fedora-29 Vagrant VM (same as CI), the test I added in the first commit is printing the error data to `stderr`, and the spec is trying to read from the `stdout` file.  This is not the behavior of this test on OSX where the error is printed to `stdout` as expected when I wrote it.

Still trying to find what is controlling the `stdout`/`stderr` decision making when printing between `pexpect`/`ansible`/`ansible-runner`/`omglolwtfbffbbq`... taking a bit...

NickLaMuro:
Okay, a `git push --force` later and I think I have figured out what the issue was, though not what I thought above with it being `stderr`/`stdout`... the update with that is `pty.h` that the Python `pty` module uses will always combine the `stdout`/`stderr` streams, so what I thought it was originally above was actually a red herring...

The real issue was still OS dependent, but was rather related the code that was left in the `else:` of the `while` that was added by @AlanCoding in this PR:

https://github.com/ansible/ansible-runner/pull/256

On Fedora29, it seems like an extra `\n` was being added to the output that cause the out to be processed twice in the `if data and '\n' in data ...` block.  In the previous form of this code (which using `flush_remaining_buffer`, it meant the stdout line would be dropped from the buffer when it was turned into an `event` with that code.  I will try to push a separate branch that had the old implementation in place for comparison.

By changing up the last commit in this PR to instead not create a new method, but print lines as part of the `if` block that creates events, it means we are only printing new data for lines that should be considered events.  If we are in the middle of processing an event, that block isn't entered because there is a separate check for that.

Hopefully this will fix the tests on the CI server as well.

NickLaMuro:
Okay, well, I just pushed what I think should fix the tests for py27, but I also have my old solution pushed up here for comparison.:

https://github.com/NickLaMuro/ansible-runner/commit/d56b64e6ae2f134036c3f8a748189bbd5b4d2e51

matburt:
Looking good, do you think this is ready?

NickLaMuro:
@matburt I think so.  However, since I am messing with some previous bug fixes that have been opened by the core team, I wouldn't mind if someone wants to take the time before merging to properly confirm everything that was fixed previously is still working (instead of just relying on the test suite).

Up to you, but since I have a integration tests that covers my use case that was failing previously, there isn't much on my end I can think of left to do.

matburt:
I think I'm going to freeze this one until after the 1.4 release and we can see about getting it into a 1.4.X

NickLaMuro:
> I think I'm going to freeze this one until after the 1.4 release and we can see about getting it into a 1.4.X

Fine for me.  I don't have a deadline, but I think it is a common enough use case and isn't handled properly in the `stdout` output that it **should** be fixed.  However, **when** that happens is definitely flexible and I don't think it needs to be bumped up in priority at this point.

👍 from me.

test/integration/test_display_callback.py
```diff
@@ -311,3 +317,95 @@ def test_module_level_no_log(executor, playbook):
...
+        playbook=os.path.join(private_data_dir, 'fake_playbook.yml')
+    )
+
+    executor.run()
+    stdout = executor.stdout.read()
+    assert "ERROR! the playbook:" in stdout
+    assert "could not be found" in stdout
+
+
+def test_output_when_given_non_playbook_script(private_data_dir):
```
AlanCoding, line 346: I tried running these tests locally, and both

 - test_output_when_given_invalid_playbook
 - test_output_when_given_non_playbook_script

seem to be passing with current `devel`.

I think that makes sense because https://github.com/ansible/ansible-runner/issues/29 was closed, and PR 34 was merged.

Given what you're saying here, I think these 2 tests are fine for the purpose of adding regression coverage. However, I haven't processed the other code changes you're making here, so I can't say merge it now.

test/integration/test_display_callback.py
```diff
@@ -311,3 +317,95 @@ def test_module_level_no_log(executor, playbook):
...
+@pytest.mark.parametrize('json_mode', [True])
+@pytest.mark.parametrize('playbook', [
+{'listvars.yml': '''
+- name: List Variables
+  connection: local
+  hosts: all
+  tasks:
+    - name: Display all variables/facts known for a host
+      debug:
+        var: hostvars[inventory_hostname]
```
AlanCoding, line 398: My preference for test writing here would be that we don't gather facts, and we don't rely on printing something from Ansible itself, which can change. It's possible to construct a string in Ansible itself of the desired length.

I ran this test and confirmed it failed, the JSON data got printed up to 4976 lines, totaling 18.8 k characters. I don't know what the interaction there is with pexpect's max 2k characters.

test/integration/test_display_callback.py
```diff
@@ -311,3 +317,95 @@ def test_module_level_no_log(executor, playbook):
...
+    # When the json flag is used, it is possible to output more data than
+    # pexpect's maxread default of 2000 characters.  As a result, if not
+    # handled properly, the stdout can end up being corrupted with partial
+    # non-event matches with raw "non-json" lines being intermixed with json
+    # ones.
+    #
+    # This tests to confirm we don't polute the stdout output with non-json
+    # lines when a single event has a lot of output.
+    executor.run()
+    assert all(line[0] == "{" for line in executor.stdout.readlines())
```
AlanCoding, line 411: I'm also somewhat unclear what this is looking for. Like, it's printing a dictionary, but only the first line needs to start with "{"

AlanCoding:
Slightly revised test:

```python
@pytest.mark.parametrize('playbook', [
{'listvars.yml': '''
- name: List Variables
  connection: local
  hosts: localhost
  gather_facts: false
  tasks:
    - name: Print a lot of lines
      debug:
        msg: "{{ ('F' * 150) | list }}"
'''},  # noqa
])
def test_large_stdout_parsing_when_using_json_output(executor, playbook):
    # When the json flag is used, it is possible to output more data than
    # pexpect's maxread default of 2000 characters.  As a result, if not
    # handled properly, the stdout can end up being corrupted with partial
    # non-event matches with raw "non-json" lines being intermixed with json
    # ones.
    #
    # This tests to confirm we don't polute the stdout output with non-json
    # lines when a single event has a lot of output.
    executor.config.env['ANSIBLE_NOCOLOR'] = str(True)
    executor.run()
    text = executor.stdout.read()
    assert text.count('"F"') == 150
```

This is very fast, and surprisingly it seems to be able to hit the bug.

```
>       assert text.count('"F"') == 150
E       assert 235 == 150
E        +  where 235 = <built-in method count of str object at 0x7f941b883800>('"F"')
E        +    where <built-in method count of str object at 0x7f941b883800> = '\nPLAY [List Variables] **********************************************************\n\nTASK [Print a lot of lines] ***...alhost                  : ok=1    changed=0    unreachable=0    failed=0    skipped=0    rescued=0    ignored=0   \n\n'.count
```

And you kind of get the intuitive sense that the assertion is written correctly by testing with 100 and finding that it passes. Also, the stdout shows some duplicated output. All signs seem to point to this being the same thing as what you're seeing, and actually seems to escalate the severity of it.

Let me know how you want to go about this, if you want to rebase and incorporate changes that's fine with me, but I will have more testing I have to do after a rebase happens.

NickLaMuro:
@AlanCoding thank for for the review, however...

This PR is over a year old now, and even my last response to Matthew's most recent comment is over the year mark as well.  At the time, I was working with a code base that I was completely unfamiliar with, a language/ecosystem that is not a primary one for me, and it was the first time I have ever worked with the `zuul` CI environment.  Further more, it seems like the dev setup has completely changed from when I originally setup this PR:

https://github.com/ansible/ansible-runner/commit/7cd038bcb87d8faf9462a3825bd83512c8441fda

And as mentioned in [a previous comment](https://github.com/ansible/ansible-runner/pull/335#issuecomment-529059145), I had to spin up a Vagrant VM just to properly replicate CI failures.

* * *

My priorities have vastly changed since first opening this PR, and as a result, this PR will not take a priority for me any time soon.  At the time, it was a multi-day struggle just to get a working dev environment, so coupled with the new changes to the developer setup, I don't really have the time to reinvest in something like this at this time.

I will try and provide a few comments to what you have already reviewed for some context (from what I can remember), but if this is something that you or your team has decided is now necessary to fix, feel free to take any or none of these changes to fix them in a separate PR as you see fit.  I have no desire to receive credit for any of the previous work I did, so you are welcome to start from scratch using conventions that make more sense for this project if that is the easiest path forward.

Thanks.

test/integration/test_display_callback.py
NickLaMuro: When working properly, the JSON formatted output should print one JSON object per line, so this was a "low tech" way of confirming that was the case.  It is possible that I also wanted to avoid parsing JSON because prior to the fix, it would cause the test suite to stop abruptly, but I can't recall.

However, I also realize this test is brittle, so I am fine with it changing or being removed.

test/integration/test_display_callback.py
NickLaMuro: Commenting in response to the above and https://github.com/ansible/ansible-runner/pull/335#issuecomment-698426758 as it seems to be about this test/comment as well.

* * *

> My preference for test writing here would be that we don't gather facts...

I think that is a fair point, and have no problem with going that route.  This was just a real world use case that I could point to in the tests to say how this could be triggered without needing any external dependencies to do so, but I also agree that relying on OS specific facts is a recipe for disaster, so I think using what you have in https://github.com/ansible/ansible-runner/pull/335#issuecomment-698426758 makes sense.

> I ran this test and confirmed it failed, the JSON data got printed up to 4976 lines, totaling 18.8 k characters. I don't know what the interaction there is with pexpect's max 2k characters.

I vaguely recall that happening for me as well and it was odd, but I am pretty sure there was a reason for it, but I don't recall what it was off hand (if I ever did figure it out).  There was a lot of time digging through the guts of `pexpect` that I didn't take notes on, so I can't remember off hand.

I think what ends up happening is it has something to do with this block of code:

https://github.com/ansible/ansible-runner/blob/f6f6a24538817a84d09feeac85a747b5b9238c0e/ansible_runner/utils.py#L256-L264

Where the `else` is triggered, and the result is that it prints both the line eventually, but also a duplicate of the JSON blob with the newlines not escaped (hence all the extra lines).  Again, this is just what I can vaguely remember, as it has been a while since I have looked at this issue.

* * *

My only suggestion with the suggested change of `"{{ ('F' * 150) | list }}"` would be to bump the 150 up to something like 4000 so we are for sure exceeding the 2k limit of `pexpect`'s default, but perhaps I am missing something with the `| list` portion of that line (which I admit to not fully understanding).

AlanCoding:
Oh of course, 🚚  --> https://github.com/ansible/ansible-runner/pull/531

test/integration/test_display_callback.py
NickLaMuro: Falls in line somewhat with what I said in [this comment](https://github.com/ansible/ansible-runner/pull/335/files#r494469400)

The first two commits of this PR:

https://github.com/ansible/ansible-runner/pull/335/commits/e3a408771ac7d7004f6ab4e1efb0e0e127013dfb
https://github.com/ansible/ansible-runner/pull/335/commits/7a155ab89411fbfa6a3d2adae0c34380f81df032

Made reference to the following PRs respectively:

- https://github.com/ansible/ansible-runner/pull/34
- https://github.com/ansible/ansible-runner/pull/256

Those PRs made the bulk of the rational of the existing block that I simply moved for the fix in this PR:

```python
if not self.suppress_ansible_output:
  sys.stdout.write(
    data.encode('utf-8') if PY2 else data
  )
self._handle.write(data)
self._handle.flush()
```

and I assume they were either untested via an integration test (so **not** actually using `pexpect`), or not tested at all.  

That said, I am a bit hazy on the details at this point, but I believe the intent was that before I just moved this code, I wanted something in place that confirmed the existing fixes.  I would _**hope**_ that I had the discipline at the time to first run my tests with a original fix reverted to confirm it didn't work, then replace it to test that the test was valid... but I can't recall if I did...

Point being, these tests were simply to make sure the functionality/fixes from the aforementioned PRs still worked as intended after my fix in the third commit.

PR #336, opened 2019-08-22
[test_display_callback.py] Use ANSIBLE_PYTHON_INTERPRETER var

NickLaMuro:
Fixes test issues when using newer versions of `ansible` that have this patch:

[https://github.com/ansible/ansible/pull/50163](https://github.com/ansible/ansible/pull/50163)

Since there are tests that end up triggering this warning and mess with the event locations that are returned.

Made a quick refactor to the fixture in this test file as well to break out the default env vars so it was a bit more DRY and re-usable.  Related, and note worthy, I worked on these fixes while working on the following PR:

https://github.com/ansible/ansible-runner/pull/335

And decided to make this an isolated patch instead of building of of that one (or vice versa).  As a result, one of those PR is going to have merge conflicts once the other is merged, so I am prepared to rebase once that happens.

### DIFF
--- a/test/integration/test_display_callback.py
+++ b/test/integration/test_display_callback.py
@@ -5,6 +5,8 @@ import os
 import yaml
 import six
 
+from sys import executable as ANSIBLE_PYTHON_INTERPRETER
+
 from ansible import __version__ as ANSIBLE_VERSION
 
 from ansible_runner.interface import init_runner
@@ -15,13 +17,21 @@ HERE = os.path.abspath(os.path.dirname(__file__))
 
 
 @pytest.fixture()
-def executor(tmpdir, request):
+def default_envvars():
+    return {
+        "ANSIBLE_PYTHON_INTERPRETER": ANSIBLE_PYTHON_INTERPRETER,
+        "ANSIBLE_DEPRECATION_WARNINGS": "False"
+    }
+
+
+@pytest.fixture()
+def executor(tmpdir, request, default_envvars):
     private_data_dir = six.text_type(tmpdir.mkdir('foo'))
 
     playbooks = request.node.callspec.params.get('playbook')
     playbook = list(playbooks.values())[0]
     envvars = request.node.callspec.params.get('envvars')
-    envvars = envvars.update({"ANSIBLE_DEPRECATION_WARNINGS": "False"}) if envvars is not None else {"ANSIBLE_DEPRECATION_WARNINGS": "False"}
+    envvars = envvars.update(default_envvars) if envvars is not None else default_envvars
 
     r = init_runner(
         private_data_dir=private_data_dir,
@@ -167,7 +177,7 @@ def test_callback_plugin_no_log_filters(executor, playbook):
     - shell: echo "PUBLIC"
     - shell: echo "PRIVATE"
       no_log: true
-    - uri: url=https://example.org username="PUBLIC" password="PRIVATE"
+    - uri: url=https://example.org user="PUBLIC" password="PRIVATE"
     - copy: content="PRIVATE" dest="/tmp/tmp_no_log"
 '''},  # noqa
 ])


AlanCoding:
I can, indeed, replicate the failure locally. I see how it was probably caused by the insertion of another warning, creating an unexpected verbose event.

I applied your patch and it seems to fix that particular issue, but the test still fails with

```
py.test test/integration/test_display_callback.py::test_callback_plugin_task_args_leak[playbook0]
================================================================================ test session starts ================================================================================
platform darwin -- Python 3.6.5, pytest-5.1.2, py-1.8.0, pluggy-0.12.0
rootdir: /Users/alancoding/Documents/repos/ansible-runner
collected 1 item                                                                                                                                                                    

test/integration/test_display_callback.py F                                                                                                                                   [100%]

===================================================================================== FAILURES ======================================================================================
__________________________________________________________________ test_callback_plugin_task_args_leak[playbook0] ___________________________________________________________________

executor = <ansible_runner.runner.Runner object at 0x10f90f128>
playbook = {'no_log_on_ok.yml': '\n- name: args should not be logged when no_log is set at the task or module level\n  connection... url=https://example.org username="PUBLIC" password="PRIVATE"\n    - copy: content="PRIVATE" dest="/tmp/tmp_no_log"\n'}
default_envvars = {'ANSIBLE_DEPRECATION_WARNINGS': 'False', 'ANSIBLE_PYTHON_INTERPRETER': '/Users/alancoding/.virtualenvs/runner/bin/python3.6'}

    @pytest.mark.parametrize('playbook', [
    {'no_log_on_ok.yml': '''
    - name: args should not be logged when no_log is set at the task or module level
      connection: local
      hosts: all
      gather_facts: no
      tasks:
        - shell: echo "PUBLIC"
        - shell: echo "PRIVATE"
          no_log: true
        - uri: url=https://example.org username="PUBLIC" password="PRIVATE"
        - copy: content="PRIVATE" dest="/tmp/tmp_no_log"
    '''},  # noqa
    ])
    def test_callback_plugin_task_args_leak(executor, playbook, default_envvars):
        executor.run()
        events = list(executor.events)
        assert events[0]['event'] == 'playbook_on_start'
        assert events[1]['event'] == 'playbook_on_play_start'
    
        # task 1
        assert events[2]['event'] == 'playbook_on_task_start'
        assert events[3]['event'] == 'runner_on_start'
        assert events[4]['event'] == 'runner_on_ok'
    
        # task 2 no_log=True
        assert events[5]['event'] == 'playbook_on_task_start'
        assert events[6]['event'] == 'runner_on_start'
        assert events[7]['event'] == 'runner_on_ok'
        assert 'PUBLIC' in json.dumps(events)
        assert 'PRIVATE' not in json.dumps(events)
        # make sure playbook was successful, so all tasks were hit
>       assert not events[-1]['event_data']['failures'], 'Unexpected playbook execution failure'
E       AssertionError: Unexpected playbook execution failure
E       assert not {'localhost': 1}

test/integration/test_display_callback.py:202: AssertionError
------------------------------------------------------------------------------- Captured stdout call --------------------------------------------------------------------------------

PLAY [args should not be logged when no_log is set at the task or module level] ***

TASK [shell] *******************************************************************
changed: [localhost]

TASK [shell] *******************************************************************
changed: [localhost]

TASK [uri] *********************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (uri) module: username Supported parameters include: attributes, backup, body, body_format, client_cert, client_key, content, creates, delimiter, dest, directory_mode, follow, follow_redirects, force, force_basic_auth, group, headers, http_agent, method, mode, owner, regexp, remote_src, removes, return_content, selevel, serole, setype, seuser, src, status_code, timeout, unix_socket, unsafe_writes, url, url_password, url_username, use_proxy, validate_certs"}

localhost                  : ok=2    changed=2    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   

================================================================================= 1 failed in 2.98s =================================================================================
```

I applied an additional diff:

```diff
diff --git a/test/integration/test_display_callback.py b/test/integration/test_display_callback.py
index 9b651c9..ca130ad 100644
--- a/test/integration/test_display_callback.py
+++ b/test/integration/test_display_callback.py
@@ -177,7 +177,7 @@ def test_callback_plugin_no_log_filters(executor, playbook):
     - shell: echo "PUBLIC"
     - shell: echo "PRIVATE"
       no_log: true
-    - uri: url=https://example.org username="PUBLIC" password="PRIVATE"
+    - uri: url=https://example.org user="PUBLIC" password="PRIVATE"
     - copy: content="PRIVATE" dest="/tmp/tmp_no_log"
 '''},  # noqa
 ])
```

With your diff + this diff, the test passes for me locally.

I suspect this might related to a behavior change in Ansible core. The `url_*` parameters are preferred, user/password are aliases, so I would be happy to go with that if the test matrix for ansible runner still works. Ping @ryanpetrello in case he has any additional thoughts. The playbook involved is:

https://github.com/ansible/ansible-runner/blob/39644c670c0cb30930c7199bce686debe30aac6f/test/integration/test_display_callback.py#L162-L171

NickLaMuro:
@AlanCoding Thanks for looking into this.  This was originally opened to address some failures I was seeing in https://github.com/ansible/ansible-runner/pull/335 locally, but both seem to have CI failures that I am not able to reproduce, so I appreciate the assist.

I will try make those changes today when time permits, but as we have already worked around this bug ourselves here:

https://github.com/ManageIQ/manageiq/pull/19147

I hope you forgive me if I don't make this my first priority.  😅

NickLaMuro:
Sorry for the delay on this, but it looks like I was getting the failures locally that are described here:

- https://github.com/pexpect/pexpect/issues/512
- https://github.com/pexpect/ptyprocess/pull/47

Which took me a while to track down.  Didn't do anything as a result, but I now have a understanding of why that is happening better, and that I can pull a "Lando" and say "It's not my fault".

* * *

Regarding the issue @AlanCoding mentioned, I wasn't able to get that failure to replicate on my machine, but I did rebase in the suggested fix anyway since it does seem to happen in CI, and is consistent with how a different playbook in the same test is set up:

https://github.com/ansible/ansible-runner/blob/e1f0d86643c7eca08cc17411ddbb5c3d1d60b26b/test/integration/test_display_callback.py#L292-L305

So figured it wouldn't hurt.

test/integration/test_display_callback.py
```diff
@@ -167,11 +175,11 @@ def test_callback_plugin_no_log_filters(executor, playbook):
     - shell: echo "PUBLIC"
     - shell: echo "PRIVATE"
       no_log: true
-    - uri: url=https://example.org username="PUBLIC" password="PRIVATE"
+    - uri: url=https://example.org user="PUBLIC" password="PRIVATE"
     - copy: content="PRIVATE" dest="/tmp/tmp_no_log"
 '''},  # noqa
 ])
-def test_callback_plugin_task_args_leak(executor, playbook):
+def test_callback_plugin_task_args_leak(executor, playbook, default_envvars):
```
AlanCoding: I don't think the fixture `default_envvars` is needed for this test? Might be good to clean that up.

NickLaMuro: Yeah, still working out the ins and outs of `pytest`, so I think this was something I tried while trying to get things working.  It works locally for me without it, so I will push up that change.

NickLaMuro:
Current failures on CI:

```
fatal: [localhost]: FAILED! => {"changed": false, "checksum": "b0b7ba4666a92774e0d1afe02f2e20982b307a35", "msg": "Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!"}
```

Seem like they are a build server configuration issue, correct?

ryanpetrello:
@NickLaMuro any chance you're running w/ an environment variable set to an integer or float?

I found another way to reproduce that expect bug recently, and wonder if that's what you've encountered:

https://github.com/pexpect/ptyprocess/pull/47#issuecomment-530054194

AlanCoding:
I'm relatively sure the stuff in this was covered by other PRs. If you manage to still hit failures that shouldn't be happening, we can return to it.

PR #346, opened 2019-08-28
Collect performance data during playbook runs

jladdjr:
Feature ticket:
https://github.com/ansible/tower/issues/3684

### DIFF
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -4,10 +4,12 @@ import time
 import json
 import errno
 import signal
+from subprocess import Popen, PIPE
 import shutil
 import codecs
 import collections
 import datetime
+import logging
 
 import six
 import pexpect
@@ -18,6 +20,8 @@ import ansible_runner.plugins
 from .utils import OutputEventFilter, cleanup_artifact_dir, ensure_str, collect_new_events
 from .exceptions import CallbackError, AnsibleRunnerException
 from ansible_runner.output import debug
+
+logger = logging.getLogger('ansible-runner')
 
 
 class Runner(object):
@@ -140,6 +144,26 @@ class Runner(object):
             for k, v in self.config.env.items()
         }
 
+        # Prepare to collect performance data
+        if self.config.resource_profiling:
+            cgroup_path = '{0}/{1}'.format(self.config.resource_profiling_base_cgroup, self.config.ident)
+
+            import getpass
+            import grp
+            user = getpass.getuser()
+            group = grp.getgrgid(os.getgid()).gr_name
+
+            cmd = 'cgcreate -a {user}:{group} -t {user}:{group} -g cpuacct,memory,pids:{}'.format(cgroup_path, user=user, group=group)
+            proc = Popen(cmd, stdout=PIPE, stderr=PIPE, shell=True)
+            _, stderr = proc.communicate()
+            if proc.returncode:
+                # Unable to create cgroup
+                logger.error('Unable to create cgroup: {}'.format(stderr))
+                raise RuntimeError('Unable to create cgroup: {}'.format(stderr))
+            else:
+                logger.info("Created cgroup '{}'".format(cgroup_path))
+
+
         self.status_callback('running')
         self.last_stdout_update = time.time()
         try:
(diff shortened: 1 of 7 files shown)


ryanpetrello:
@jladdjr I think `master` is the correct branch to target 1.4.x cc @matburt

packaging/rpm/ansible-runner.spec.j2
```diff
@@ -47,6 +47,7 @@ Requires:       python-daemon
 %if 0%{?el7}
 Requires:       pexpect >= 4.6
 Requires:       python-psutil
+Requires:       libcgroup-tools
```
ryanpetrello: @matburt should we *require* this, or is the type of thing AWX should require downstream?  I imagine not *all* runner users will care about this feature.

ansible_runner/runner.py
```diff
@@ -140,6 +144,18 @@ def run(self):
...
 
+        # Prepare to collect performance data
+        if self.config.resource_profiling:
+            cgroup_path = '{0}/{1}'.format(self.config.resource_profiling_base_cgroup, self.config.ident)
+            cmd = 'cgcreate -g cpuacct,memory,pids:{}'.format(cgroup_path)
+            proc = Popen(cmd, stdout=PIPE, stderr=PIPE, shell=True)
+            _, stderr = proc.communicate()
+            if proc.returncode:
+                # Unable to create cgroup, disable profiling
+                logger.warning('Unable to create cgroup: {}'.format(stderr))
```
ryanpetrello: Wondering if `logger.error` might make more sense.

ansible_runner/runner_config.py
```diff
@@ -110,6 +112,11 @@ def __init__(self,
         self.process_isolation_hide_paths = process_isolation_hide_paths
         self.process_isolation_show_paths = process_isolation_show_paths
         self.process_isolation_ro_paths = process_isolation_ro_paths
+        self.resource_profiling = resource_profiling
+        self.resource_profiling_base_cgroup = resource_profiling_base_cgroup
+        self.resource_profiling_cpu_poll_interval = resource_profiling_cpu_poll_interval
+        #TKTK: Update with options for mem, pid count
```
ryanpetrello: `TKTK` 😄 ?

packaging/rpm/ansible-runner.spec.j2
matburt: This should *not* be required by the package since it's an optional element.

ansible_runner/runner.py
matburt: Yep, raise the severity of this.

jladdjr:
Updated docs:

- [x] Updated parameters in docs/intro.rst to match current parameters taken in the code
- [x] Provided more information on characteristics of what happens when perf stats is enabled
- [x] Info on how to set where data is collected, what data will be found there and what is the organization of the data
- [x] Concise example of what that data looks like.

jladdjr:
Also made another pass and made sure tests were current.

@matburt - this is ready for a final review.

ansible_runner/runner_config.py
```diff
@@ -192,13 +202,38 @@ def prepare(self):
...
+            self.env['CGROUP_DISPLAY_RECAP'] = 'False'
+
         self.env['PYTHONPATH'] = python_path + callback_dir
         if self.roles_path:
             self.env['ANSIBLE_ROLES_PATH'] = ':'.join(self.roles_path)
 
         if self.process_isolation:
             self.command = self.wrap_args_with_process_isolation(self.command)
 
+        if self.resource_profiling and self.execution_mode == ExecutionMode.ANSIBLE_PLAYBOOK:
```
matburt: Sorry, I didn't notice this from my previous reviews but why is this limited to just `ANSIBLE_PLAYBOOK`? Seems like it would be useful for ad-hoc mode also?  If not then maybe we should generate an error rather than silently not attaching the cgroup details.

jladdjr: @matburt - 👍 Filed:
https://github.com/ansible/ansible-runner/issues/356

PR #348, opened 2019-08-30
Allow limiting the ansible event datastructure

matburt:
This adds two tunables:

* omit_event_data: excludes everything under the event['event_data']
                   property
* only_failed_event_data: As above, but will include event data for
                          failed events

The idea here is to allow the caller to further constrain the size of
the data payload returned to the interested system.

### DIFF
--- a/ansible_runner/display_callback/events.py
+++ b/ansible_runner/display_callback/events.py
@@ -134,24 +134,9 @@ class EventContext(object):
         return ctx
 
     def get_begin_dict(self):
+        omit_event_data = os.getenv("RUNNER_OMIT_EVENTS", "False").lower() == "true"
+        include_only_failed_event_data = os.getenv("RUNNER_ONLY_FAILED_EVENTS", "False").lower() == "true"
         event_data = self.get()
-        # TODO: Need to rework these values
-        if os.getenv('JOB_ID', ''):
-            event_data['job_id'] = int(os.getenv('JOB_ID', '0'))
-        if os.getenv('AD_HOC_COMMAND_ID', ''):
-            event_data['ad_hoc_command_id'] = int(os.getenv('AD_HOC_COMMAND_ID', '0'))
-        if os.getenv('PROJECT_UPDATE_ID', ''):
-            event_data['project_update_id'] = int(os.getenv('PROJECT_UPDATE_ID', '0'))
-        event_data.setdefault('pid', os.getpid())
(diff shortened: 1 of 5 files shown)


ansible_runner/display_callback/events.py
```diff
@@ -134,40 +134,45 @@ def get(self):
...
-        # TODO: Need to rework these values
+        event = event_data.pop('event', None)
+        if not event:
+            event = 'verbose'
+            for key in ('debug', 'verbose', 'deprecated', 'warning', 'system_warning', 'error'):
+                if event_data.get(key, False):
+                    event = key
+                    break
+        event_dict = dict(event=event)
+        should_process_event_data = (include_only_failed_event_data and event in ('runner_on_failed', 'runner_on_async_failed', 'runner_on_item_failed')) \
```
ryanpetrello: Should you future-proof this with `event.endswith('_failed')` ?

test/integration/test_events.py
```diff
@@ -54,6 +57,24 @@ def test_basic_serializeable():
     json.dumps(events)
 
 
+def test_event_omission():
+    tdir = tempfile.mkdtemp()
+    r = run(private_data_dir=tdir,
+            inventory="localhost ansible_connection=local",
+            omit_event_data=True,
+            playbook=[{'hosts': 'all', 'gather_facts': False, 'tasks': [{'debug': {'msg': "test"}}]}])
+    assert not any([x['event_data'] for x in r.events])
```
ryanpetrello: Do we have tests that verify the `event_data` key for the inverse (when `omit_event_data` is False)?

ansible_runner/display_callback/events.py
matburt: It's a good question. I'm not sure I'm confident that would always match the desired pattern?

test/integration/test_events.py
matburt: https://github.com/ansible/ansible-runner/blob/master/test/integration/test_events.py#L41

Since we tend *not* to rely on this data internally we don't spend much time validating its internals.

ansible_runner/display_callback/events.py
ryanpetrello: 🤷‍♂ I don't care a _lot_ about this.  I just know when `on_async_XYZ` became a thing, we had to update code in a bunch of places.

PR #352, opened 2019-09-13
Fixing issue on inventory variable for runner

### LINKED ISSUE #351: Inventory variable in runner incorrectly parsed
I used the following syntax to execute the ansible runner:

```
r = ansible_runner.run(private_data_dir='/home/cumulus/ansible-runner-api/tmp/', playbook='GetDeviceList.yml', limit=r_limit, inventory="SITE1")
```

When doing this, Ansible runner would create a file called `/home/cumulus/ansible-runner-api/tmp/inventory/hosts` rather than looking for a file named `SITE1` to read in. Based on the documentation within the code, the `inventory=` function should do the following:

```
    :param inventory: Overridees the inventory directory/file (supplied at *private_data_dir/inventory*) with
                      a specific host or list of hosts. This can take the form of
      - Path to the inventory file in the *private_data_dir*
      - Native python dict supporting the YAML/json inventory structure
      - A text INI formatted string
      - A list of inventory sources, or an empty list to disable passing inventory
```

The behaviour of the function is in direct conflict with the first bullet point:
```
      - Path to the inventory file in the *private_data_dir*
```

The expected behaviour should be that the string passed to the variable `inventory=` should be the location of the inventories files within the `private_data_dir`. Instead, what happens is that the string passed into the variable `inventory=` gets written to the file named `hosts`.

The problem is in the following code snippet:
```
ansible-runner/utils.py [185..192]
...
    obj = kwargs.get('inventory')
    if obj and isinventory(obj):
        path = os.path.join(private_data_dir, 'inventory')
        if isinstance(obj, Mapping):
            kwargs['inventory'] = dump_artifact(json.dumps(obj), path, 'hosts.json')
        elif isinstance(obj, string_types):
            if not os.path.exists(obj):
                kwargs['inventory'] = dump_artifact(obj, path, 'hosts')
...
```

The problem code is the last two conditionals. First, the `inventory=` variable is validated as a stringtype. This will succeed. Then, the entire variable input, which is only the string for the relative path of the inventory starting from `private_data_dir` is validated on whether the path exists. This will always fail. As such, to fix this code, the following fix is required:

```
    obj = kwargs.get('inventory')
    if obj and isinventory(obj):
        path = os.path.join(private_data_dir, 'inventory')
        if isinstance(obj, Mapping):
            kwargs['inventory'] = dump_artifact(json.dumps(obj), path, 'hosts.json')
        elif isinstance(obj, string_types):
            if not os.path.exists(os.path.join(path,obj)):
                kwargs['inventory'] = dump_artifact(obj, path, 'hosts')
            else:
                kwargs['inventory'] = os.path.join(path,obj)
```

Even this code is clunky because it assumes that the `inventory/` directory will be the only place where inventories can reside. Withing grossly rewriting the expectations of where the file will be located (and potentially affecting the first conditional action), the simple fix is an adjustment of the path check.

rdarbha:
This pull request is to resolve #351 based on my interpretation of the documentation and code.

### DIFF
--- a/ansible_runner/utils.py
+++ b/ansible_runner/utils.py
@@ -188,8 +188,10 @@ def dump_artifacts(kwargs):
         if isinstance(obj, Mapping):
             kwargs['inventory'] = dump_artifact(json.dumps(obj), path, 'hosts.json')
         elif isinstance(obj, string_types):
-            if not os.path.exists(obj):
+            if not os.path.exists(os.path.join(path,obj)):
                 kwargs['inventory'] = dump_artifact(obj, path, 'hosts')
+            else:
+                kwargs['inventory'] = os.path.join(path,obj)
 
     for key in ('envvars', 'extravars', 'passwords', 'settings'):
         obj = kwargs.get(key)
--- a/ansible_runner/interface.py
+++ b/ansible_runner/interface.py
@@ -87,7 +87,7 @@ def run(**kwargs):
     :param host_pattern: The host pattern to match when running in ad-hoc mode.
     :param inventory: Overridees the inventory directory/file (supplied at ``private_data_dir/inventory``) with
                       a specific host or list of hosts. This can take the form of
-      - Path to the inventory file in the ``private_data_dir``
+      - Path to the inventory file in the ``private_data_dir/inventories`` directory. Takes in a string for the relative location to the hosts file within the ``private_data_dir/inventories`` directory.
       - Native python dict supporting the YAML/json inventory structure
       - A text INI formatted string
       - A list of inventory sources, or an empty list to disable passing inventory


ansible_runner/utils.py
```diff
@@ -188,8 +188,10 @@ def dump_artifacts(kwargs):
         if isinstance(obj, Mapping):
             kwargs['inventory'] = dump_artifact(json.dumps(obj), path, 'hosts.json')
         elif isinstance(obj, string_types):
-            if not os.path.exists(obj):
+            if not os.path.exists(os.path.join(path,obj)):
                 kwargs['inventory'] = dump_artifact(obj, path, 'hosts')
+            else:
+                kwargs['inventory'] = os.path.join(path,obj)
```
AlanCoding: Since it was previously possible to use this by providing an absolute path, should we maybe use a fallback where we treat it as an absolute path? Maybe not exactly as a fallback, but depending on what `os.path.isabs` gives.

AlanCoding:
This fix makes sense to me, but I think we should have some test coverage for it, in addition to the comment I left in code.

matburt:
@rdarbha mind adding some tests for this?

matburt:
ping @rdarbha

matburt:
ping @rdarbha otherwise we'll need to close this.

PR #361, opened 2019-09-30
allow ANSIBLE_HOST_KEY_CHECKING to be configured

ryanpetrello:


wenottingham:
This still overrides the ansible.cfg config file

ryanpetrello:
@wenottingham yep.

wenottingham:
This still overrides the ansible.cfg config file

> @wenottingham yep.

Not really a fan of that... it then requires everything to be done outside of the project.

ryanpetrello:
Yea, I'm not in disagreement with you, just stating that this is a how *number* of AWX-specific settings work (there's no code in runner or AWX to ascertain config values from `ansible.cfg` files - this is a long-standing issue that I'm arguing is outside the scope of _this_ PR).

PR #362, opened 2019-10-01
ansible_runner: updated new_args.extend to os.path.realpath()

falcon78921:
Updated `new_args.extend()` to use `os.path.realpath(self.project_dir)`.
This remedies issues in AWX when the playbook is referenced via symlink and job isolation is enabled (i.e. `bubblewrap`).


ryanpetrello:
@falcon78921 if I remember, the issue is that symlinks break in this code path when `bwrap` is used?

ryanpetrello:
cc @matburt

falcon78921:
@ryanpetrello yes

Issue #51, opened 2018-06-07
Allow providing multiple ssh keys


matburt:
Currently runner assumes a single private key be provided through the module interface and via ``env/ssh_key``.  We should support multiples since behind the scenes we are using ssh-agent.

cdvv7788:
What should be the interface if I want to add several keys? Do you have something in mind?

matburt:
I have two thoughts on this:

a) Have a directory that contains a single key per file OR
b) Support multiple private key stanzas in a single file.

The latter would be easier to deal with when doing something like... creating named pipes (which we support reading from)... it also fits the .ssh directory semantics better the former may be easier from a client management perspective.

bbaassssiiee:
There is a var for this that can be set in ansible.cfg, inventory, group_vars,  host_vars: ansible_ssh_private_key_file

cdvv7788:
@matburt any new ideas about this? This should be quite easy to add (just adding new ssh-add commands at https://github.com/ansible/ansible-runner/blob/master/ansible_runner/runner_config.py#L503) but we need to get a clear interface.
Maybe we could have env/ssh_key be a file or a folder, and if it is a folder we assume that the contents are all keys that must be added? I think that this could be cleaner that several keys on a single file, in the case that some issue needs to be tracked down.

matburt:
I'm good with the file vs folder distinction. The other bit is changing the module interface to be able to accept a list rather than just a big string and then maintaining backwards compatibility for it.

cdvv7788:
@matburt please check the PR I opened. I refactored the code a little to support both cases (obviously it changes the class structure a little) and seems to work. Please give me some feedback, and let me know if this is the right path to follow.

jkohen:
Just as an FYI, when designing a solution for this issue, please consider that in order to support SSH certificates, Ansible-runner will need to take pairs of files: https://github.com/ansible/ansible-runner/issues/488

AlanCoding:
@john-westcott-iv has started some research into the options for managing multiple ssh keys. By that, I mean the tradeoffs of mapping groups/hosts to specific SSH keys versus letting it try multiple keys for every host.

We also had a really good idea from @shanemcd to put the ssh-agent and ssh-add commands into a separate script to make this feel more manageable. That could be done separately from this to pave some groundwork.

PR #365, opened 2019-10-08
Multiple keys support. POC. Do not merge!

cdvv7788:
I refactored a couple things related to keys so ansible-runner can load multiple of them.
This obviously breaks a lot of tests, and I will need to add new tests if we go ahead with it (along of a couple code smells).
For now, this is just a proof of concept to open further discussion about the feature.
This is directly related to https://github.com/ansible/ansible-runner/issues/51

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -181,9 +181,10 @@ class RunnerConfig(object):
 
         # write the SSH key data into a fifo read by ssh-agent
         if self.ssh_key_data:
-            self.ssh_key_path = os.path.join(self.artifact_dir, 'ssh_key_data')
-            open_fifo_write(self.ssh_key_path, self.ssh_key_data)
-            self.command = self.wrap_args_with_ssh_agent(self.command, self.ssh_key_path)
+            for key, value in enumerate(self.ssh_key_data):
+                ssh_key_path = os.path.join(self.artifact_dir, 'ssh_key_data_{}'.format(key))
+                open_fifo_write(ssh_key_path, value)
+                self.command = self.wrap_args_with_ssh_agent(self.command, ssh_key_path)
 
         # Use local callback directory
         callback_dir = self.env.get('AWX_LIB_DIRECTORY', os.getenv('AWX_LIB_DIRECTORY'))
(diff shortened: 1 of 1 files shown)


ansible_runner/runner_config.py
```diff
@@ -181,9 +181,10 @@ def prepare(self):
 
         # write the SSH key data into a fifo read by ssh-agent
         if self.ssh_key_data:
-            self.ssh_key_path = os.path.join(self.artifact_dir, 'ssh_key_data')
-            open_fifo_write(self.ssh_key_path, self.ssh_key_data)
-            self.command = self.wrap_args_with_ssh_agent(self.command, self.ssh_key_path)
+            for key, value in enumerate(self.ssh_key_data):
```
matburt: Something that we do with some of the other options is test to see if this is a list and if it is then we can enumerate over it, otherwise treat it as a single entity. That may be the best way to support backwards compatibility on the interface here.

Otherwise we'd end up breaking existing consumers.

ansible_runner/runner_config.py
```diff
@@ -288,10 +289,16 @@ def prepare_env(self):
             self.settings = dict()
 
         try:
-            self.ssh_key_data = self.loader.load_file('env/ssh_key', string_types)
+            self.ssh_key_data = [self.loader.load_file('env/ssh_key', string_types)]
         except ConfigurationError:
-            output.debug("Not loading ssh key")
-            self.ssh_key_data = None
+            try:
+                key_list = os.listdir(self.loader.abspath('env/ssh_key'))
```
matburt: Instead of doing this we could do an `isdir(...)` to test if this is a directory and load the files up that way?

ansible_runner/runner_config.py
cdvv7788: @matburt I also changed the single case (the current case) to be inside of a list (https://github.com/ansible/ansible-runner/pull/365/files#diff-c62b92cfe059167de0305220ea16db97R292). If we have a list in both cases, we can have the backwards compatibility without needing additional checks, which makes the code at large more obscure.

ansible_runner/runner_config.py
cdvv7788: Well, I am not a fan of nested try/excepts either. I could remove the second one and handle everything inside of that condition. However, having both cases handled with lists simplify the code a little.

Also, I am wondering is self.ssh_key_path is used somewhere else. In this new iteration, that value would not make much sense when having several keys. We could just use a dictionary with the path as the key.

cdvv7788:
@matburt I left some commentaries to your review. Did you have some time to review them?

ansible_runner/runner_config.py
ghost: If you want to use sub folders in env/ssh_key, dump_artifacts(kwargs) in utils.py should be modified:

```suggestion
    for key in ('ssh_key', 'cmdline'):
        obj = kwargs.get(key)
            path = os.path.join(private_data_dir, 'env')
            if isinstance(obj, list):
                os.makedirs(os.path.join(path, key), mode=0o700)
                for idx, item in enumerate(obj):
                    dump_artifact(str(item), os.path.join(path, key), str(idx))
            else:
                dump_artifact(str(kwargs[key]), path, key)
            kwargs.pop(key)
```

Issue #367, opened 2019-10-11
Runner fails to load env/cmdline that includes Unicode characters in Python2


ryanpetrello:
Clone the default demo project and edit `env/cmdline`:

```shell
~/dev/ansible-runner/demo cat env/cmdline
--limit Iñtërnâtiônàlizætiøn
~/dev/ansible-runner/demo ansible-runner run . -m shell -a 'pwd' --hosts localhost
Traceback (most recent call last):
  File "/home/ryanpetrello/dev/ansible-runner/ansible_runner/__main__.py", line 594, in main
    res = run(**run_options)
  File "/home/ryanpetrello/dev/ansible-runner/ansible_runner/interface.py", line 177, in run
    r = init_runner(**kwargs)
  File "/home/ryanpetrello/dev/ansible-runner/ansible_runner/interface.py", line 65, in init_runner
    rc.prepare()
  File "/home/ryanpetrello/dev/ansible-runner/ansible_runner/runner_config.py", line 173, in prepare
    self.prepare_command()
  File "/home/ryanpetrello/dev/ansible-runner/ansible_runner/runner_config.py", line 342, in prepare_command
    self.command = self.generate_ansible_command()
  File "/home/ryanpetrello/dev/ansible-runner/ansible_runner/runner_config.py", line 367, in generate_ansible_command
    args = shlex.split(cmdline_args)
  File "/usr/lib64/python2.7/shlex.py", line 275, in split
    lex = shlex(s, posix=posix)
  File "/usr/lib64/python2.7/shlex.py", line 25, in __init__
    instream = StringIO(instream)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf1' in position 9: ordinal not in range(128)
```

This only happens in Python2, not Python3.

matburt:
Merged

PR #368, opened 2019-10-14
Fix unicode problems for Python 2.7

jbradberry:
The shlex library under Python 2.7 seems pretty strongly to want a str
(i.e. bytes) rather than a unicode string, but under Python 3 the
opposite is true.

Also, update the tests to take the string type into account.

ref #367

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -16,13 +16,13 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-import os
-import re
-import pexpect
-import stat
-import shlex
-import tempfile
 import logging
+import os
+import pexpect
+import re
+import shlex
+import stat
+import tempfile
 
 import six
 from uuid import uuid4
@@ -33,7 +33,7 @@ except ImportError:
 
 from distutils.dir_util import copy_tree
 
-from six import iteritems, string_types
+from six import iteritems, string_types, ensure_binary
 
 from ansible_runner import output
 from ansible_runner.exceptions import ConfigurationError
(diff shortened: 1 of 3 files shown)


elyezer:
> Also, update the tests to take the string type into account.

:heart:

test/unit/test_runner_config.py
```diff
@@ -1,17 +1,11 @@
...
+from functools import partial
+from io import StringIO
 import os
 import re
-import shlex
-
-try:
-    from StringIO import StringIO
-except ImportError:
-    from io import StringIO
```
AlanCoding: A lot of the import changes here are non-functional. Did this change to StringIO affect functionality?

matburt: I was curious about this also, but it does pass all tests so...?

test/unit/test_runner_config.py
```diff
@@ -288,22 +282,21 @@ def test_generate_ansible_command_with_api_extravars():
     assert cmd == ['ansible-playbook', '-i', '/inventory', '-e', 'foo="bar"', 'main.yaml']
 
 
-@pytest.mark.parametrize('cmdline', [
-    '--tags foo --skip-tags'
-    '--limit "䉪ቒ칸ⱷ?噂폄蔆㪗輥"'
+@pytest.mark.parametrize('cmdline,tokens', [
+    (u'--tags foo --skip-tags', ['--tags', 'foo', '--skip-tags']),
+    (u'--limit "䉪ቒ칸ⱷ?噂폄蔆㪗輥"', ['--limit', '䉪ቒ칸ⱷ?噂폄蔆㪗輥']),
```
AlanCoding: It seems like the change made to this test was very subtle. Was this able to catch the bug before you made the patch?

test/unit/test_runner_config.py
jbradberry: `io.StringIO` seems to behave a bit more consistently, and checks the type and raises an error if bytes are passed in, even under Python 2.

jbradberry: and if we are to deal with problems like this, I feel very strongly that we need to be more conscious of the string type being used in these tests.

test/unit/test_runner_config.py
jbradberry: yep!

ansible_runner/runner_config.py
```diff
@@ -335,8 +335,10 @@ def prepare_command(self):
         and if not calls :py:meth:`ansible_runner.runner_config.RunnerConfig.generate_ansible_command`
         """
         try:
-            cmdline_args = self.loader.load_file('args', string_types)
-            self.command = shlex.split(cmdline_args.decode('utf-8'))
+            cmdline_args = self.loader.load_file('args', string_types, encoding=None)
+            if six.PY2:
+                cmdline_args = ensure_binary(cmdline_args, encoding='utf-8')
+            self.command = shlex.split(cmdline_args)
```
beeankha: ✨

Issue #121, opened 2018-08-15
cancelling a job via stop will leave the pid behind


matburt:
This prevents another job from being started

matburt:
I think this is fixed #376

PR #376, opened 2019-10-22
Remove PID file after killing daemon

Tranquilled:
Leaving the PID file in the directory prevents the creation of a new daemon context. This removes the file, allowing a new context to be created.

### DIFF
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -379,7 +379,7 @@ class Runner(object):
         return all_host_events
 
     @classmethod
-    def handle_termination(self, pid, pidfile=None, is_cancel=True):
+    def handle_termination(cls, pid, pidfile=None, is_cancel=True):
         '''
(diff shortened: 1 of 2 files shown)


Tranquilled:
Should close #121

ansible_runner/runner.py
```diff
@@ -379,14 +379,16 @@ def host_events(self, host):
         return all_host_events
 
     @classmethod
-    def handle_termination(cls, pid, is_cancel=True):
+    def handle_termination(self, pid, pidfile=None, is_cancel=True):
```
matburt: This is a class method, don't change this to `self`

Tranquilled: Got it, I hadn't been aware of the convention. Change in latest commit.

Tranquilled:
test_playbook_start fails when I run it on master as well.

PR #377, opened 2019-10-22
Make sleep duration in temp file clearing stage of run configurable

### LINKED ISSUE #366: Configurable Timeout?
Hi

I'm the maintainer of the AWX-RPM project, and we are seeing issues where the cleanup of ansible_runner doesn't work.. 

As a user mentions here: https://github.com/MrMEEE/awx-build/issues/163, this can be fixed by extending the wait in the _delete function..

Is this something that we could change to a configurable option?

Regards

Martin

Tranquilled:
This makes it possible to adjust for systems where deletion takes more resources. Should resolve #366, though per the linked conversation I'm unsure if this actually solves the commenter's problem.

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -336,6 +336,13 @@ def main(sys_args=None):
              "ansible-playbook output (default=None)"
     )
 
+    runner_group.add_argument(
+        "--delete-sleep",
+        default=1,
+        type=float,
+        help="Seconds to wait between attempts to remove temporary files (default=1)"
+    )
+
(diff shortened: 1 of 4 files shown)


matburt:
I don't really want to surface this as a parameter this high up in the application. I think I'd rather like to understand what is holding this file open and solve it without the sleep.

We don't see this problem on the official AWX builds, so I'm wondering if it's something specific to their implementation.

I'd like to see more root cause done on it.

Tranquilled:
That's fair. Thanks for your reflection.

PR #378, opened 2019-10-22
Set version to be the short commit hash when building from repo

### LINKED ISSUE #253: RFE: if pip installed from git, emit version number that shows git hash
Right now the version emitted by `ansible-runner --version` appears to be hard coded. This can lead to some doubt when attempting to determine what changeset is installed on a machine.

Ideally, only released versions of `ansible-runner` would show the version number like `1.3.1` and other installed changesets would show a version like `657fc6d75b32844aa0d240604f138f0e8baf5335`.

`ansible` does something like this

Tranquilled:
Default version is still hardcoded. Creates a warning as the hashes aren't valid versions. Returns default when building from outside the repo folder, but the setup.py script doesn't work from outside the folder either.

Should resolve #253 if the setuptools.dist warning doesn't cause problems.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ ifeq ($(OFFICIAL),yes)
     RELEASE ?= 1
 else
     ifeq ($(origin RELEASE), undefined)
+        VERSION = $(shell git describe --tags)
         RELEASE := 0.git$(shell date -u +%Y%m%d%H).$(shell git rev-parse --short HEAD)
     endif
 endif
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ with open('README.md', 'r') as f:
 
 def version(default):
     try:
-        process = subprocess.check_output(['git', 'describe', '--tags'], text=True)
+        process = subprocess.check_output(['git', 'describe', '--tags'], universal_newlines=True)
         commit = re.match(r'([\d\.]+)[\d-]*([a-g0-9-]*)', process)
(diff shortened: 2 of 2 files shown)


Tranquilled:
Sorry about the force-push; I committed from a computer with the wrong e-mail address configured.

Makefile
```diff
@@ -13,6 +13,7 @@ ifeq ($(OFFICIAL),yes)
     RELEASE ?= 1
 else
     ifeq ($(origin RELEASE), undefined)
+		VERSION = $(shell git describe --tags)
```
shanemcd: This needs to be 1 tab followed by the proper number of spaces. See line below as reference.

AlanCoding:
A lot of changes have landed since this was started, could you rebase against current `devel`?

Tranquilled:
> A lot of changes have landed since this was started, could you rebase against current `devel`?

Done

setup.py
```diff
@@ -2,12 +2,30 @@
...
+                return commit.group(1)
+        else:
+            return default
+    except (Exception):
+        return default
+
+
 setup(
     name="ansible-runner",
     version="2.0.0",
```
AlanCoding, line 31: The `version` method above doesn't seem to ever get called.

There's also the thorny issue of how to deal with official releases, which really would need to be "2.0.0".

shanemcd:
Versioning is now automated with PBR.

PR #380, opened 2019-10-23
Pass "die-with-parent" option to bwrap.

wenottingham:
From the man page:
  Ensures child process (COMMAND) dies when bwrap's parent dies. Kills (SIGKILL) all
  bwrap sandbox processes in sequence from parent to child including COMMAND process
  when bwrap or bwrap's parent dies.

We've seen cases where SSH controlmaster processes detach and stay running
leaving a stale bubblewrap process that prevents cleanup. This seems to
catch those cases.

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -460,7 +460,7 @@ class RunnerConfig(object):
         '''
         cwd = os.path.realpath(self.cwd)
         self.process_isolation_path_actual = self.build_process_isolation_temp_dir()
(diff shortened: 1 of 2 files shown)


matburt:
This is interesting, now I definitely recall seeing this in part of our initial analysis.

angystardust:
I can confirm that this little magic trick fixes my template jobs on awx 8.0! Thanks

aquilinux:
i can confirm that --die-with-parent fixes https://github.com/MrMEEE/awx-build/issues/163 for me as well.

Thanks

PR #381, opened 2019-10-24
Revert accidental indentation change from ac63349d0

wenottingham:
cc @amerck - was this intended? It appears to be breaking things.


amerck:
Good catch! No, this was not intended. I'll blame my editor... Apologies for any inconvenience.

PR #387, opened 2019-11-11
Respect user-specified stdout callback

AlanCoding:


strider:
Tested this PR because I need it and it works as expected! +1

strider:
Hi @AlanCoding,
Any chance to get this PR merged ASAP?

AlanCoding:
While this is probably functional, I don't think a near-term merge is viable. Here is what I would like to do and have been pitching to people when I get a chance:

The content in https://github.com/ansible/ansible-runner/tree/master/ansible_runner/display_callback is shared code in multiple display callbacks (pretty much just job vs. ad hoc commands). The imports work the way they do today because this was all written before Ansible supported user-defined module_utils paths.

This PR does some ugly stuff juggling the DOCUMENTATION string because of the way that imports are done. If we restructure the code to the more proper format without PATH hacks, then we won't need that, and this feature will look almost trivial.

However, adding to the module_utils paths will almost certain break some use cases where the user is already using some custom module_utils paths. So we need to pull in some other code (originally done in AWX) to respect user-defined configs when ansible-runner itself relies on some config vars for basic operation.

So basically, I want these two other items in to pay down some tech debt before truly supporting this. I do very strongly want to do those. It is viable, but less trivial than the hacky stuff here.

cloudnull:
Hi @AlanCoding, maybe we could sync up on IRC (strider, cloudnull), or on the mailing list, to talk about how we can help. We're now using ansible-runner within the OpenStack TripleO project and having the ability to set user defined callbacks is important to us. If we can help usher this feature along by working on the other two items we'd be happy to do so.

AlanCoding:
I'm ready to take this out of draft and call it ready for review. Before I do that, I will squash commits, but I made this branch to preserve this prior history:

https://github.com/ansible/ansible-runner/compare/devel...AlanCoding:stdout_callback_backup?expand=1

So after this comment, cleanup of this PR will start.

ansible_runner/display_callback/module.py
```diff
@@ -60,6 +84,12 @@ class BaseCallbackModule(CallbackBase):
         'playbook_on_no_hosts_remaining',
     ]
 
+    @classmethod
+    def get_documentation(cls):
+        if hasattr(cls, 'PROXY_DOCUMENTATION'):
+            return cls.PROXY_DOCUMENTATION
+        return BASE_DOCUMENTATION % cls.CALLBACK_NAME
```
AlanCoding: My ears are open if someone wants to suggest a more beautiful way of creating the DOCUMENTATION string in this specific logic.

The "perfect" solution would be for awx_display and minimal to both extend from the same doc_fragment (which would still need to be overwritten by the user callback, if present). But that won't be available until we are fully in the world of collections. So to me, the general thing being doing here seems fully justifiable.

ansible_runner/display_callback/module.py
```diff
@@ -374,7 +404,7 @@ def v2_runner_on_failed(self, result, ignore_errors=False):
             event_loop=self._get_event_loop(result._task),
         )
         with self.capture_event_data('runner_on_failed', **event_data):
-            super(BaseCallbackModule, self).v2_runner_on_failed(result, ignore_errors)
+            super(BaseCallbackModule, self).v2_runner_on_failed(result, ignore_errors=ignore_errors)
```
AlanCoding: in Ansible core, these are kwargs, not args. Thus the change.

ansible_runner/runner_config.py
```diff
@@ -195,6 +195,9 @@ def prepare(self):
         if python_path and not python_path.endswith(':'):
             python_path += ':'
         self.env['ANSIBLE_CALLBACK_PLUGINS'] = ':'.join(filter(None,(self.env.get('ANSIBLE_CALLBACK_PLUGINS'), callback_dir)))
+        # If stdout callback specified, pass info for this stdout callback plugin to subclass it
+        if 'ANSIBLE_STDOUT_CALLBACK' in self.env and self.env['ANSIBLE_STDOUT_CALLBACK'] not in ('awx_display', 'minimal'):
+            self.env['RUNNER_STDOUT_CALLBACK_PROXY'] = self.env['ANSIBLE_STDOUT_CALLBACK']
```
AlanCoding: Needed this name `RUNNER_STDOUT_CALLBACK_PROXY` to be very specific. It could have a 20 digit uuid in its name for all I care, it's only passing it from parent context to child context and is a single static string.

test/data/callback_options/project/task_status.yml
```diff
@@ -0,0 +1,37 @@
...
+
+- hosts: 7_unreachable
+  gather_facts: false
+  name: Target unreachable hosts in order to get unreachable status
+  tasks:
+
+    - name: Reach out to the unreachable host
+      ping:
+      vars:
+        ansible_host: 'invalid.invalid'
```
AlanCoding: Somewhere else I'm going with this - I want to test @wenottingham's #405 with it, because I'm worried about our ability (my ability) to correctly identify all task-ending callback types.

test/data/display_callback/project/task_status.yml
```diff
@@ -0,0 +1,37 @@
...
+
+- hosts: 7_unreachable
+  gather_facts: false
+  name: Target unreachable hosts in order to get unreachable status
+  tasks:
+
+    - name: Reach out to the unreachable host
+      ping:
+      vars:
+        ansible_host: 'invalid.invalid'
```
AlanCoding: yeah, I don't like the duplication either, but runner right now does not like to combine programatic and file inputs, so that's why it ended up like this.

test/data/README.md
```diff
@@ -0,0 +1,15 @@
+### Ansible Runner Test Data Directories
+
+Subfolders in this directory should contain test cases in the form of
```
matburt: 👍

AlanCoding:
This was mistakenly merged into the master branch, which changed to the devel branch at some point. That's not good.

PR #389, opened 2019-11-12
Prepare for centos-8 testing

pabelanger:
Because centos-8 images don't have python2 by default, and since python2
is EOL in less then a month, use pip3 when possible.  This should mean,
only centos-7 will continue to use pip2, and fedora / centos-8 python3.

Depends-On: https://github.com/ansible/ansible-runner/pull/388


pabelanger:
recheck

PR #392, opened 2019-11-13
Make sure we install sshpass into the ssh container image

### LINKED ISSUE #390: sshpass is missing from Docker image
Since version 1.4, sshpass is missing from the ansible-runner Docker image.
It would be helpful to include it by default.

matburt:
fixes #390


matburt:
recheck

PR #394, opened 2019-11-27
adds new typing system to runner

privateip:
This commit adds a new typing system to ansible-runner that allows
for programmatically buiding playbook and inventory objects.  The added
objects for playbook and inventory allow developers to construct a
playbook and/or inventory using native Python constructs instead of
having to create them out-of-band.  Once created the objects can be
serialized to JSON and injected directly to ansible-runner for
execution.

This commit also updates Pipefil with missing developer packages and
adds python coverage reporting to tox.

Signed-off-by: Peter Sprygada <psprygad@redhat.com>


ryanpetrello:
Can we add some docs outlining what this does?

privateip:
@ryanpetrello the docstrings _should_ be self explanatory about how to use the native objects.  I want to build out some usage documentation but was waiting until we had agreement to merge this PR first.

privateip:
Updated PR with docs attached forthcoming,

privateip:
@ryanpetrello PR updated with documentation

docs/playbook_api.rst
```diff
@@ -0,0 +1,522 @@
...
+                    "command": "ls -1"
+                }
+            ],
+            "hosts": "localhost"
+        }
+    ]
+
+Task lists can also contain :class:`Block
+<ansible_runner.playbook.tasks.Block>` items.  To create a new task block
+simple omit the `action` keyword argument.
```
AlanCoding: simple -> simply

AlanCoding:
I think the playbook_api.rst file is about the most relevant to the question

https://github.com/ansible/ansible-runner/pull/394/files#diff-5e38d0d2ccb38f05ee67557235b818f8

I just want clarification on this

> can can be serialized to a JSON data structure that can be directly consumable by **Ansible Runner**.

...if you do what? JSON isn't an inventory format. Is the intent that the user produces the JSON output and dumps it to a file? Then enables a custom inventory plugin to tell Ansible how to read it? User creates an executable script that reads the JSON content and prints it? Is it passed directly through the programatic interface where ansible_runner processes it?

privateip:
@AlanCoding i have updated the PR to include a new generic usage example that shows how the `Inventory` and `Playbook` types can be used as input to the `run()` function.  Hopefully that helps to put the entire picture together.  It should be noted that currently the `Inventory` and `Playbook` types are not in the critical code path for runner and must be explicitly used by the developer.

To address your specific quesitons...

> ...if you do what? 

See updated PR usage example

> JSON isn't an inventory format. Is the intent that the user produces the JSON output and dumps it to a file? 

Not direclty an inventory format but because of the relationship between JSON and YAML, a JSON structure can be read by the YAML inventory plugin.  This takes advantage of that.

> Then enables a custom inventory plugin to tell Ansible how to read it? 

Nope, uses the YAML inventory plugin 

> User creates an executable script that reads the JSON content and prints it? 

Nope, the instantiation of the objects are intended to be done by the application developer invoking (or embedding) `ansible_runner`

> Is it passed directly through the programatic interface where ansible_runner processes it?

The output from `serialize()` is passed to the appropriate kwarg for `run()` or `run_async()`

docs/playbook_api.rst
privateip: fixed

AlanCoding:
Interesting, that information helps.

Ansible Tower itself uses JSON outputted by a script. I always shied away from the YAML option because the examples tend to assume a strict tree structure for the groups. In other words, it's not obvious from the get-go how to make parent-child relationships of A->C and B->C, where A,B,C are all groups.

...reading more...

When you do `inventory.children.new('group_y')`, what is the type that gives? A `Group` or a `Child`?

The native types for Ansible itself are `Host`, `Group` and some weirder things like `InventoyData`, and the loader classes and the plugins (which you don't need). Children are just a python list that primarily exists on the `Group` objects (I am more than happy to neglect the ancestor caching that goes on in memory). The issue of shared parents is one of those classic gotchas. In YAML inventory it's possible to give multiple versions of the same group, and I worry about that looking at this `Child` class. Using YAML, you'll probably want to serialize the same group twice in the A,B,C case, and this isn't necessarily a problem as long as you're not at huge scale.

privateip:
> When you do inventory.children.new('group_y'), what is the type that gives? A Group or a Child?

When calling `new()` it is always a new instance of `Child` and they can be nested.  See the following example.

```
>>> inventory = Inventory()
>>> 
>>> child_A = inventory.children.new('child_A')
>>> child_B = child_A.children.new('child_B')
>>> child_C = child_B.children.new('child_C')
>>> child_A1 = inventory.children.new('child_A1')
>>> 
>>> import json
>>> print(json.dumps(inventory.serialize(), indent=4))
{
    "all": {
        "children": {
            "child_A": {
                "children": {
                    "child_B": {
                        "children": {
                            "child_C": {
                                "vars": {}
                            }
                        }, 
                        "vars": {}
                    }
                }, 
                "vars": {}
            }, 
            "child_A1": {
                "vars": {}
            }
        }
    }
}
>>> 
>>> type(child_A)
<class 'ansible_runner.inventory.children.Child'>
>>> type(child_B)
<class 'ansible_runner.inventory.children.Child'>
>>> type(child_C)
<class 'ansible_runner.inventory.children.Child'>
>>> type(child_A1)
<class 'ansible_runner.inventory.children.Child'>
```
 > The issue of shared parents is one of those classic gotchas. In YAML inventory it's possible to give multiple versions of the same group, and I worry about that looking at this Child class. Using YAML, you'll probably want to serialize the same group twice in the A,B,C case, and this isn't necessarily a problem as long as you're not at huge scale.

I think it's important to state that the initial commit of the `Inventory` object is not designed to be an optimized "intelligent" data structure.  It just eases the burden of creating a valid inventory that can be fed into `run()` and actioned upon.  This could be a follow up project at some point down the road.  For now, it is the developers responsibility to optimize inventory creation.

AlanCoding:
That does A->B and B->C. I am worried about the correctness of doing A->C and B->C, probably following the syntax of `child_A.children.new('child_C')` and `child_B.children.new('child_C')`, although it's not really new in the latter case. Maybe it would be done like `child_B.children['child_C'] = child_A.children.new('child_C')`

And I think I'm starting to get that all groups are denoted by the `Child` class here. The two relations `children` and `hosts` do make sense, and are pretty standard.

privateip:
You can also explicitly set the child objects instead of using the `new()` constructor and all type validations still are enforce.  That means you can do something like the following:

```
>>> from ansible_runner.inventory import Inventory
>>> from ansible_runner.inventory.children import Child
>>> 
>>> 
>>> inventory = Inventory()
>>> 
>>> child_A = Child()
>>> child_B = Child()
>>> child_C = Child()
>>> 
>>> inventory.children['child_A'] = child_A
>>> inventory.children['child_B'] = child_B
>>> inventory.children['child_A'].children['child_C'] = child_C
>>> inventory.children['child_B'].children['child_C'] = child_C
>>> 
>>> import json
>>> print(json.dumps(inventory.serialize(), indent=4))
{
    "all": {
        "children": {
            "child_A": {
                "children": {
                    "child_C": {}
                }
            }, 
            "child_B": {
                "children": {
                    "child_C": {}
                }
            }
        }
    }
}
```

In this case, both `child_A` children and `child_B` children contain a reference to `child_C` so any updates to `child_C` are reflected as such

``` 
>>> child_C.hosts.new('localhost')
{}
>>> print(json.dumps(inventory.serialize(), indent=4))
{
    "all": {
        "children": {
            "child_A": {
                "children": {
                    "child_C": {
                        "hosts": {
                            "localhost": {}
                        }
                    }
                }
            }, 
            "child_B": {
                "children": {
                    "child_C": {
                        "hosts": {
                            "localhost": {}
                        }
                    }
                }
            }
        }
    }
}
```

AlanCoding:
ok thanks, that does match the expectation correctly

PR #399, opened 2019-12-11
Changes for ppc64le support

seth-priya:
Changes for adding ppc64le support by introducing an additional "ARCH" build arg in Dockerfile. Additional devel packages needed for "pip install" to succeed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,13 +11,10 @@ ADD demo/inventory /runner/inventory
 # Install Ansible Runner
 RUN yum-config-manager --add-repo https://releases.ansible.com/ansible-runner/ansible-runner.el7.repo && \
 	yum install -y epel-release && \
-	yum install -y python-pip ansible-runner bubblewrap sudo rsync openssh-clients sshpass \
-			gcc gcc-c++ make python-devel libffi-devel openssl-devel && \
-	pip install --upgrade pip && \
+	yum install -y python-pip python-cryptography ansible-runner bubblewrap sudo rsync openssh-clients sshpass && \
 	pip install --no-cache-dir ansible && \
 	localedef -c -i en_US -f UTF-8 en_US.UTF-8 && \
 	chmod +x /bin/tini /bin/entrypoint && \
-	yum remove -y --setopt=tsflags=noscripts gcc gcc-c++ make python-devel libffi-devel openssl-devel && \
 	rm -rf /var/cache/yum
 
 # In OpenShift, container will run as a random uid number and gid 0. Make sure things
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,8 @@
 FROM centos:7
 
-ADD https://github.com/krallin/tini/releases/download/v0.18.0/tini /bin/tini
+ARG ARCH
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -9,10 +11,14 @@ ADD demo/inventory /runner/inventory
 # Install Ansible Runner
 RUN yum-config-manager --add-repo https://releases.ansible.com/ansible-runner/ansible-runner.el7.repo && \
 	yum install -y epel-release && \
-	yum install -y python-pip ansible-runner bubblewrap sudo rsync openssh-clients sshpass && \
+	yum install -y python-pip ansible-runner bubblewrap sudo rsync openssh-clients sshpass \
+			gcc gcc-c++ make python-devel libffi-devel openssl-devel && \
```
matburt: What do we need a full toolchain for here?

wenottingham: that smells like the python-cryptography dependency of ansible attempting to be built natively rather than using the system package. This should be fixed to ensure that you're pulling the system python-cryptography build

seth-priya: @matburt @wenottingham thanks for reviewing and providing your feedback, yes the issue is because of the cryptography module. the "pip" is not yet supported on ppc64le, as also, confirmed from https://cryptography.io/en/latest/installation/. 

So, it can either be built from source, which happens by default if the necessary *-devel packages are present in the container, hence I installed and later removed them.

The other option, I think is to try and go via the "conda" route as the requisite packages are present there https://anaconda.org/conda-forge/cryptography/files?channel=main , any thoughts here?

wenottingham: That doesn't make sense. If you're using the same CentOS base, a python-cryptography build already exists at the system RPM level. Use that.

seth-priya: yes, missed that, will make the change, thank you

seth-priya: @wenottingham @matburt have updated, please review, thank you!

seth-priya: any thoughts? thank you!!

matburt: On the surface this is fine I reckcon, we just don't have any need or tooling for this so it's not possible for us to generate or test official images based on this.

seth-priya: Thank you @matburt do you think it would make sense to merge the changes so that folks who need Power images can still build it off the official source for now? For generation of official images, if I may ask, is there an internal CI that you are using? We have support for Power on Travis today, so I can help setup things there or we can help provide Power nodes and help add them to your setup? We can work on enhancing the tooling, look at any other alternatives you suggest as well.

seth-priya:
@matburt @wenottingham just a gentle reminder - please let me know if anything else needs to be done from my side on this or would you consider merging the changes? Thank you!

matburt:
image

Rebase instead of merging from master. The commit history of this PR will need to be scrubbed of that merge commit now.

seth-priya:
@matburt thanks, will have a look, if it becomes too ambiguous, I'll raise a new one

seth-priya:
Closing this, will reopen a new one with the latest upstream code, built and validated with the updated centos:8 based dockerfile

PR #400, opened 2019-12-17
Respect stdout callback being set

willthames:
If stdout callback is set, either through config or environment
variable, it should be respected, not overridden.


willthames:
This seems to work if the environment variable is set (as expected) but not if `stdout_callback` is set in ansible.cfg

It's not clear to me if there's a mechanism to respect ansible.cfg for this variable, but it's an acceptable failure for me.

matburt:
This *does* have the effect of short circuiting some of the capability of Runner for gathering this information as part of its event payloads. I wonder if we should wrap this up into a larger effort to clarify support for ansible.cfg.

willthames:
It's not just not supporting ansible.cfg though - there *has* to be some way to override the stdout callback - you can't possibly predict that people will want only either one of the two built in stdout callbacks

wenottingham:
see #387

PR #404, opened 2020-01-15
Set created time for verbose events

AlanCoding:
in AWX, there is a gap between the time when an event is produced by the `ansible` or `ansible-playbook` process and the time this event is saved to the database. I would like to be able to reliably obtain this time delta from historical data in the database. Currently, however, many events have `created` and `modified` fields identically equal. In these cases, no measurement of production time is ever taken or recorded.

This adds the field for the case where the callback receiver does not add it. Obviously this is the case for verbose events, because the code path in the callback receiver is never touched.

I believe that these types of events could have (previously) had a created/modified delta, but only due to updating fields after the initial insert, which are operations that were recently removed.

### DIFF
--- a/test/unit/test_runner.py
+++ b/test/unit/test_runner.py
@@ -82,7 +82,16 @@ def test_cancel_callback_error(rc):
         Runner(config=rc, cancel_callback=kaboom).run()
 
 
-@pytest.mark.serial
+def test_verbose_event_created_time(rc):
+    rc.command = ['echo', 'helloworld']
+    runner = Runner(config=rc)
+    status, exitcode = runner.run()
+    assert status == 'successful'
+    assert exitcode == 0
+    for event in runner.events:
+        assert 'created' in event, event
+
+
 @pytest.mark.parametrize('value', ['abc123', six.u('Iñtërnâtiônàlizætiøn')])
 def test_env_vars(rc, value):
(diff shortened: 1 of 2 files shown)


test/unit/test_runner.py
```diff
@@ -82,6 +82,16 @@ def kaboom():
...
 
 
+def test_verbose_event_created_time(rc):
+    rc.command = ['echo', 'helloworld']
+    runner = Runner(config=rc)
+    status, exitcode = runner.run()
+    assert status == 'successful'
+    assert exitcode == 0
+    for event in runner.events:
+        assert 'created' in event, event
```
AlanCoding, line 92: fails prior to patch. But I wouldn't object to getting rid of it, since it's covered by modifications to assertions in other tests below.

AlanCoding:
Calling the field `start` would be more expressive, since it's usually set at the beginning of event data collection, but the runner event data should be a stable data structure, and I don't consider it to be an option to remove the `created` field.

For verbose events, it's probably the case that start time equals the end time. This is because, when there is no event contextual data to combine with, we dispatch it right away. (we may dispatch _multiple_ events)

These details should probably go into the docs. The closest I can find is here:

https://ansible-runner.readthedocs.io/en/latest/intro.html#runner-artifact-job-events-host-and-playbook-events

Things like `uuid`, `counter`, and `created` are real bread & butter fields for integrations to do stuff with.

chrismeyersfsu (approved):

PR #405, opened 2020-01-21
Annotate all runner_on_* events with their start/stop/duration times

wenottingham:
This is incredibly useful information to have for a variety of needs.

Stolen from https://github.com/ansible/ansible-baseline/.

### DIFF
--- a/ansible_runner/display_callback/module.py
+++ b/ansible_runner/display_callback/module.py
@@ -20,6 +20,7 @@ from __future__ import (absolute_import, division, print_function)
 # Python
 import collections
 import contextlib
+import datetime
 import sys
 import uuid
 from copy import copy
@@ -34,6 +35,10 @@ from .events import event_context
 from .minimal import CallbackModule as MinimalCallbackModule
 
 CENSORED = "the output has been hidden due to the fact that 'no_log: true' was specified for this result"  # noqa
+
+
+def current_time():
+    return datetime.datetime.utcnow()
 
 
 class BaseCallbackModule(CallbackBase):
@@ -62,6 +67,7 @@ class BaseCallbackModule(CallbackBase):
 
     def __init__(self):
         super(BaseCallbackModule, self).__init__()
+        self._host_start = {}
         self.task_uuids = set()
         self.duplicate_task_counts = collections.defaultdict(lambda: 1)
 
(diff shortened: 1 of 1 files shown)


AlanCoding:
Right now I also have https://github.com/ansible/ansible-runner/pull/404 open.

Some quick thoughts:

 - if a client wants the duration, they can subtract the values on their own, no need to do it here
 - the start timestamp is duplicated with the `created` timestamp, ideally we should only keep one
 - this doesn't work for verbose events, and I don't know if it can be made to work

wenottingham:
>  if a client wants the duration, they can subtract the values on their own, no need to do it here

I disagree - we want to avoid as much post-processing of the returned events as possible - if you're operating across one million events doing data processing, having to query all the events and then compute can be expensive.

> the start timestamp is duplicated with the created timestamp, ideally we should only keep one

Will look.

AlanCoding:
This is the other place this is done:

https://github.com/ansible/ansible-runner/blob/e1f0d86643c7eca08cc17411ddbb5c3d1d60b26b/ansible_runner/display_callback/events.py#L158

That also does isoformat

wenottingham:
> This is the other place this is done:
> 
> https://github.com/ansible/ansible-runner/blob/e1f0d86643c7eca08cc17411ddbb5c3d1d60b26b/ansible_runner/display_callback/events.py#L158
> 
> That also does isoformat

That's when it processes it from the disk, if I'm reading that right, which is already after the event has finished.  That's sufficiently removed from the actual event generation in Ansible that I'd prefer to do that directly on the task start.

AlanCoding:
> That's when it processes it from the disk, if I'm reading that right, which is already after the event has finished.

I do see now how these are really different animals.  Logic for `created` is in generation of the begin dict, but I think that also applies to all the timestamps saved in this PR too. Aside from time spent in the callback plugin itself, the `start` time should be the same as the `created` timestamp for the runner_on_start events and the `end` time should be the same as the `created` for the runner_on_ok/skipped/etc. events. So I do think that all of this should be reconstructable from the client's perspective.

AlanCoding:
⚠️  D-D-D-DANGER ZONE ⚠️

This PR is targeting the `master` branch, but the default branch was changed to `devel`. If you can change your base branch, please do that now. Otherwise you may need to close and make a new PR.

AlanCoding:
Apparently it let me, so I went ahead and changed the base

ansible_runner/display_callback/module.py
```diff
@@ -498,6 +520,7 @@ def v2_runner_on_start(self, host, task):
             host=host.get_name(),
             task=task
         )
+        self._host_start[host.get_name()] = current_time()
```
ryanpetrello, line 523: @matburt @wenottingham,

This doesn't work on Ansible pre-2.8, because `runner_on_start` was introduced in 2.8:

https://github.com/ansible/ansible/pull/47684/

PR #409, opened 2020-02-07
[WIP] Run display callback as collection

AlanCoding:
This initial version is not yet compatible with pre-2.8 Ansible, so that still needs to be worked on. I want to get the matrix results, and then go from there.


AlanCoding:
meant for this to be a draft PR, anyway, things have been getting more complicated with it

PR #412, opened 2020-02-17
Allow customizing display callback, refactor test data

AlanCoding:
I mistakenly merged https://github.com/ansible/ansible-runner/pull/387

That should be reverted (possibly force-pushed) in the `master` branch. This is targeting the `devel` branch correctly.


matburt:
@AlanCoding what's going on here?

AlanCoding:
I don't understand the question, see the history in #387

I updated this somewhat, however, I'm going to close it because of developments in collections since it was first opened. With the new rules imposed, I believe this would block us from ever having runner content pass `ansible-test` sanity tests in any meaningful way. I've touched on that rarely, but I do believe it should be a top priority for this project.

PR #415, opened 2020-02-22
Add boto to Docker image so we can run against AWS

fsproru:
Hey there, submitting a PR for adding boto library to the docker image which is necessary to run against AWS. There are more details in this issue: https://github.com/ansible/ansible/issues/15019

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ ADD demo/inventory /runner/inventory
 RUN yum-config-manager --add-repo https://releases.ansible.com/ansible-runner/ansible-runner.el7.repo && \
 	yum install -y epel-release && \
 	yum install -y python-pip ansible-runner bubblewrap sudo rsync openssh-clients sshpass && \
-	pip install --no-cache-dir ansible && \
+	pip install --no-cache-dir ansible boto botocore boto3 && \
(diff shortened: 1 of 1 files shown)


andyfeller:
How do you decide which pip modules to package with this and which end users layer on top?  Should we add whatever is needed for azure or gce?

fsproru:
@andyfeller, good question. I tend to make Docker images that have everything required to solve a particular task. If the goal of this image is to enable the users to run Ansible against Azure and Google cloud, I would add the packages for it.

andyfeller:
My point is this shouldn’t be done because it’s an individual use case rather than something needed unilaterally by all ansible users.  It also doesn’t get people out of repackaging this image when they need other pip module, rpm packages, and configuration need to be customized.

fsproru:
No worries, I’ll update the form readme with the modifications added so people willing to run against AWS can use our fork. If you want to mention the url to the fork in the original readme, that’s fine too. This way people find the image they need with the right packages installed.

shanemcd:
I share the concerns @andyfeller mentioned above. You can always inherit `FROM ansible/ansible-runner` and do whatever you need to get the job done.

fsproru:
Yep, makes sense. Thanks guys. Closing this one.

PR #422, opened 2020-03-06
fix: extra_vars type in RunnerConfig

mnecas:
There was issue with extra_vars that they were passing only string values.
This fix will allow return lists and booleans etc.


mwperina (approved):
+1

mnecas:
Hi, @matburt,
could you please take a look at this PR?

matburt (changes requested):
Just a few comments:

Rewrite the commits to be fewer in number and more accurately describe the change that's occurring and why.

Also, rather than changing the tests to cover just this use case, add the original tests back in and write new tests to cover the new data type(s).

matburt (approved):

PR #432, opened 2020-03-18
Allow running jobs via Receptor 

ghjm:
```
$ receptor --node-id node1 node & ansible-runner run . -m ping --hosts localhost --via-receptor node1
[1] 166987
localhost | SUCCESS => {
    "changed": false,
    "ping": "pong"
}
```


ansible_runner/receptor_plugin.py
```diff
@@ -0,0 +1,164 @@
...
+        'ident',
+        'ignore_logging',
+        'inventory',
+        'limit',
+        'module',
+        'module_args',
+        'omit_event_data',
+        'only_failed_event_data',
+        'playbook',
+        'verbosity',
```
ryanpetrello: Is this list hand-curated from somewhere, and is there a reason we're whitelisting things (I wonder how maintainable this will be as runner config gains new parameters/flags)?  It's kind of gross that `ansible_runner.run` is just a giant docstring that takes blanket `**kwargs` (not your doing, I realize), but is there some underlying list of these defined in code somewhere that we could glean using `inspect.getargspec`?

ansible_runner/receptor_plugin.py
```diff
@@ -0,0 +1,164 @@
...
+            kwargs = json.dumps(remote_options, cls=UUIDEncoder)
+            ti = tarfile.TarInfo('kwargs')
+            ti.size = len(kwargs)
+            ti.mtime = time.time()
+            tar.addfile(ti, io.BytesIO(kwargs.encode('utf-8')))
+        tmpf.flush()
+
+        # Run the job via Receptor
+        if receptor_node_id:
+            receptor_args = f"--node-id {receptor_node_id} node --server-disable".split()
```
ryanpetrello, line 114: *Is* runner at this point dedicated to py3-only?  I realize that if you're using this feature, and thus receptor, you're sort of locked into python3, but I wonder if this f-string is going to get handled as a syntax error for those using runner under py2?

ansible_runner/receptor_plugin.py
```diff
@@ -0,0 +1,164 @@
...
+        else:
+            receptor_args = "node --server-disable".split()
+        config = ReceptorConfig(receptor_args)
+        config._is_ephemeral = True
+        controller = Controller(config)
+        controller.run(run_func)
```
ryanpetrello:
```python
try:
    run()
finally:
    cleanup_tmpdir()
```

ansible_runner/receptor_plugin.py
```diff
@@ -0,0 +1,164 @@
...
+        controller.cleanup_tmpdir()
+
+        res = Receptor_Runner_Result()
+        res.rc = 0
+        return res
```
ryanpetrello: Unless you intend to reference this class in some way, you *can* just use this one liner if you're just looking for a way to have an object w/ `rc=0`.

```python
return type('ReceptRunnerResult', (), {'rc': 0})
```

ansible_runner/receptor_plugin.py
```diff
@@ -0,0 +1,164 @@
...
+
+def run_via_receptor(receptor_node, receptor_peer, receptor_node_id, run_options):
+
+    async def read_responses():
+        while True:
+            message = await controller.recv()
+            if message.header.get("eof", False):
+                break
+            elif message.payload:
+                content = json.loads(message.payload.readall())
```
ryanpetrello, line 57: This is pretty swanky.

ryanpetrello (approved):
This needs documentation under https://github.com/ansible/ansible-runner/tree/devel/docs for how receptor integration works.

Otherwise, this looks pretty reasonable to me.

@matburt do you have thoughts about this PR?

ansible_runner/receptor_plugin.py
ghjm: Curated is too strong a word.  It's the parameters that I thought made sense to run remotely.  I specifically want to avoid trying to pass along handlers that can't be called from the remote side, or anything that isn't json serializable.  Will look into finding a better way to do this.

ansible_runner/receptor_plugin.py
ghjm: Runner is still being linted on python 2.7.  If it doesn't need to be, then that would surely make things easier.  As far as the f-string, on python 2.7, this module doesn't get imported at all.  See https://github.com/ansible/ansible-runner/blob/6c37d73c1ddbf34c829bd3adf8280ee1c7a12054/ansible_runner/__main__.py#L45.

ansible_runner/receptor_plugin.py
ghjm: wilco

ansible_runner/receptor_plugin.py
ghjm: Nice

ansible_runner/receptor_plugin.py
ghjm: What, that it's a break and an elif?

ryanpetrello: No, just that it all works :D

ryanpetrello:
Thanks for addressing feedback, this looks good to me 👍

ansible_runner/receptor_plugin.py
ghjm: I don't see how we could generate this list by introspecting the `run` method.  `inspect.getargspec` just tells us that it takes kwargs.  I guess we could annotate the docstring in some way and then parse it with a package like `docstring-parser` and look for some kind of remote keyword on each documented argument, but that seems pretty gross, and no more maintainable than having this whitelist.

I moved the list to a variable outside the function so that at least if some future code needs to know its content, it can read it.  Please let me know if you've got a better idea how to do this.

ryanpetrello: > parse it with a package like docstring-parser

ryanpetrello: > I moved the list to a variable outside the function so that at least if some future code needs to know its content, it can read it

I don't have any better ideas off the top of my head (other than runner itself actually itemizing its argspec instead of taking in arbitrary `**kwargs`).  Moving this into a defined list is fine for now, I was just curious where this list came from.

chrismeyersfsu:
Not to be that guy but ... tests?

ansible_runner/receptor_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+    return func
+
+
+@receptor_plugin_export
+def execute(message, config, result_queue):
+    private_dir = None
+    try:
+        private_dir = tempfile.TemporaryDirectory()
+        with tarfile.open(fileobj=message.fp, mode='r:gz') as tar:
+            tar.extractall(path=private_dir.name)
```
chrismeyersfsu: This is dangerous if you don't completely trust the creator of the tar file.

```
2. Symlink file attack:

    foo links to /etc/passwd.
    Another member named foo follows, its data overwrites the target file's data.
```
https://bugs.python.org/issue21109
https://bugs.python.org/issue21109#msg215656

ansible_runner/receptor_plugin.py
```diff
@@ -0,0 +1,162 @@
...
+    'playbook',
+    'verbosity',
+)
+
+
+def run_via_receptor(receptor_node, receptor_peer, receptor_node_id, run_options):
+
+    async def read_responses():
+        while True:
+            message = await controller.recv()
```
chrismeyersfsu, line 53: If a network blip happens and receptor disconnects. What happens?

chrismeyersfsu (commented):
Looks good. This is best tested at an integration level.

ansible_runner/receptor_plugin.py
ghjm: Receptor has internal reconnect logic, and won't drop buffer content on disconnect.  It's supposed to be resilient to transient network errors.

For the temporary receptor node inside Runner, it won't reconnect and will error out.  This is kind of unavoidable with the ephemeral node architecture and is one of the motivations for switching to a Unix domain socket instead of ephemeral Receptor-protocol nodes.

ansible_runner/receptor_plugin.py
```diff
@@ -0,0 +1,171 @@
...
+            elif message.payload:
+                content = json.loads(message.payload.readall())
+                c_header = content[0]
+                c_type = c_header['type']
+                if c_type == 'event':
+                    data = content[1]
+                    if event_handler:
+                        event_handler(data)
+                    if 'stdout' in data and data['stdout']:
+                        print(data['stdout'])
```
ghjm, line 66: Need to do something here better than just `print` ... we shouldn't be printing to stdout if we're being called via the Python API.

ansible_runner/receptor_plugin.py
ghjm: Worked around this by changing from `tarfile` to `zipfile`.  Oddly, the Python `zipfile` class has built-in protection against these traversal attacks, even though `tarfile` does not.

ghjm:
This pull request now depends on https://github.com/project-receptor/receptor/pull/221, which should be merged first.

docs/receptor.rst
```diff
@@ -0,0 +1,51 @@
+.. _receptor_plugin:
```
matburt: I don't see how this would be reachable from the docs if it isn't at least added to `index.rst`

Should probably also add a note here that this would require installing Runner on the remote Receptor node also.

docs/receptor.rst
```diff
@@ -0,0 +1,51 @@
...
+
+Remote job execution via Receptor
+=================================
+
+`Receptor <http://www.github.com/ansible/receptor>`_ is a system for remotely executing jobs and returning results,
```
matburt: https://github.com/project-receptor/receptor here

ansible_runner/receptor_plugin.py
```diff
@@ -0,0 +1,174 @@
...
+                break
+            elif message.payload:
+                content = json.loads(message.payload.readall())
+                c_header = content[0]
+                c_type = c_header['type']
+                if c_type == 'event':
+                    data = content[1]
+                    if event_handler:
+                        event_handler(data)
+                    if 'stdout' in data and data['stdout']:
```
matburt, line 64: Only if the suppress ansible output is false?

matburt (commented):
I left some comments on some minor things, having said that:

My general concern with event processing is that there are two paths here that data can take when coming back in over the receptor network...  they can be emitted over the callback interface or bypassed entirely. The run method also doesn't return a Runner object which breaks interface continuity.

What I'd like to see here is when invoking Runner in this way it still populates the result directory and returns the Runner object as an interface to the result information itself.

docs/receptor.rst
ghjm: Not including this in `index.rst` was an oversight.  Will fix.

The doc does say you need Runner on the remote Receptor node, in the section titled `Server Configuration`.

ghjm:
> What I'd like to see here is when invoking Runner in this way it still populates the result directory and returns the Runner object as an interface to the result information itself.

Yes, I agree 100% with this approach.  But I'd like to make the Unix socket change to Receptor first, so I'm not writing too much Runner code that has to be rewritten after that change to Receptor.  Can we merge this in its current state of implementing only the `interface.run()` method, which is enough for the CLI to work?  Maybe call it a tech preview of Receptor-in-Runner, with a fuller implementation to follow?

ansible_runner/receptor_plugin.py
ghjm: The suppress ansible output option is only on the runner config object - it isn't a parameter to the run() method.  Since the present Receptor implementation only supports the run() method, it can't do anything with this.  (See my other comment.)

docs/receptor.rst
ghjm: Fixed

matburt:
Yep, sounds good to me!

matburt (approved):

Issue #435, opened 2020-03-26
callback plugin broken for older versions of anisble


kdelee:
##### ISSUE TYPE
 - Bug Report

##### SUMMARY

Running jobs w/ virtual environments in awx using older ansible versions is broken.
Symptom that tests are picking up is no `runner_on_*` events.

If you look at the stdout of the jobs run in environments using ansible 2.7.15, 2.6.1, 2.5.6, 2.4.6, or 2.3.3 (probably others), you see

```
[WARNING]: Failure using method (v2_runner_on_ok) in callback plugin
(<ansible.plugins.callback.awx_display.CallbackModule 
```

![image](https://user-images.githubusercontent.com/29239279/77662115-d6807b80-6f51-11ea-9d84-bc91f5891175.png)

ryanpetrello:
Looking at this now, was introduced by: https://github.com/ansible/ansible-runner/pull/405

ryanpetrello:
This is why it's broken in older Ansible versions: https://github.com/ansible/ansible-runner/pull/405/files#r398654057

PR #436, opened 2020-03-26
fix a bug that breaks the callback plugin in older versions of ansible

ryanpetrello:
tracking start/end times accurately relies on v2_runner_on_start, which
was introduced in Ansible 2.8 (and doesn't exist in older versions of
Ansible)

see: https://github.com/ansible/ansible-runner/issues/435

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,11 @@
 Changelog
 ---------
 
+1.4.6 (2020-03-26)
+++++++++++++++++++
+- Fixed a bug that broke Ansible playbook execution prior to version 2.8 of
+  Ansible.
+
 1.4.5 (2020-03-19)
 ++++++++++++++++++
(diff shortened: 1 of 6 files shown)


ryanpetrello:
The implications of this change are that we can't accurately track `start` and `duration` prior to 2.8:

image

(but falling back to `null` is better than exploding)

matburt (approved):

wenottingham (approved):
ACK assuming tests pass

kdelee (approved):
I have tested w/ 2.7.15 which had same symptoms of other older versions. Now testing w/ 2.6, 2.5, 2.4, and 2.3 but prognosis is good

kdelee:
@ryanpetrello testing 2.7, 2.6, 2.5, 2.4, and 2.3. All working now

matburt:
recheck

PR #440, opened 2020-04-03
Suggestion to add extravars_file

mnecas:
Suggestion to add extravars_file, which would allow parallel execution of playbook so the extravars file would not overwrite itself.
I know about that the `self.extra_vars` can be a file, but ansible-runner-service expects extravars as JSON so it always needs to be a dictionary.

Another solution for parallel execution in ansible-runner-service. Which forces to use command extra vars instead of file.  
https://github.com/ansible/ansible-runner-service/pull/48

WDYT @matburt? 
This is just a suggestion, if you think is a good use, I'll add test and update docs.

cc @mwperina

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -75,7 +75,7 @@ class RunnerConfig(object):
                  private_data_dir=None, playbook=None, ident=uuid4(),
                  inventory=None, roles_path=None, limit=None, module=None, module_args=None,
                  verbosity=None, quiet=False, json_mode=False, artifact_dir=None,
-                 rotate_artifacts=0, host_pattern=None, binary=None, extravars=None, suppress_ansible_output=False,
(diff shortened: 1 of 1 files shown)


matburt:
ansible-runner itself allows parallel execution and supply an extra var file is directly supported from within the private-data-dir.

I'll leave this open in the interest of supply extra vars file in the same basic way where giving something like a full @-path as an extra var file *should* work.

matburt:
(replacing my original comment)

ansible-runner itself allows parallel execution and supply an extra var file is directly supported from within the private-data-dir.  I'd rather take an @-var file path in the current extra_vars parameter... if it starts with a file treat it as a path to a vars file and pass that through to ansible. If not, process it as it's doing now.

This also will need some tests.

mnecas:
Thanks for the info.

Issue #438, opened 2020-04-02
Extra variables new line formatting


mnecas:
Hi @matburt,
found an issue with extra variables. When using `\n` in extra variable ansible interprets it as space instead of the new line and `\\n` is a new line.
I was trying to debug it as much as possible but ended up on `pexpect`.

ansible-runner version: python3-ansible-runner-1.4.5-1

## Steps to reproduce it:
**Create playbook test.yml**
```yaml
- name: test
  hosts: localhost
  connection: local
  gather_facts: true
  tasks:
    - name: Test
      copy:
        content: "{{ resp }}"
        dest: "/tmp/testfile"
```

With ansible-runner-service:
```
curl https://127.0.0.1:50001/api/v1/playbooks/test.yml \
--insecure \
-H "Content-Type: application/json" \
-d '{"resp":"test0\ntest1\\ntest2"}'
```

With python:
```python
command = ['ssh-agent', 'sh', '-c', 'ansible-playbook -e \'{"resp":"test0\\ntest1\ntest2"}\' test.yml']
pexpect.spawn(command[0], command[1:], encoding='utf-8').expect("ansible") 
```
result in the file
```
test0
test1 test2
```

## Suggested workaround:
https://github.com/ansible/ansible-runner/blob/devel/ansible_runner/runner_config.py#L402
add `.replace("\n","\\n")`

mnecas:
cc @mwperina

matburt:
This is showing mixed newline escaping. On one hand (and it's reversed between the curl and python command example) you have `\n` between two elements and on the other you have `\\n`. I would fully expect that to yield different behavior so I don't specifically agree with the workaround.

I'll leave this open to see if anyone else runs into this. I suspect that most who use complex extra vars mostly use the `-e @file` mechanism. If there is more feedback on how escaped newlines are treated I'm happy to take it. I probably wouldn't expect it to be treated the same.

AlanCoding:
Looks like that was resolved in the linked PR, let me know if I'm wrong

PR #444, opened 2020-04-14
Serialize python objects before adding them to extra_vars_list

rabi:
This would ensure that any python type is handled properly.
Without that dictionaries with multiline strings end up in
incorrect format.

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -16,6 +16,7 @@
 # specific language governing permissions and limitations
 # under the License.
 #
+import json
 import logging
 import os
 import pexpect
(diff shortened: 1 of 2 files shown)


rabi:
@matburt @mnecas Would appreciate if someone can check this? We're kind of blocked. As a WA we're now forced to write the extravars to a file by default and use it.

mnecas (commented):
LGTM

mnecas:
I have created an issue and gave there a different solution but IMHO this is a way better solution. https://github.com/ansible/ansible-runner/issues/438

matburt (approved):

PR #445, opened 2020-04-22
[KLUDGE] Make it seem like Ansible succeeds if there is nothing to do

domq:
This is a workaround to https://github.com/ansible/awx/issues/2893. I would understand if you wanted to push into the branch ☺

### DIFF
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -234,6 +234,8 @@ class Runner(object):
             self.status_callback('canceled')
         elif child.exitstatus == 0 and not self.timed_out:
             self.status_callback('successful')
+        elif "Specified hosts and/or --limit does not match any hosts" in self.stdout.read():
+            self.status_callback('successful')
         elif self.timed_out:
             self.status_callback('timeout')
         else:


ansible_runner/runner.py
```diff
@@ -234,6 +234,8 @@ def _decode(x):
             self.status_callback('canceled')
         elif child.exitstatus == 0 and not self.timed_out:
             self.status_callback('successful')
+        elif "Specified hosts and/or --limit does not match any hosts" in self.stdout.read():
+            self.status_callback('successful')
```
ryanpetrello, line 238: Er...this seems like a _failure_ to me - you've specified hosts that don't match anything in the inventory, so the playbook _didn't run_.

domq: I think that besides the mathematical question / koan (i.e. what is the AND of an empty list of Booleans?), there is a practical benefit to doing things this way.  For instance, when running ansible-runner within an Ansible Tower job in a container group that has more members than the inventory size (e.g. when inventory size is 1 while doing a canary run), some containers are bound to receive no work; and we want them to show up green, not red. See [this thread](https://github.com/ansible/awx/issues/2893#issuecomment-518424030) for the background.

AlanCoding:
_If_ something like this is done (and that's still a big if), I would defer to the consensus positions articulated in https://github.com/ansible/ansible/issues/42107

Specifically the linked gist https://gist.github.com/jjshoe/ace3070906e5bc5cc432, that would directly replace this, with the text wording ambiguity other hazards. In other words, that logic could be put into the ansible-runner callback plugin. Then the fail-on-no-hosts thing becomes an ansible-runner setting somehow... exactly how not sure.

matburt:
No, we're not going to do this.

PR #446, opened 2020-04-29
WIP: ability to run jobs in containers

shanemcd:


kdelee:
Changes requested:
---------------------------

- [ ] If user has `containerized: true` but not able to launch with `container_runtime`, fail gracefully. have `containerized` internally default to `false`
- [ ] If user does not set `container_runtime` try to use first `podmand` then `docker` if neither launch, fail gracefully
- [ ] `container_name` need setting for settings file for container name which is assumed to include tag and/or registry name (FQContainerName). Default to `ansible/ansible-runner`
- [ ] ????? `container_pull` default true because this is what docker will do for us. For some use cases people may want to make sure not to pull before launching, using whatever is available or failing if not locally available. Not sure if we want to deal w/ this at the runner level or not.
- [ ] `container_volume_mounts` is a settings file option but seems to have some problems around whitspace. Confirm it can take multiple volumes to mount. Additionally right now only supports "symmetric" mounts, e.g. `"/tmp:/tmp" but is common docker use case to mount different local dir to dir in container. As a minimum need better documention of option and whitespace handling
- [ ] add cli flag for `containerized`
- [ ] add cli flag for `container-runtime`
- [ ] add cli flag for `container-name`
- [ ] add cli flag for `container-volume-mounts`
- [ ] ???? if add `container_pull` option then add `container-pull` cli option
- [ ] add env var for `RUNNER_CONTAINERIZED`
- [ ] add env var for `RUNNER_CONTAINER_RUNTIME`
- [ ] add env var for `RUNNER_CONTAINER_NAME`
- [ ] add env var for `RUNNER_CONTAINER_VOLUME_MOUNTS`
- [ ] add env var for `RUNNER_CONTAINER_PORTS`
- [ ] ???? if add `container_pull` option then `RUNNER_CONTAINER_PULL`
- [ ] ensure all internal methods accept all container options
- [ ] if container does not have whatever we need to call ansible-playbook, handle error gracefully and place something in the artifacts file and not just error to stdout so that when user that consumes artifacts to find results, they have evidence that helps them find out what happened.



Test case ideas:
----------------------
Accept settings + Run Job + Artifacts are present:
  - configured w/ settings file
  - configured w/ env vars
  - configured w/ cli options
  - use a playbook that looks at facts that can tell us if in a container, and that fails if it is not in a container
  - check that job result in the tests

Throws error if `containerized` is true but can't use container runtime (use foobar):
  - use easiest setting option

Throws error if can't find container and can't pull it:
  - use easiest setting option

Test can launch job w/ podman and docker.
  - need to investigate zuul capability

Given a container that does exist and is pullable, but does not provide expected interface for runner to run a job in (hello-world container for example), what happens? Do we have error handling?

kdelee:
# in-PR Bug tracking
----------------------------
- [ ] seems like container is failing to launch when I set `container_volume_mounts:/tmp` but does not throw error and not sure whats up -- this gets back to erroring if can't launch w/ the container runtime specified and `containerized: true`
- [ ] when job DOES launch in container, seems to not respect to correct private data dir, at least in a poetry dev env. If I have in settings file `containerized: true` and `container_runtime: docker` it DOES launch in the container, so must have found the settings file in the correct data dir, but then once inside container, is picking up old version of `demo` private data dir, so I'm thinking it is getting it from the copy of the `demo` dir in the poetry env????

PR #448, opened 2020-05-04
Bump to 1.4.7

rabi:
- Includes fix to serialize extra_vars.

Also chages the rpm package name to python-ansible-runner to avoid
NVR mismatches.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,10 @@
 Changelog
 ---------
 
+1.4.7 (2020-05-04)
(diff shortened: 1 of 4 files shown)


rabi:
Looks like latest virtualenv is borked https://github.com/pypa/virtualenv/issues/1810

gaborbernat:
Released a fix via https://pypi.org/project/virtualenv/20.0.20/

matburt:
You should ping me directly about a Runner release.

PR #453, opened 2020-05-11
switch from pipenv to poetry

maxamillion:


Makefile
```diff
@@ -69,21 +70,26 @@ clean:
...
 
 sdist: dist/$(NAME)-$(VERSION).tar.gz
 
+# Generate setup.py transiently for the sdist so we don't have to deal with
+# packaging poetry as a RPM for rpm build time dependencies.
 dist/$(NAME)-$(VERSION).tar.gz:
-	$(DIST_PYTHON) setup.py sdist
+	$(PYTHON) packaging/poetry-gen-setup.py
+	poetry build -f sdist
+	rm -f setup.py
```
shanemcd: Should we drop this line now that it's checked into the repo?

maxamillion: @shanemcd yes, good catch. Thanks :)

shanemcd (approved):
😎

PR #470, opened 2020-06-25
Containerized Playbook Runs

jladdjr:
Branching off of https://github.com/ansible/ansible-runner/pull/446

Depends-On: https://github.com/ansible/project-config/pull/533

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,9 @@
 *,cover
 .venv
 .env
+poetry.lock
+pytestdebug.log
+test/integration/containerized/priv_data/artifacts/
+new_logfile
+otherartifacts/
+test/integration/project/
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,6 @@
 *,cover
 .venv
 .env
-poetry.lock
 pytestdebug.log
 test/integration/containerized/priv_data/artifacts/
 new_logfile
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 
 /test/integration/artifacts
 /demo/artifacts
+/demo/daemon.log
 /docs/_build
 /.tox
 /dist
@@ -17,12 +18,9 @@
 /*.egg-info
 *.py[c,o]
 .pytest_cache
+pytestdebug.log
+test/integration/containerized/priv_data/artifacts/
 .coverage
 *,cover
 .venv
 .env
-pytestdebug.log
-test/integration/containerized/priv_data/artifacts/
-new_logfile
-otherartifacts/
-test/integration/project/
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,9 +29,6 @@ ENV LC_ALL=en_US.UTF-8
 ENV RUNNER_BASE_COMMAND=ansible-playbook
 ENV HOME=/runner
 
-ENV ANSIBLE_STDOUT_CALLBACK=awx_display
-ENV ANSIBLE_CALLBACK_PLUGINS=/usr/lib/python3.6/site-packages/ansible_runner/callbacks
-
 WORKDIR /runner
 
 ENTRYPOINT ["entrypoint"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@ ADD demo/inventory /runner/inventory
 # Install Ansible and Runner
 ADD https://releases.ansible.com/ansible-runner/ansible-runner.el8.repo /etc/yum.repos.d/ansible-runner.repo
 RUN dnf install -y epel-release && \
-    dnf install -y ansible-runner python3-pip sudo rsync openssh-clients sshpass glibc-langpack-en && \
+    dnf install -y ansible-runner python3-pip sudo rsync openssh-clients sshpass glibc-langpack-en git && \
     alternatives --set python /usr/bin/python3 && \
     pip3 install ansible-base && \
     chmod +x /bin/tini /bin/entrypoint && \
@@ -21,15 +21,16 @@ RUN mkdir -p /runner/inventory /runner/project /runner/artifacts /runner/.ansibl
 	chmod -R g+w /runner && chgrp -R root /runner && \
 	chmod g+w /etc/passwd
 
-VOLUME /runner/inventory
-VOLUME /runner/project
-VOLUME /runner/artifacts
+VOLUME /runner
 
 ENV LANG=en_US.UTF-8
 ENV LANGUAGE=en_US:en
 ENV LC_ALL=en_US.UTF-8
 ENV RUNNER_BASE_COMMAND=ansible-playbook
 ENV HOME=/runner
+
+ENV ANSIBLE_STDOUT_CALLBACK=awx_display
+ENV ANSIBLE_CALLBACK_PLUGINS=/usr/lib/python3.6/site-packages/ansible_runner/callbacks
 
 WORKDIR /runner
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,7 +10,7 @@ ADD demo/inventory /runner/inventory
 ADD https://releases.ansible.com/ansible-runner/ansible-runner.el8.repo /etc/yum.repos.d/ansible-runner.repo
 RUN dnf install -y epel-release && \
     dnf install -y ansible-runner python3-pip sudo rsync openssh-clients sshpass glibc-langpack-en git && \
-    pip3 install git+https://github.com/ansible/ansible.git && \
+    pip3 install ansible && \
     chmod +x /bin/tini /bin/entrypoint && \
     rm -rf /var/cache/dnf
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,8 +10,7 @@ ADD demo/inventory /runner/inventory
 ADD https://releases.ansible.com/ansible-runner/ansible-runner.el8.repo /etc/yum.repos.d/ansible-runner.repo
 RUN dnf install -y epel-release && \
     dnf install -y ansible-runner python3-pip sudo rsync openssh-clients sshpass glibc-langpack-en git && \
-    alternatives --set python /usr/bin/python3 && \
-    pip3 install ansible-base && \
+    pip3 install git+https://github.com/ansible/ansible.git && \
     chmod +x /bin/tini /bin/entrypoint && \
     rm -rf /var/cache/dnf
 
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -502,17 +502,17 @@ def main(sys_args=None):
     )
 
     playbook_group.add_argument(
-        "--container-volume-mounts",
+        "--container-volume-mount",
         dest="container_volume_mounts",
-        nargs='*',
-        help="List of bind mounts (in the form 'host_dir:/container_dir)'"
+        action='append',
+        help="Bind mounts (in the form 'host_dir:/container_dir)'. Can be used more than once to create multiple bind mounts."
     )
 
     playbook_group.add_argument(
-        "--container-options",
+        "--container-option",
         dest="container_options",
-        nargs='*',
-        help="List of container options to pass to execution engine."
+        action='append',
+        help="Container options to pass to execution engine. Can be used more than once to send multiple options."
     )
 
     playbook_group.add_argument(
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -442,22 +442,23 @@ def main(sys_args=None):
         "--process-isolation",
         dest="process_isolation",
         action="store_true",
-        help="limits what directories on the filesystem the playbook run "
-             "has access to, defaults to /tmp (default=False)"
+        help="Isolate execution. Two methods are supported: (1) using a sandbox (e.g. bwrap) which will "
+             "by default restrict access to /tmp or (2) using a container engine (e.g. podman or docker) "
+             "to execute **Ansible**. (default=False)"
     )
 
     playbook_group.add_argument(
         "--process-isolation-executable",
         dest="process_isolation_executable",
         default="bwrap",
-        help="process isolation executable that will be used. (default=bwrap)"
+        help="Process isolation executable or container engine used to isolate execution. (default=bwrap)"
     )
 
     playbook_group.add_argument(
         "--process-isolation-path",
         dest="process_isolation_path",
         default="/tmp",
-        help="path that an isolated playbook run will use for staging. "
+        help="Path that an isolated playbook run will use for staging. "
              "(default=/tmp)"
     )
 
@@ -465,7 +466,7 @@ def main(sys_args=None):
         "--process-isolation-hide-paths",
         dest="process_isolation_hide_paths",
         nargs='*',
-        help="list of paths on the system that should be hidden from the "
+        help="List of paths on the system that should be hidden from the "
              "playbook run (default=None)"
     )
 
@@ -473,7 +474,7 @@ def main(sys_args=None):
         "--process-isolation-show-paths",
         dest="process_isolation_show_paths",
         nargs='*',
-        help="list of paths on the system that should be exposed to the "
+        help="List of paths on the system that should be exposed to the "
              "playbook run (default=None)"
     )
 
@@ -481,16 +482,37 @@ def main(sys_args=None):
         "--process-isolation-ro-paths",
         dest="process_isolation_ro_paths",
         nargs='*',
-        help="list of paths on the system that should be exposed to the "
+        help="List of paths on the system that should be exposed to the "
              "playbook run as read-only (default=None)"
     )
 
     playbook_group.add_argument(
         "--directory-isolation-base-path",
         dest="directory_isolation_base_path",
-        help="copies the project directory to a location in this directory "
+        help="Copies the project directory to a location in this directory "
              "to prevent multiple simultaneous executions from conflicting "
              "(default=None)"
+    )
+
+    playbook_group.add_argument(
+        "--container-image",
+        dest="container_image",
+        default="ansible/ansible-runner",
+        help="Container image to use when running an ansible task"
+    )
+
+    playbook_group.add_argument(
+        "--container-volume-mounts",
+        dest="container_volume_mounts",
+        nargs='*',
+        help="List of bind mounts (in the form 'host_dir:/container_dir)'"
+    )
+
+    playbook_group.add_argument(
+        "--container-options",
+        dest="container_options",
+        nargs='*',
+        help="List of container options to pass to execution engine."
     )
 
     playbook_group.add_argument(
@@ -527,33 +549,6 @@ def main(sys_args=None):
         "--resource-profiling-results-dir",
         dest='resource_profiling_results_dir',
         help="Directory where profiling data files should be saved. Defaults to None (profiling_data folder under private data dir is used in this case).")
-
-    playbook_group.add_argument(
-        "--containerized",
-        dest='containerized',
-        action="store_true",
-        help="Indicates ansible task should be executed using an execution environment")
-
-    playbook_group.add_argument(
-        "--container-runtime",
-        dest="container_runtime",
-        default="podman",
-        help="Container engine to use when running an ansible task"
-    )
-
-    playbook_group.add_argument(
-        "--container-image",
-        dest="container_image",
-        default="ansible/ansible-runner",
-        help="Container image to use when running an ansible task (see --containerized)"
-    )
-
-    playbook_group.add_argument(
-        "--container-volume-mounts",
-        dest="container_volume_mounts",
-        default=None,
-        help="Comma-separated list of bind mounts (e.g. 'host_dir:/container_dir,host_dir2:/container_dir2')"
-    )
 
     if len(sys.argv) == 1:
         parser.print_usage()
@@ -641,6 +636,9 @@ def main(sys_args=None):
                                    process_isolation_hide_paths=args.process_isolation_hide_paths,
                                    process_isolation_show_paths=args.process_isolation_show_paths,
                                    process_isolation_ro_paths=args.process_isolation_ro_paths,
+                                   container_image=args.container_image,
+                                   container_volume_mounts=args.container_volume_mounts,
+                                   container_options=args.container_options,
                                    directory_isolation_base_path=args.directory_isolation_base_path,
                                    resource_profiling=args.resource_profiling,
                                    resource_profiling_base_cgroup=args.resource_profiling_base_cgroup,
@@ -648,10 +646,6 @@ def main(sys_args=None):
                                    resource_profiling_memory_poll_interval=args.resource_profiling_memory_poll_interval,
                                    resource_profiling_pid_poll_interval=args.resource_profiling_pid_poll_interval,
                                    resource_profiling_results_dir=args.resource_profiling_results_dir,
-                                   containerized=args.containerized,
-                                   container_runtime=args.container_runtime,
-                                   container_image=args.container_image,
-                                   container_volume_mounts=args.container_volume_mounts,
                                    limit=args.limit,
                                    via_receptor=args.via_receptor,
                                    receptor_peer=args.receptor_peer,
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -528,6 +528,33 @@ def main(sys_args=None):
         dest='resource_profiling_results_dir',
         help="Directory where profiling data files should be saved. Defaults to None (profiling_data folder under private data dir is used in this case).")
 
+    playbook_group.add_argument(
+        "--containerized",
+        dest='containerized',
+        action="store_true",
+        help="Indicates ansible task should be executed using an execution environment")
+
+    playbook_group.add_argument(
+        "--container-runtime",
+        dest="container_runtime",
+        default="podman",
+        help="Container engine to use when running an ansible task"
+    )
+
+    playbook_group.add_argument(
+        "--container-image",
+        dest="container_image",
+        default="ansible/ansible-runner",
+        help="Container image to use when running an ansible task (see --containerized)"
+    )
+
+    playbook_group.add_argument(
+        "--container-volume-mounts",
+        dest="container_volume_mounts",
+        default=None,
+        help="Comma-separated list of bind mounts (e.g. 'host_dir:/container_dir,host_dir2:/container_dir2')"
+    )
+
     if len(sys.argv) == 1:
         parser.print_usage()
         print_common_usage()
@@ -621,6 +648,10 @@ def main(sys_args=None):
                                    resource_profiling_memory_poll_interval=args.resource_profiling_memory_poll_interval,
                                    resource_profiling_pid_poll_interval=args.resource_profiling_pid_poll_interval,
                                    resource_profiling_results_dir=args.resource_profiling_results_dir,
+                                   containerized=args.containerized,
+                                   container_runtime=args.container_runtime,
+                                   container_image=args.container_image,
+                                   container_volume_mounts=args.container_volume_mounts,
                                    limit=args.limit,
                                    via_receptor=args.via_receptor,
                                    receptor_peer=args.receptor_peer,
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -442,16 +442,16 @@ def main(sys_args=None):
         "--process-isolation",
         dest="process_isolation",
         action="store_true",
-        help="Isolate execution. Two methods are supported: (1) using a sandbox (e.g. bwrap) which will "
-             "by default restrict access to /tmp or (2) using a container engine (e.g. podman or docker) "
-             "to execute **Ansible**. (default=False)"
+        help="Isolate execution. Two methods are supported: (1) using a container engine (e.g. podman or docker) "
+             "to execute **Ansible**. (2) using a sandbox (e.g. bwrap) which will by default restrict access to /tmp "
+             "(default=False)"
     )
 
     playbook_group.add_argument(
         "--process-isolation-executable",
         dest="process_isolation_executable",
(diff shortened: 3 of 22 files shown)


jladdjr:
Copying over @kdelee's testing checklist from https://github.com/ansible/ansible-runner/pull/446#issuecomment-649684538:

Changes requested:
---------------------------

- [x] If user has `containerized: true` but not able to launch with `container_runtime`, fail gracefully. 
- [ ] ~.. and have `containerized` internally default to `false`~ (jladd: not sure that if user requests containerized execution that we want to switch to a regular run if we don't find the container engine)
- [ ] ~If user does not set `container_runtime` try to use first `podmand` then `docker` if neither launch, fail gracefully~ (update: jladd: I'm a little iffy on this. think that we want to avoid surprising a user by switching out the container engines)
- [x] `container_name` need setting for settings file for container name which is assumed to include tag and/or registry name (FQContainerName). Default to `ansible/ansible-runner`
- [ ] ????? `container_pull` default true because this is what docker will do for us. For some use cases people may want to make sure not to pull before launching, using whatever is available or failing if not locally available. Not sure if we want to deal w/ this at the runner level or not.
- [x] `container_volume_mounts` is a settings file option but seems to have some problems around whitspace. Confirm it can take multiple volumes to mount. Additionally right now only supports "symmetric" mounts, e.g. `"/tmp:/tmp" but is common docker use case to mount different local dir to dir in container. As a minimum need better documention of option and whitespace handling
- [x] add cli flag for `containerized`
- [x] add cli flag for `container-runtime`
- [x] add cli flag for ~`container-name`~ `container-image`
- [x] add cli flag for `container-volume-mounts`
- [ ] ???? if add `container_pull` option then add `container-pull` cli option
~- [ ] add env var for `RUNNER_CONTAINERIZED`~
~- [ ] add env var for `RUNNER_CONTAINER_RUNTIME`~
~- [ ] add env var for `RUNNER_CONTAINER_NAME`~
~- [ ] add env var for `RUNNER_CONTAINER_VOLUME_MOUNTS`~
~- [ ] add env var for `RUNNER_CONTAINER_PORTS`~
- [ ] ???? if add `container_pull` option then `RUNNER_CONTAINER_PULL`
- [x] ensure all internal methods accept all container options
- [x] if container does not have whatever we need to call ansible-playbook, handle error gracefully and place something in the artifacts file and not just error to stdout so that when user that consumes artifacts to find results, they have evidence that helps them find out what happened.



Test case ideas:
----------------------
Accept settings + Run Job + Artifacts are present:
  - configured w/ settings file ✅ 
  - ~configured w/ env vars~
  - configured w/ cli options ✅ 
  - use a playbook that looks at facts that can tell us if in a container, and that fails if it is not in a container ✅ 
  - check that job result in the tests ✅ 

Throws error if `containerized` is true but can't use container runtime (use foobar):
  - use easiest setting option

Throws error if can't find container and can't pull it:
  - use easiest setting option

Test can launch job w/ podman and docker.
  - need to investigate zuul capability (UPDATE: request to add docker / podman to zuul [here](https://github.com/ansible/ansible-runner/issues/480))

Given a container that does exist and is pullable, but does not provide expected interface for runner to run a job in (hello-world container for example), what happens? Do we have error handling?

jladdjr:
# in-PR Bug tracking
----------------------------
(originally from https://github.com/ansible/ansible-runner/pull/446#issuecomment-649808220)

- [ ] seems like container is failing to launch when I set `container_volume_mounts:/tmp` but does not throw error and not sure whats up -- this gets back to erroring if can't launch w/ the container runtime specified and `containerized: true`
- [ ] when job DOES launch in container, seems to not respect to correct private data dir, at least in a poetry dev env. If I have in settings file `containerized: true` and `container_runtime: docker` it DOES launch in the container, so must have found the settings file in the correct data dir, but then once inside container, is picking up old version of `demo` private data dir, so I'm thinking it is getting it from the copy of the `demo` dir in the poetry env????
- [x] not 100% if this is a "bug" but it is inconvenient that some files created by runner when running inside a container end up being owned by root. This may have something to do w/ how I have docker/podman setup so would like @jladdjr see if this happens as well (edit by @kdelee ). Files include the`$PRIVATE_DATA_DIR/.ansible` folder and `$PRIVATE_DATA_DIR/artifacts/<hash>/job_events/<hash>-partial.json`

ansible_runner/runner_config.py
```diff
@@ -84,7 +84,7 @@ def __init__(self,
                  resource_profiling_results_dir=None,
                  tags=None, skip_tags=None, fact_cache_type='jsonfile', fact_cache=None, ssh_key=None,
                  project_dir=None, directory_isolation_base_path=None, envvars=None, forks=None, cmdline=None, omit_event_data=False,
-                 only_failed_event_data=False):
+                 only_failed_event_data=False, containerized=False, container_runtime=False):
```
AlanCoding: Is it planned to add this to the CLI? I would like to get more documentation and schema around the options. See the ansible-builder PR https://github.com/ansible/ansible-builder/pull/16

That adds:

```
$ ansible-builder build --help
usage: ansible-builder build [-h] [-t TAG] [-f FILENAME] [-b BASE_IMAGE]
                             [-c BUILD_CONTEXT]
                             [--container-runtime CONTAINER_RUNTIME]

optional arguments:
  -h, --help            show this help message and exit
  -t TAG, --tag TAG     The name for the container being built.
  -f FILENAME, --file FILENAME
                        The definiton of the execution environment.
  -b BASE_IMAGE, --base-image BASE_IMAGE
                        The parent image for the execution environment.
  -c BUILD_CONTEXT, --context BUILD_CONTEXT
                        The directory to use for the build context. Defaults
                        to $PWD/context.
  --container-runtime CONTAINER_RUNTIME
                        Specifies which container runtime to use.
```

Where the use looks like:

```
$ ansible-builder build -c alan/bc/ -f alan/alan_ee.yml --container-runtime=docker
```

I would like to see some example commands for runner, and also start syncing the schema of these options between the two projects so that they are the same interface.

I could see the `--tag` option coming into play here.

Dockerfile
```diff
@@ -21,16 +21,17 @@ RUN mkdir -p /runner/inventory /runner/project /runner/artifacts /runner/.ansibl
...
+VOLUME /runner
 
 ENV LANG=en_US.UTF-8
 ENV LANGUAGE=en_US:en
 ENV LC_ALL=en_US.UTF-8
 ENV RUNNER_BASE_COMMAND=ansible-playbook
 ENV HOME=/runner
 
+ENV ANSIBLE_STDOUT_CALLBACK=awx_display
+ENV ANSIBLE_CALLBACK_PLUGINS=/usr/lib/python3.6/site-packages/ansible_runner/callbacks
```
shanemcd: @AlanCoding Didnt you have some thoughts about putting the callback plugin into a collection? If we did that, then we wouldn't need this, right?

AlanCoding: yes, I would absolutely like to do that

I went over what I wanted to do with Shane, and didn't see any of this container-isolation stuff as something which would fundamentally change a collection conversion.

This line, however, is something that I don't think we need. Down below, I see that `ansible-runner` gets called. If that's the entry point, then we shouldn't be setting anything like this beforehand.

That gets set in the runner code itself, only for the `ansible-playbook` process

https://github.com/ansible/ansible-runner/blob/f73705bf9b09424c9e5353d7f5119147e844eaa5/ansible_runner/runner_config.py#L198

you don't need to have it set in the process where `ansible-runner` is invoked. So just delete these 2 lines.

AlanCoding:
github doesn't seem to automatically link this, so dropping link https://github.com/jladdjr/ansible-runner/pull/1

jladdjr:
recheck

Dockerfile
```diff
@@ -9,7 +9,7 @@ ADD demo/inventory /runner/inventory
 # Install Ansible and Runner
 ADD https://releases.ansible.com/ansible-runner/ansible-runner.el8.repo /etc/yum.repos.d/ansible-runner.repo
 RUN dnf install -y epel-release && \
-    dnf install -y ansible-runner python3-pip sudo rsync openssh-clients sshpass glibc-langpack-en && \
+    dnf install -y ansible-runner python3-pip sudo rsync openssh-clients sshpass glibc-langpack-en git && \
     alternatives --set python /usr/bin/python3 && \
     pip3 install ansible && \
```
AlanCoding, line 13: @jladdjr I have https://github.com/ansible/ansible-runner/pull/473 open

I'm only committed to working in conjunction with your branch, not with devel, and I don't urgently need that change.

So I can close that if you get the change in here.

test/integration/containerized/priv_data/project/test-container.yml
```diff
@@ -0,0 +1,38 @@
...
+  gather_facts: true
+  vars:
+    cgroup_search_terms: docker\|lxc\|podman\|oci
+    is_container: false
+    is_podman: false
+    is_docker: false
+  tasks:
+    - debug: msg="Test if we are in a container!"
+
+    - name: 'Look container env var defined in /proc/1/environ'
```
kdelee: `Look for container....`

test/integration/containerized/priv_data/project/test-container.yml
```diff
@@ -0,0 +1,38 @@
...
+      shell: "grep 'container=' /proc/1/environ"
+      register: container_proc_environ
+      ignore_errors: true
+
+    - name: 'Look container info defined in /proc/1/cgroup'
```
kdelee: `Look for container....`

test/integration/containerized/test_cli_containerized.py
```diff
@@ -0,0 +1,12 @@
...
+
+HERE = os.path.abspath(os.path.dirname(__file__))
+
+
+def test_module_run(cli):
+    cli(['-m', 'ping','--hosts', 'localhost', 'run', os.path.join(HERE, 'priv_data')])
+
+
+def test_playbook_run(cli):
+    cli(['run', os.path.join(HERE,'priv_data'), '-p', 'test-container.yml'])
```
kdelee, line 20: :bulb: I want to add a test that runs a playbook that fails and confirms we get non-zero exit code as I could see running via the container maybe messing w/ our exit code what do you think

jladdjr:
recheck

jladdjr:
@kdelee :

jladdjr:
> If user does not set container_runtime try to use first podmand then docker if neither launch, fail gracefully

@kdelee - wondering if we really want this behavior - here's an example of how I think this could lead to unexpected behavior:

1. User has been using to docker
2. .. but switches over to podman
3. When configuring ansible-runner, they specify podman, as this is now their default setup.
4. They misconfigured podman (e.g. have their registries not quite setup right, thought they had built an image, but it's not around, etc.)
5. .. but when runner switches to docker things manage to work.

If I were this user, I think I'd actually prefer things to halt with podman, since that's what I asked for, instead of it switching to another tool, that I may consider to be deprecated.

If we _do_ want this fallback behavior, think we should make the setting `default_container_runtime` to emphasize that the other know runtime will be tried.

@matburt @shanemcd - ^ let me know if you favor one scenario over the other or have any other ideas for how we should handle this

test/integration/containerized/test_cli_containerized.py
jladdjr: @kdelee - good question - I poked around some existing tests and found [this one](https://github.com/jladdjr/ansible-runner/blob/containerized_playbook_runs/test/integration/test_events.py#L14) that already asserts that the run was successful. It also does a sanity check on several events that we expect to see. What do you think about parameterizing that to run w/ and w/out containerization?

I think some of the other tests in that module may also be good candidates for parameterization if we think that running inside a container puts those other aspects of a run at risk. The `test_basic_events` test may give us the most bang for our buck, though.

jladdjr: got `test_basic_events` to work w/ containerization - actually helped me work out some kinks / understand how some things were connected:
https://github.com/jladdjr/ansible-runner/blob/containerized_playbook_runs/test/integration/test_events.py#L14-L55

.gitignore
```diff
@@ -21,3 +21,9 @@
 *,cover
 .venv
 .env
+poetry.lock
```
matburt: poetry.lock needs to be checked in, this is kind-of the point of poetry... to make sure everyone uses the same level of dependencies.

Dockerfile
matburt: If we do that then lets discuss it outside of the context of this PR since it's a pretty major component for Runner itself.

ansible_runner/__main__.py
```diff
@@ -528,6 +528,33 @@ def main(sys_args=None):
...
+
+    playbook_group.add_argument(
+        "--container-image",
+        dest="container_image",
+        default="ansible/ansible-runner",
+        help="Container image to use when running an ansible task (see --containerized)"
+    )
+
+    playbook_group.add_argument(
+        "--container-volume-mounts",
```
matburt: I can see the purpose for volume mounts here but it may be necessary to support arbitrary container parameterization (such as resource usage, network attachment, etc). Either in the form of catchall parameterization or config file.

ansible_runner/runner_config.py
```diff
@@ -170,8 +176,8 @@ def prepare(self):
                                                                                                            self.directory_isolation_path))
                 copy_tree(self.project_dir, self.directory_isolation_path, preserve_symlinks=True)
 
-        self.prepare_inventory()
         self.prepare_env()
+        self.prepare_inventory()
```
matburt, line 189: What precipitates this inversion?

ansible_runner/runner_config.py
```diff
@@ -393,7 +418,11 @@ def generate_ansible_command(self):
             exec_list.append(self.limit)
 
         if self.loader.isfile('env/extravars'):
-            exec_list.extend(['-e', '@{}'.format(self.loader.abspath('env/extravars'))])
+            if self.containerized:
+                extravars_path = '/runner/env/extravars'
```
matburt, line 440: This hard-coding doesn't bring me joy 😆 

We should be able to discover the intended base. I'd like to not assume that it would *always* be `/runner` when running containerized.

setup.cfg
```diff
@@ -18,7 +18,7 @@ exclude=.tox,venv,awx/lib/site-packages,awx/plugins/inventory/ec2.py,awx/plugins
 [flake8]
 max-line-length=160
 ignore=E201,E203,E221,E225,E231,E241,E251,E261,E265,E303,W291,W391,W293,E731,F405
-exclude=.tox,venv,awx/lib/site-packages,awx/plugins/inventory,awx/ui,awx/api/urls.py,awx/main/migrations,awx/main/south_migrations,awx/main/tests/data,node_modules/,awx/projects/,tools/docker,awx/settings/local_*.py,installer/openshift/settings.py,build/,installer/
+exclude=.tox,venv,awx/lib/site-packages,awx/plugins/inventory,awx/ui,awx/api/urls.py,awx/main/migrations,awx/main/south_migrations,awx/main/tests/data,node_modules/,awx/projects/,tools/docker,awx/settings/local_*.py,installer/openshift/settings.py,build/,installer/,packaging/
```
matburt: While we're at it, lets just remove some of these old AWX things that stayed in when I copied this file over.

ansible_runner/interface.py
```diff
@@ -131,6 +131,10 @@ def run(**kwargs):
     :param resource_profiling_memory_poll_interval: Interval (in seconds) between memory polling for determining memory usage (default: 0.25)
     :param resource_profiling_pid_poll_interval: Interval (in seconds) between polling PID count for determining number of processes used (default: 0.25)
     :param resource_profiling_results_dir: Directory where profiling data files should be saved (defaults to profiling_data folder inside private data dir)
+    :param containerized: Indicates ansible task should be executed using an execution environment (default: false)
```
matburt: I wonder if "containerization" is actually the right thing to use here. This would seem to be a modification of the process isolation system and doing it this way doesn't make bwrap and container usage mutually exclusive... which it probably should be.

ansible_runner/runner_config.py
maxamillion: @AlanCoding to end game is for `ansible-builder` and `ansible-runner` to match the CLI interface for this, I have a PR that's waiting for this to merge that will "settle up" on that once I rebase.

ansible_runner/__main__.py
maxamillion: +1 to that, I think trying to wrap each param we want to target for various isolation purposes will become painful to maintain quickly

ansible_runner/interface.py
jladdjr: @matburt That brings up a question that I've had. How much longer will we need to support both container and non-container based execution? Is that something that I could go ahead and drop in this PR?

jladdjr: If we're not going to drop non-container execution, then:

* do we want to make 'containerized' a `process_isolation` / `process_isolation_executable` setting? While containerization is a process isolation mechanism, this seems a little unintuitive.
* or, do we just want to add some logic to disable bwrap when `containerized: True`? I'm guessing this is the middle-ground that we'll end up going with.

ansible_runner/runner_config.py
jladdjr: @matburt - I was looking at this yesterday. We're mapping the private data dir on the host to `/runner` in the container, which is what allows us to hardcode `/runner` as the base. That _seems_ like it should be okay, but maybe that's not actually functionally equivalent. Let me know if you would still consider `/runner` to be hard mapping given the volume mapping.

ansible_runner/runner_config.py
shanemcd: This one was from me... 😄 Basically, `self.inventory` gets set inside of `prepare_inventory` and I kinda needed that to be set (properly) before building the command.

ansible_runner/interface.py
shanemcd: I am definitely not a fan of changing this option to `containerization`. `ansible-runner run --containerized` makes way more sense than `ansible-runner run --containerization`. Am I missing something?

And I think we should just throw an error if you try to use `--process-isolation` at the same time as `--containerized`.

ansible_runner/runner_config.py
matburt: This *is* the private data dir, though and it should be set/fetched from that:

https://github.com/ansible/ansible-runner/blob/devel/ansible_runner/runner_config.py#L88

matburt: Actually, I retract that... because I see the attachment locality of this now. I'm still not sure how I feel about this in the general sense and it might should be based on where we know we are attaching volumes if the person is doing something *other* than what we expected them to do.

jladdjr:
* Refactored the process isolation options to fold in containerization as a possible method.
* Added `container_options` as a catch-all for any other options the user might want to pass to the container engine.

Dockerfile
AlanCoding: Clarification: we should remove these settings, because they are set in the code itself.

test/integration/containerized/priv_data/env/settings
```diff
@@ -0,0 +1,7 @@
+---
+idle_timeout: 60
+job_timeout: 360
+pexpect_timeout: 10
+process_isolation: true
+process_isolation_executable: podman
+container: ansible/ansible-runner
```
AlanCoding: @kdelee this look wrong, it was changed but this isn't the syntax it changed to. Worked for me:

```yaml
---
container_image: ansible-execution-env
process_isolation_executable: docker
process_isolation: true
```

https://github.com/ansible/ansible-builder/pull/43

jladdjr: > @kdelee this look wrong, it was changed but this isn't the syntax it changed to. Worked for me:

I'm having trouble parsing this. The example you posted does reflect the latest options, though.

ansible_runner/interface.py
jladdjr: @shanemcd and I caught up about this. We ended up with these options, which folds containerization into the existing process isolation options, and adds a few options specific to containerization:

```
--process-isolation => {true, false}
--process-isolation-executable => {bwrap, podman, docker}
--container-image => defaults to `ansible/ansible-runner`
--container-volume-mounts => list of host:container dir mappings
--container-options => list of any other options you may want to pass to the execution engine
```

ansible_runner/__main__.py
jladdjr: added `--container-options` as a catch-all for any remaining options the user may want to pass to runner

jladdjr:
@kdelee - dropping the 'create env var' todo's from our list. I thought that env vars were something that runner exposed for most options. What I was (mis)remembering though was that when I did the [perf stats work](https://github.com/ansible/ansible-runner/pull/346/files#diff-c62b92cfe059167de0305220ea16db97R218), I used env vars to talk with the [cgroup_perf_recap plugin](https://docs.ansible.com/ansible/latest/plugins/callback/cgroup_perf_recap.html).

jladdjr:
@shanemcd, @matburt - one of the items that @kdelee and I thought of adding was a flag that would instruct runner not to pull down an image if one wasn't already available. Do you think this would be useful?

If so, I'm not seeing an obvious option for `docker run` that would block pulling an image. I suppose what we could do instead is check `docker images` to see if the image is already available, and stop if one isn't.

jladdjr:
> if container does not have whatever we need to call ansible-playbook, handle error gracefully and place something in the artifacts file and not just error to stdout so that when user that consumes artifacts to find results, they have evidence that helps them find out what happened.

@kdelee - Confirmed that runner provides helpful artifacts in the event that the container is missing deps. Here are the artifacts left behind when ansible itself is missing from the container. All of these are located under `artifacts/<uuid>`.

---

**job_events/1-f240c47b-aa7e-4153-8df2-d83a77df7988.json**
```
{
  "event": "verbose",
  "uuid": "f240c47b-aa7e-4153-8df2-d83a77df7988",
  "counter": 1,
  "stdout": "[FATAL tini (12)] exec ansible-playbook failed: No such file or directory",
  "start_line": 0,
  "end_line": 1,
  "runner_ident": "05725d1d-3149-40eb-a174-11ee1b2c3310"
}
```

**rc**
`127`

**status**
`failed`

**stdout**
`[FATAL tini (12)] exec ansible-playbook failed: No such file or directory`

test/integration/test_events.py
```diff
@@ -6,22 +6,38 @@
...
+        run_args.update({'process_isolation': True,
+                         'process_isolation_executable': 'podman',
+                         'container_image': 'ansible/ansible-runner',
+                         'container_volume_mounts': [f'{tdir}:{tdir}']})
+
     if not is_run_async:
-        r = run(private_data_dir=tdir,
-                inventory=inventory,
-                playbook=playbook)
+        r = run(**run_args)
```
chrismeyersfsu, line 32: Using `**` vs. explicitly listing the key/value params will build up technical debt. We had to pay this dept in AWX when we added runner support to `tasks.py`

jladdjr:
Looking for any final feedback before merge. I _think_ that everything has been covered (apart from adding some more docs, which I'm working on right now), but if I've missed anything let me know!

AlanCoding:
Before we merge this, I think we should figure out what the cause of the warning is:

```
[DEPRECATION WARNING]: 'awx_display' is subclassing DefaultCallback without the
 corresponding doc_fragment. This feature will be removed from ansible-base in 
version 2.14. Deprecation warnings can be disabled by setting 
deprecation_warnings=False in ansible.cfg.
```

That is specific to this PR, and there's a logical leap from what it changed to getting the wrong callback somehow.

AlanCoding:
I re-ran one of my examples with the playbook:

```yaml
---
- hosts: localhost
  gather_facts: false
  connection: local
  tasks:
    - debug: msg="hello"

    - command:
        cmd: printenv
```

This finds the value

```
ANSIBLE_CALLBACK_PLUGINS=/Users/alancoding/Documents/repos/ansible-runner/ansible_runner/callbacks
```

That pretty well confirms the theory, that even though the container sets `ENV` (which it shouldn't), it doesn't matter because the ansible-runner parent process will overwrite that before it calls pexpect to create the subprocess. When it does that, it uses a location on _my_ computer, which obviously doesn't exist in the container.

Even if we get that, I don't see how we run the current ansible-runner tests containerized.

AlanCoding:
Put up a proposed fix:

https://github.com/jladdjr/ansible-runner/pull/3

But again, I think we really need to ask the question of how we test against regressions like this?

ansible_runner/runner_config.py
```diff
@@ -244,12 +264,29 @@ def prepare(self):
         self.env["RUNNER_OMIT_EVENTS"] = str(self.omit_event_data)
         self.env["RUNNER_ONLY_FAILED_EVENTS"] = str(self.only_failed_event_data)
 
+        if self.containerized:
+            debug('containerization enabled')
+            for container_setting in ('process_isolation_executable', 'container_image', 'container_volume_mounts'):
```
AlanCoding: What about `container_options`? shouldn't you debug those here too?

maxamillion: +1 - container options should probably be debugged here, that and/or the full container invocation command. Maybe log `new_args` unless there's a concern of logging/leaking secret data.

AlanCoding:
I'm now thinking that the subclassing error message may be https://github.com/ansible-collections/community.general/issues/565, which was fixed on the Ansible core side, and the version in the base image is just out of date.

I also confirmed that the fix https://github.com/ansible/ansible/pull/70228 was not applied to the source that I inspected inside of the container.

ansible_runner/__main__.py
```diff
@@ -442,16 +442,16 @@ def main(sys_args=None):
...
+             "to execute **Ansible**. (2) using a sandbox (e.g. bwrap) which will by default restrict access to /tmp "
+             "(default=False)"
     )
 
     playbook_group.add_argument(
         "--process-isolation-executable",
         dest="process_isolation_executable",
-        default="bwrap",
-        help="Process isolation executable or container engine used to isolate execution. (default=bwrap)"
+        default="prodman",
```
shanemcd:
```suggestion
        default="podman",
```

jladdjr: sounds like some kind of super-hero - fixed

maxamillion (approved):

jladdjr:
> I'm now thinking that the subclassing error message may be ansible-collections/community.general#565, which was fixed on the Ansible core side, and the version in the base image is just out of date.

@AlanCoding - I just built a container that installed ansible-base from source and confirmed I didn't see the warning.

ansible_runner/runner_config.py
jladdjr: Good catch - updated

ansible_runner/__main__.py
```diff
@@ -442,57 +442,79 @@ def main(sys_args=None):
...
+        "--container-image",
+        dest="container_image",
+        default="ansible/ansible-runner",
+        help="Container image to use when running an ansible task"
+    )
+
+    playbook_group.add_argument(
+        "--container-volume-mount",
+        dest="container_volume_mounts",
+        action='append',
```
ryanpetrello, line 507: `append` I like this API, nice 👍

ansible_runner/runner_config.py
```diff
@@ -189,7 +204,9 @@ def prepare(self):
 
         # Use local callback directory
         callback_dir = self.env.get('AWX_LIB_DIRECTORY', os.getenv('AWX_LIB_DIRECTORY'))
-        if callback_dir is None:
+        if callback_dir is None and self.containerized:
+            callback_dir = '/usr/lib/python3.6/site-packages/ansible_runner/callbacks'
```
ryanpetrello, line 208: Ignoring the AWX use case, and considering people who might just install and use runner's CLI, this assumes you've installed runner in your global `site-packages`, which might break some runner use cases (what about people who install runner into a virtualenv?)

test/integration/conftest.py
```diff
@@ -18,3 +18,20 @@ def rc(request, tmpdir):
     rc.pexpect_timeout = .1
     rc.pexpect_use_poll = True
     return rc
+
+
+# TODO: determine if we want to add docker / podman
```
ryanpetrello, line 23: Yes, we should totally do this 😄

ryanpetrello (approved):
minimal feedback, 👍 this is pretty darn fancy 😄

ansible_runner/runner_config.py
AlanCoding: That gets messy because venv paths become part of the software interface. Job isolation will not be compatible with all images that have Ansible, we thought that EEs would have a fairly minimal set of requirements (ansible / runner), but I'm questioning that as we get into the details.

If you're running an image that uses venvs, then how does that get communicated from the image to `ansible-runner`? Even if the image itself had some metadata associated with it, runner would have to run docker/podman commands to find that out. See https://github.com/ansible/ansible-builder/issues/33 about venvs. 

This is a potential solution, take it with a _large_ grain of salt:

 - In the base image, copy the runner callback stuff to `/usr/share/ansible/plugins/callback`, since this is a default value of `ANSIBLE_CALLBACK_PLUGINS`
 - if the job is containerized, don't set this path at all, because there's no need to.

Considering the import hacks in the callback, I anticipate that this may take some work to get right. On the other hand... if we put the callback into a collection inside of `/usr/share/collections` this would all start to make a lot more sense.

shanemcd (approved):

jladdjr:
recheck

jladdjr:
recheck

ansible_runner/runner_config.py
```diff
@@ -189,7 +204,9 @@ def prepare(self):
 
         # Use local callback directory
         callback_dir = self.env.get('AWX_LIB_DIRECTORY', os.getenv('AWX_LIB_DIRECTORY'))
-        if callback_dir is None:
+        if callback_dir is None and self.containerized:
```
shanemcd:
```suggestion
        if self.containerized:
```

jladdjr:
recheck

PR #479, opened 2020-07-08
test_basic_events should run serially

jladdjr:
Looks like this might be another test w/ flake because of the move to parallelize test runs with xdist.


jladdjr:
recheck

kdelee (approved):

Issue #232, opened 2019-03-25
git package missing from docker Image


cloudbree:
Even though the runner image contains the ansible-galaxy script, it seems to be missing the git binary, which would might be required to install role dependencies from a git repository. Is that intended behaviour ? I think in a cloud-native scenario, that could be a very useful addition to the image. What do you think ? Is that something that we could include within this image ?

P.S.: If it would not be a problem, we can open a PR.

Thanks!

matburt:
Yep, I'd be amenable to that.

cloudbree:
Hi @matburt ,

That's great mews!

Just submitted the PR.

Thanks!

PR #481, opened 2020-07-11
Add git to Docker image (again)

nejch:
This was originally reported in https://github.com/ansible/ansible-runner/issues/232, fixed in https://github.com/ansible/ansible-runner/pull/233, then lost in refactoring (https://github.com/ansible/ansible-runner/commit/56f14a3ee15826147809d219b1090d1017776d06#diff-3254677a7917c6c01f55212f86c57fbf).

As stated in the original issue, ansible-galaxy installs from git sources fail without git.


matburt:
This will come back in with #470

PR #482, opened 2020-07-16
Implement stdout callback and its logic as a collection

AlanCoding:


ansible_runner/runner_config.py
```diff
@@ -188,18 +185,16 @@ def prepare(self):
...
-        self.env['ANSIBLE_CALLBACK_PLUGINS'] = ':'.join(filter(None,(self.env.get('ANSIBLE_CALLBACK_PLUGINS'), callback_dir)))
+        runner_root = os.path.split(os.path.abspath(__file__))[0]
+        collection_dir = os.path.join(runner_root, 'lib')
+
+        # The new method with collection
+        self.env['ANSIBLE_COLLECTIONS_PATH'] = ':'.join(filter(None,(self.env.get('ANSIBLE_COLLECTIONS_PATHS'), collection_dir)))
+
         if 'AD_HOC_COMMAND_ID' in self.env:
-            self.env['ANSIBLE_STDOUT_CALLBACK'] = 'minimal'
+            self.env['ANSIBLE_STDOUT_CALLBACK'] = 'runner.wrapper.minimal'
```
shanemcd, line 224: Why is it `runner.wrapper.minimal` here and `ansible.runner.minimal` above?

ansible_runner/runner_config.py
```diff
@@ -188,18 +185,16 @@ def prepare(self):
...
 
         # Use local callback directory
-        callback_dir = self.env.get('AWX_LIB_DIRECTORY', os.getenv('AWX_LIB_DIRECTORY'))
-        if callback_dir is None:
-            callback_dir = os.path.join(os.path.split(os.path.abspath(__file__))[0],
-                                        "callbacks")
-        python_path = self.env.get('PYTHONPATH', os.getenv('PYTHONPATH', ''))
-        if python_path and not python_path.endswith(':'):
-            python_path += ':'
-        self.env['ANSIBLE_CALLBACK_PLUGINS'] = ':'.join(filter(None,(self.env.get('ANSIBLE_CALLBACK_PLUGINS'), callback_dir)))
```
shanemcd: 🎉

ansible_runner/runner_config.py
AlanCoding: A goof.

I had meant to change everything to `runner.wrapper` because being in the `ansible` namespace carries with it a number of connotations which I do not wish to imply here.

ansible_runner/runner_config.py
```diff
@@ -210,26 +208,23 @@ def prepare(self):
...
-        if self.containerized:
-            self.env['ANSIBLE_CALLBACK_PLUGINS'] = callback_dir
-        else:
+        # Use local collection directory
+        collection_dir = self.env.get('AWX_LIB_DIRECTORY', os.getenv('AWX_LIB_DIRECTORY'))
+
+        if not self.containerized:
+            collection_dir = self.env.get('AWX_LIB_DIRECTORY', os.getenv('AWX_LIB_DIRECTORY'))
+            if collection_dir is None:
+                collection_dir = os.path.join(os.path.split(os.path.abspath(__file__))[0], "lib")
```
AlanCoding, line 216: This now shares some of the elements of #512

Dockerfile
```diff
@@ -6,17 +6,17 @@ ADD demo/project /runner/project
...
-#ADD https://releases.ansible.com/ansible-runner/ansible-runner.el8.repo /etc/yum.repos.d/ansible-runner.repo
-#RUN dnf install -y ansible-runner
+# Install Ansible
 RUN dnf install -y python3-pip rsync openssh-clients sshpass glibc-langpack-en git \
     https://github.com/krallin/tini/releases/download/v0.19.0/tini_0.19.0-amd64.rpm && \
     rm -rf /var/cache/dnf
 
 RUN dnf install -y gcc python3-devel
-RUN pip3 install bindep https://github.com/ansible/ansible/archive/devel.tar.gz \
-    https://github.com/ansible/ansible-runner/archive/devel.tar.gz
```
AlanCoding, line 19: @matburt I've updated this to work in the containerized world, and it's passing some cursory integration tests I've ran locally (requires building image, so CI isn't equipped to test that).

This line shows the outright removal of the `ansible-runner` install. This is replaced by an `ADD` layer above that adds the collection directory with the callback, plus adding `psutil` directly.

The bigger picture is that we can package and vendor the collection separately. So ultimately, this could be handled in `ansible-builder` so that instead of enumerating requirements of the base image, we can just say that the manifest of collections needs to include the runner collection in order to be usable by runner.

utils/entrypoint.sh
```diff
@@ -11,14 +11,4 @@ runner:x:`id -u`:`id -g`:,,,:/home/runner:/bin/bash
...
-if [[ -n "${LAUNCHED_BY_RUNNER}" ]]; then
-    RUNNER_CALLBACKS=$(python3 -c "import ansible_runner.callbacks; print(ansible_runner.callbacks.__file__)")
-
-    # TODO: respect user callback settings via
-    # env ANSIBLE_CALLBACK_PLUGINS or ansible.cfg
-    export ANSIBLE_CALLBACK_PLUGINS="$(dirname $RUNNER_CALLBACKS)"
-
-    export ANSIBLE_STDOUT_CALLBACK=awx_display
-fi
-
```
AlanCoding, line 21: @shanemcd this entrypoint stuff exists in order to ascertain the location of the `ansible-runner` install and put them in the env vars needed for Ansible. This proposal makes that unnecessary by putting the collection in a location expected by Ansible, which is `/usr/share/ansible/collections`, which is the same as `ansible-builder`.

It's not just `ANSIBLE_CALLBACK_PLUGINS`, it's also `PYTHONPATH` that has historically been set and used this location. However, I think that may been erroneous because the system path change was done in the callback python code.

AlanCoding:
Replaced by https://github.com/ansible/ansible-runner/pull/957

PR #485, opened 2020-07-19
Fix podman test failures

shanemcd:
Sadly, I dont have an explanation for this one yet. I am able to reproduce the error reliably, but only when running under tox:

```
Error: could not get runtime: error creating runtime static files directory /root/.local/share/containers/storage/libpod: mkdir /root/.local/share/containers: permission denied
```

This error does not happen when running the tests directly with pytest, only when running under tox.

I am able to prevent this error from happening by explicitly telling podman to use the current user's home directory. I would love to get to the bottom of this one at some point.

### DIFF
--- a/test/unit/test_runner_config.py
+++ b/test/unit/test_runner_config.py
@@ -4,6 +4,7 @@ from functools import partial
 from io import StringIO
 import os
 import re
+from pathlib import Path
 
 import six
 from pexpect import TIMEOUT, EOF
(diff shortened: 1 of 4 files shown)


test/unit/test_runner_config.py
```diff
@@ -606,19 +607,22 @@ def test_containerization_settings(mock_mkdir, container_runtime):
...
-    extra_container_args = []
+    extra_global_args = []
+    extra_run_args = []
     if container_runtime == 'podman':
-        extra_container_args = ['--quiet']
+        extra_global_args = [f'--root={str(Path.home())}/.local/share/containers/storage']
+        extra_run_args = ['--quiet']
     else:
-        extra_container_args = ['--user={os.getuid()}']
+        extra_run_args = ['--user={os.getuid()}']
```
kdelee: is this use of `--user=<some user id>` something that we should add to docs? this is to avoid the files getting owned by root if using docker, right?

shanemcd: Maybe not. We wont officially support Docker downstream.

matburt (approved):

tox.ini
```diff
@@ -10,6 +10,8 @@ isolated_build = true
 basepython = python3
 
 [testenv]
+passenv = HOME
+skip_install = true
```
jladdjr, line 14: 👍  good catch

jladdjr (approved):
image

\o/

PR #497, opened 2020-08-21
Fix permission denied error when running under docker

shanemcd:


AlanCoding (approved):

AlanCoding:
So I think https://hub.docker.com/r/ansible/ansible-runner has already been updated with this, trying to test that now.

PR #498, opened 2020-08-21
Actually fix permissions across rootless podman / docker

shanemcd:


AlanCoding (approved):
I've only tested one particular system - the Docker for Mac that I was using locally. Checking out this branch and rebuilding the image, it's working again, so 👍

Dockerfile
```diff
@@ -35,11 +35,11 @@ RUN for dir in \
       /runner/inventory \
       /runner/project \
       /runner/artifacts ; \
-    do mkdir -m 0775 -p $dir ; chmod g+rw $dir ; chown -R runner:runner $dir ; done && \
+    do mkdir -m 0775 -p $dir ; chmod g+rwx $dir ; chgrp root $dir ; done && \
```
AlanCoding, line 38: Still processing things myself, I now kind of see how this makes the layers closer to:

```
RUN chgrp -R 0 /some/directory && \
    chmod -R g=u /some/directory
```

from https://docs.openshift.com/container-platform/3.3/creating_images/guidelines.html, with a notable difference being the named `root` group, which was created a few lines above.

PR #504, opened 2020-08-26
Add build arg for Ansible install command

AlanCoding:
With this branch, I have uploaded the "2.9" tag here:

https://quay.io/repository/alancoding/ansible-runner?tab=tags

Only tested with Docker.


AlanCoding:
closing after discussion about how to do a better version of this

PR #505, opened 2020-08-26
Stream between a control and a worker instance of runner

jbradberry:
TODO:

- [x] have a worker-side runner optionally emit to a specified file/pipe the artifacts after all job events are sent
- [x] have a control-side runner optionally emit the named private data dir as streamed zip data over a given file/pipe
- [x] allow a worker-side runner to take a file/pipe for the incoming data instead of the name of a private data dir
- [x] allow a control-side runner to accept events and artifacts over a given input file/pipe and distribute them out appropriately to the event handler and other necessary logic


jbradberry:
Here's my demo workflow.

Setup in console:

    $ mkfifo /tmp/control.fifo; mkfifo /tmp/worker.fifo
    $ cp -ia ~/ansible-runner/demo/ /tmp/demo

Controller session:

    $ ipython
   
    In [1]: import ansible_runner
   
    In [2]: f_out = open('/tmp/control.fifo', 'wb')
   
    In [3]: f_in = open('/tmp/worker.fifo', 'rb')
   
    In [4]: r = ansible_runner.run(private_data_dir='/tmp/demo', playbook='test.yml'
       ...: , control_in=f_in, control_out=f_out)

Worker session:

    $ ipython
   
    In [1]: import ansible_runner

    In [2]: f_in = open('/tmp/control.fifo', 'rb')

    In [3]: f_out = open('/tmp/worker.fifo', 'wb')

    In [4]: r = ansible_runner.run(worker_in=f_in, worker_out=f_out)

Cleanup in console:

    $ rm -rf /tmp/control.fifo /tmp/worker.fifo /tmp/demo

matburt:
Commenting since this was marked as ready for review 😉 

This needs to be well documented under the Runner documentation on how to use this generally. The commits also need to be rewritten.

matburt:
This could probably also use some tests.

ansible_runner/streaming.py
```diff
@@ -0,0 +1,214 @@
...
+            else:
+                self.event_callback(data)
+
+        if self.finished_callback is not None:
+            self.finished_callback(self)
+        return self.status, self.rc
+
+    def send_job(self):
+        self.config.prepare()
+        remote_options = {key: value for key, value in self.kwargs.items() if key in remote_run_options}
```
AlanCoding: It seems like it would be easier to maintain a list of options to _exclude_ rather than a list of options to _allow_.

If I understand the problem this is trying to solve, you want to remove all options _related to streaming_. Those options should be a fairly static and definite list, like `control_in`, `control_out`, `worker_in`, etc.

ghjm:
What do you think of a different approach to the CLI:
```
ansible-runner run demo -p test.yml --via-worker 'ansible-runner worker'
```

The `--via-worker` option causes runner to function in controller mode, execute a subprocess via the shell, and set `control-in` and `control-out` to the stdout and stdin of the subprocess.

This CLI would directly allow running a job through Receptor:
```
ansible-runner run demo -p test.yml --via-worker 'receptorctl work submit playbook --node wherever --payload - --follow'
```

(The `playbook` work type on the `wherever` Receptor node would be defined to run `ansible-runner worker`.)

ghjm:
It would also be interesting, though perhaps out of scope for this PR, to consider the case where the controller does not continuously run during playbook execution.  So we would have two different processes:

1. Run ansible-runner.run() (or the CLI) in controller "phase 1" mode, produce a unit-of-work blob, and send it to Receptor for future execution.
2. Maybe much later, run ansible-runner.run() (or the CLI) in controller "phase 2" mode, where it retrieves the results from Receptor and makes callbacks and/or writes events and artifacts to the local data dir.

This would allow something like:
```
ansible-runner bundle demo -p test.yml | receptorctl work submit playbook --node wherever --payload -
```
followed by
```
receptorctl work results abCdEfG | ansible-runner interpret demo
```
(I'm sure there are better command names than "bundle" and "interpret")

ghjm:
Or for that matter, ignoring Receptor entirely, something like:
```
ansible-runner bundle demo -p test.yml | ansible-runner worker tempdir | ansible-runner interpret demo
```

ansible_runner/streaming.py
```diff
@@ -0,0 +1,214 @@
...
+            kwargs = json.dumps(remote_options, cls=UUIDEncoder)
+            archive.writestr('kwargs', kwargs)
+            archive.close()
+        buf.flush()
+
+        data = {
+            'private_data_dir': True,
+            'payload': base64.b64encode(buf.getvalue()).decode('ascii'),
+        }
+        self.control_out.write(json.dumps(data).encode('utf-8'))
```
ghjm: Do we really need to encode the data zip as JSON?  This will considerably increase the size of the data being sent over the wire.

ansible_runner/streaming.py
```diff
@@ -0,0 +1,214 @@
...
+        self.worker_out.write(json.dumps(status).encode('utf-8'))
+        self.worker_out.write(b'\n')
+        self.worker_out.flush()
+
+    def event_handler(self, event_data):
+        self.worker_out.write(json.dumps(event_data).encode('utf-8'))
+        self.worker_out.write(b'\n')
+        self.worker_out.flush()
+
+    def artifacts_handler(self, artifact_dir):
```
ghjm, line 122: If we were to store a list of artifacts during the run, we could then bundle all of them as a single zip file after the end of the JSON events, and send a single "all artifacts" zip at the end of the process - again saving bandwidth if there are any artifacts including large binary data.

ansible_runner/streaming.py
jbradberry: This has now been fixed.

ansible_runner/streaming.py
jbradberry: In practice, I am sending all of the artifact files together as a single zip.  I used this callback/handler paradigm, though, since it felt like it fit well with the other callbacks.

ghjm: Oh, I get it, you're sending a JSON stub with `artifacts: True` and then you know to read a zip file from the input.  So if there _were_ intermingled job events and artifacts, it would still work.  Nicely done!

jbradberry: Probably not any more, though, since now I'm unconditionally read()-ing after getting the `artifacts: True` blob and treating it as a single zip file.

jbradberry: Now I'm doing `artifacts: <len>`, so the potential for intermingling is back on.

ansible_runner/streaming.py
jbradberry: > If I understand the problem this is trying to solve, you want to remove all options _related to streaming_. Those options should be a fairly static and definite list, like `control_in`, `control_out`, `worker_in`, etc.

I don't believe that is actually the case.  There are a number of parameters that do not make sense to be user-provided on the worker side of this streaming scenario.  Off the top of my head, any of the callbacks/handlers.

ghjm: Yeah, if someone adds a new option, it's kinda 50/50 whether it will make sense to send to the remote side, and the safe default is not to send it.

jbradberry:
New command line demo

Setup:

    $ cp -ia ~/ansible-runner/demo/ /tmp/demo

Individual commands:

    $ ansible-runner transmit /tmp/demo -p test.yml > /tmp/output

    $ ansible-runner worker < /tmp/output > /tmp/output2

    $ ansible-runner process < /tmp/output2

The files `/tmp/output` and `/tmp/output2` look correct.  Note that creating a pipe chain of these commands currently breaks with a `BrokenPipeError`.  Also, the process command doesn't currently output anything.

ghjm:
`/tmp/output` looks like it is JSON encoded ... can this be binary zip data, to avoid the size increase?

jbradberry:
> `/tmp/output` looks like it is JSON encoded ... can this be binary zip data, to avoid the size increase?

Yeah, I was thinking of giving it the same treatment as the artifacts from the worker.

ansible_runner/__main__.py
```diff
@@ -652,6 +622,22 @@ def main(sys_args=None):
...
+    add_args_to_parser(transmit_subparser, DEFAULT_CLI_ARGS['positional_args'])
+
+    worker_subparser = subparser.add_parser(
+        'worker',
+        help="Execute work streamed from a controlling instance"
+    )
+
+    process_subparser = subparser.add_parser(
+        'process',
+        help="Receive the output of remote ansible-runner work and distribute the results"
```
AlanCoding, line 644: I kind of get transmit and worker a little bit, but I don't follow process. Why wouldn't the thing that transmits also process? I see that this takes a logfile, but would it not normally write to the artifacts directory?

ansible_runner/streaming.py
```diff
@@ -0,0 +1,220 @@
...
+
+class Processor(object):
+    def __init__(self, _input=None, status_handler=None, event_handler=None,
+                 artifacts_handler=None, cancel_callback=None, finished_callback=None, **kwargs):
+        if _input is None:
+            _input = sys.stdin.buffer
+        self._input = _input
+
+        self.kwargs = kwargs
+        self.config = ansible_runner.RunnerConfig(**kwargs)
```
AlanCoding: I don't see how kwargs would be enough to construct a RunnerConfig, at least not from the CLI options

```
$ ansible-runner process --help
usage: ansible-runner process [-h] [--version] [--debug] [--logfile LOGFILE]

optional arguments:
  -h, --help         show this help message and exit
  --version          show program's version number and exit
  --debug            enable ansible-runner debug output logging
                     (default=False)
  --logfile LOGFILE  log output messages to a file (default=None)
```

Additionally, all the parameters have already been specified in the original transmit(?) command.

ansible_runner/__main__.py
jbradberry: Originally I implemented a controller and a worker, but it was decided that having the ability to asynchronously process the results would be good.

These subcommands take the `--logfile` flag because that flag is part of the generic args group of flags.  I'm not sure it's meaningful here, as you point out, but I'm also not sure if it would be broken either.

ansible_runner/streaming.py
jbradberry: You're right, I'm not sure that this is meaningful or even not broken the way I have it here.  However, the handlers require a `RunnerConfig` object, so some kind of workaround is going to have to be found.

AlanCoding: I want to formalize my suggestions on this matter. I started out here not understanding what the requirements for the plugins are, and I have found these uses:

https://github.com/ansible/ansible-runner-http/blob/master/ansible_runner_http/events.py
https://github.com/ansible/ansible-runner-aio/blob/master/ansible_runner_aio/events.py
https://github.com/wtcross/ansible-runner-mqtt/blob/master/ansible_runner_mqtt/events.py

Those are all I can find on GitHub generally.

----
You had mentioned the idea of streaming back the details of the config object from the worker back to the controller. I think that's a bad idea. Looking further into it, I believe it wouldn't even _work_. In the http example, for example, it queries a setting like `runner_config.settings.get("runner_http_url", None)`, which uses `settings` as defined here.

https://github.com/ansible/ansible-runner/blob/b3b2be41340f1ae65899b8a504b5660cc8f1922b/ansible_runner/runner_config.py#L319

Observations from the example plugins:

a. The plugin never uses runner settings themselves, it uses plugin-specific settings
b. These settings can only be provided in the `env/settings` file

----
Wrapping this up, I don't _like_ the idea `ansible-runner process` accepting a positional argument for private_data_dir. But that's literally what the plugins require. There's no ambiguity. Maybe you could make it optional, because 99.99% of cases won't have any plugins and it's not needed in those places.

It's also established that plugins will be ran on the controller side. Combine the other precedents, that means that settings for the plugins need to be given in `env/settings` of the private_data_dir on the controller. Slap a bow on that and put it in docs.

This is all assuming that a CLI entrypoint for process is needed in the first place, and I don't feel like it should be needed for an initial merge.

shanemcd:
We should _probably_ do something like `docker save` does:

```
$ docker save awx
cowardly refusing to save to a terminal. Use the -o flag or redirect
```

ansible_runner/streaming.py
jbradberry: Also https://github.com/elmarx/ansible-runner-kafka

jbradberry: "assuming that a CLI entrypoint for process is needed in the first place": this is a red herring.  The parameters are the same regardless of whether you invoke runner using the CLI or by importing it as a library.

AlanCoding: Sorry that wasn't very clear.

Where I was going with that is, we could have `ansible-runner transmit_and_process` as opposed to `transmit` and `process`, if we don't expect to use them independently of each other for the moment.

I think there is enough information here to implement the process functionality on its own, constructing the RunnerConfig locally. Even so, a combined transmit-and-process command would eliminate _any_ question about implementation because there is no need to re-construct a RunnerConfig for a previously-started job.

ansible_runner/streaming.py
AlanCoding: There are 50 parameters at this interface.

https://github.com/ansible/ansible-runner/blob/b7c247aa224e45eb86c3ce5b79a025e8ac430716/ansible_runner/interface.py#L88-L151

See in `Worker` how it uses `r = ansible_runner.interface.run(**self.kwargs)`, so this is the method I linked to.

You're including 12. I don't think you are correct in omitting the other 38.

To throw a few out there, `fact_cache`, `cmdline`, `process_isolation`...

ansible_runner/streaming.py
```diff
@@ -0,0 +1,221 @@
...
+        self.kwargs.update(kwargs)
+
+        self.kwargs['quiet'] = True
+        self.kwargs['private_data_dir'] = self.private_data_dir
+        self.kwargs['status_handler'] = self.status_handler
+        self.kwargs['event_handler'] = self.event_handler
+        self.kwargs['artifacts_handler'] = self.artifacts_handler
+        self.kwargs['finished_callback'] = self.finished_callback
+
+        r = ansible_runner.interface.run(**self.kwargs)
```
AlanCoding, line 104: This calls `init_runner` which creates a `RunnerConfig` and calls prepare. But this is `Worker`, and the `Transmitter` already did those things by the time it got here.

I want to get some hard and fast rules about separation of responsibilities. As both the transmitter and worker are creating runner configs, there are going to be bugs, because the two are going to be diverge somehow or another.

jbradberry: Agreed, which is why I'm sitting down and figuring out where each `RunnerConfig` attribute is coming from, and whose responsibility it should be.

jbradberry: Potentially unpopular opinion: `Transmitter` should not be responsible for creating a `RunnerConfig` at all.

jbradberry: The downside is that validation that the command can be constructed in a meaningful way happens late, in `Worker`.

ghjm: You mean validation of the command line parameters?  The whole transmit/worker/process scheme should only be called by applications like Tower that are written not to issue wrong commands.

In the unusual case that a user is running this directly and types a command that doesn't make sense, but makes enough sense to get transmitted to a worker, I think it's fine if it's the worker that detects the problem.

ansible_runner/streaming.py
```diff
@@ -0,0 +1,216 @@
...
+
+        self.status = "unstarted"
+        self.rc = None
+
+    def run(self):
+        remote_options = {key: value for key, value in self.kwargs.items() if key in remote_run_options}
+
+        private_data_dir = self.kwargs.get('private_data_dir', None)
+        self._output.write(
+            utils.stream_dir(private_data_dir, kwargs=json.dumps(remote_options, cls=UUIDEncoder)))
```
AlanCoding: to be clear, this writes a file named `kwargs` into the private data dir.

That seems like something I'd like to see rolled into the ansible-runner interface contract to start with. I might be able to help with that.

ansible_runner/streaming.py
```diff
@@ -0,0 +1,216 @@
...
+    'ident',
+    'ignore_logging',
+    'inventory',
+    'limit',
+    'module',
+    'module_args',
+    'omit_event_data',
+    'only_failed_event_data',
+    'playbook',
+    'verbosity',
```
AlanCoding: If I read this right, this allow-listing process is still going to block remote containerized jobs.

jbradberry: Yeah, after looking at the attrs on `RunnerConfig` quite a bit yesterday, I'm thinking that most of them need to be passed on.

ansible_runner/streaming.py
```diff
@@ -0,0 +1,216 @@
...
+
+        return self.status, self.rc
+
+
+class Worker(object):
+    def __init__(self, _input=None, _output=None, **kwargs):
+        if _input is None:
+            _input = sys.stdin.buffer
+        if _output is None:
+            _output = sys.stdout.buffer
```
AlanCoding, line 65: The worker is also going to be writing to `sys.stdout` in other places like these:

https://github.com/ansible/ansible-runner/blob/e6020b7d33627a5d6656dca64d7ae09dd95cc75c/ansible_runner/utils.py#L288

Those will be non-json formatted lines... does `ansible-runner process` not get those? Does it get those lines and just pass over them?

ansible_runner/streaming.py
jbradberry: Yeah, that's a holdover from the old receptor integration.  I'm thinking it should just be sent in a new initial message type.

ansible_runner/streaming.py
jbradberry: We do need to protect against this.  I'm not entirely clear how I'm going to do that yet.

AlanCoding: I ran your example and I can't find any stray output in `/tmp/output2`. I don't know why.

ansible_runner/utils.py
```diff
@@ -82,6 +83,25 @@ def check_isolation_executable_installed(isolation_executable):
...
+def stream_dir(directory, **additional):
+    buf = BytesIO()
+    with zipfile.ZipFile(buf, 'w', compression=zipfile.ZIP_DEFLATED, allowZip64=True) as archive:
+        if directory:
+            for dirpath, dirs, files in os.walk(directory):
+                relpath = os.path.relpath(dirpath, directory)
+                if relpath == ".":
+                    relpath = ""
+                for fname in files:
+                    archive.write(os.path.join(dirpath, fname), arcname=os.path.join(relpath, fname))
```
AlanCoding, line 95: I think we have to be more selective here. I ran your demo commands, and found a surprisingly large zipfile was being sent. I turned out, it's because it was sending all my past artifacts, which are in the private_data_dir.

ansible_runner/streaming.py
AlanCoding: Since you removed the `worker_in` `worker_out` kind of options, I think that _all_ the kwargs should be passed.

jbradberry: Yeah, I think you are right.

AlanCoding:
It looks like the normal use is broken

```
$ ansible-runner run demo/ -p test.yml
  File "/Users/alancoding/Documents/repos/ansible-runner/ansible_runner/__main__.py", line 926, in main
    res = run(**run_options)
  File "/Users/alancoding/Documents/repos/ansible-runner/ansible_runner/interface.py", line 217, in run
    r = init_runner(**kwargs)
  File "/Users/alancoding/Documents/repos/ansible-runner/ansible_runner/interface.py", line 91, in init_runner
    rc = RunnerConfig(**kwargs)
TypeError: __init__() got an unexpected keyword argument 'streamer'
```

ansible_runner/utils.py
```diff
@@ -82,6 +83,25 @@ def check_isolation_executable_installed(isolation_executable):
         return False
 
 
+def stream_dir(directory, **additional):
```
AlanCoding: this doesn't seem to work with non-absolute paths for `directory` like `~/Documents/repos/openshift`. I'm unclear why.

ansible_runner/utils.py
```diff
@@ -82,6 +83,25 @@ def check_isolation_executable_installed(isolation_executable):
...
+                    relpath = ""
+                for fname in files:
+                    archive.write(os.path.join(dirpath, fname), arcname=os.path.join(relpath, fname))
+
+        for fname, data in additional.items():
+            archive.writestr(fname, data)
+        archive.close()
+
+    payload = buf.getvalue()
+    return b'\n'.join((json.dumps({'zipfile': len(payload)}).encode('utf-8'), payload))
```
AlanCoding, line 99: Okay, I ran this against openshift origin, which is my large project repo example, it's 1.3 Gb, because my computer is not the best.

That returned bytes with a length of 868,468,620. It took 59 seconds to produce this.

That's roughly similar time to what I expect for doing a git clone... but the size in memory is more concerning. It would be nice if we could stream this from the transmitter to the worker and have the worker go straight to writing it out without it ever being fully held in memory.

I've been looking into the receptorctl options, it will take a zip file directly in `--payload`, but this requires writing it an additional time, which I also don't want to do, but could be worth the memory penalty.

ghjm: If you're running this via the Receptor remote worker (ie, you're submitting it as a unit of work to one Receptor node and running it on another), then Receptor itself will store the whole input payload on both nodes.  So at best, we would stream the zip content out of Runner to receptorctl, which streams it to the local Receptor node, which writes it to disk; then it is sent via Receptor to the remote Receptor node, which also writes it to disk; then it is streamed into the Runner worker, which decompresses it to disk.  So the minimum cost is one compressed and one uncompressed copy on each of the two nodes.

To avoid writing it to disk a second time when supplying it to receptorctl, you could do something along the lines of `runner transmit <args> | receptorctl <args> -f -`.

jbradberry:
> It looks like the normal use is broken
> 
> ```
> $ ansible-runner run demo/ -p test.yml
> Traceback (most recent call last):
>   File "/Users/alancoding/Documents/repos/ansible-runner/ansible_runner/__main__.py", line 926, in main
>     res = run(**run_options)
>   File "/Users/alancoding/Documents/repos/ansible-runner/ansible_runner/interface.py", line 217, in run
>     r = init_runner(**kwargs)
>   File "/Users/alancoding/Documents/repos/ansible-runner/ansible_runner/interface.py", line 91, in init_runner
>     rc = RunnerConfig(**kwargs)
> TypeError: __init__() got an unexpected keyword argument 'streamer'
> ```

@AlanCoding That's odd, because I'm definitely popping `'streamer'` out of the dict.

**edit:** oops, only if it evaluates truthy

ansible_runner/utils.py
jbradberry: Presumably because `~/` is a shell-evaluated alias for `$HOME`, or something like that.

ansible_runner/utils.py
AlanCoding: When I did the AWX job directory isolation, we changed from using the project folder in-place to copying it into the job private data directory. This caused some complaints, because it meant we were copying a large folder. That ship has sailed. I accept that these changes will result in a new process that writes it out to disk 3 or 4 separate times (as opposed to 1, or 0 in ancient history).

My comment here was more directed at the memory use spike. I don't see any reason we need to have the entire project folder in memory, and doing so could cause problems. When I give the length of 868,468,620, that is the length of the object returned by this method.

jbradberry: Are we talking about the returned artifacts dump?  Because I'm about to rip that out entirely.

AlanCoding: valid question, multiple things are calling `stream_dir`. I'm thinking of the case where `ansible-runner transmit` calls it for the whole private_data_dir. Within that directory, the `project/` folder is the concern.

jbradberry:
Latest command line demo

setup:

    $ cp -ia demo/ /tmp/demo

command:

    $ ansible-runner transmit /tmp/demo -p test.yml | ssh hostname "/home/jrb/.virtualenvs/ansible-runner/bin/ansible-runner worker" | ansible-runner process /tmp/demo

teardown:

    $ rm -rf /tmp/demo

ansible_runner/streaming.py
```diff
@@ -0,0 +1,227 @@
...
+        while True:
+            line = self._input.readline()
+            data = json.loads(line)
+
+            if 'status' in data:
+                self.status_callback(data)
+            elif 'zipfile' in data:
+                self.artifacts_callback(self._input.read(data['zipfile']))
+            elif 'eof' in data:
+                break
```
AlanCoding, line 222: AWX will still want the event_callback to be invoked for the EOF event.

ansible_runner/streaming.py
```diff
@@ -0,0 +1,227 @@
...
+            else:
+                self.event_callback(data)
```
AlanCoding, line 224: I would like to spend a little time making this `if-elif-elif` super explicit instead of this duck typing thing.

I would advocate for a single key that denotes the type of runner callback. That will be important to clearly distinguish these transport-layer _callbacks_ from _job event_ recording by the plugin or python interface. In terms of the event lifecycle we will have:

 - An event from the actual Ansible stdout callback, which are "partial" events
 - The combined event, dispatched from the worker to the processor to be used in the callback
   - tagged as a unprocessed runner callback, with the callback type
 - Processed event (processed by the callback) sent into the receiving messaging system
   - sent into a dedicated queue for this type of event, contains relational tags like job / workflow / host
 - Event in database

A purist would go even further to say they should have a version associated with them. I wouldn't follow the textbook advice for event-driven architectures there, but I could see some use for returning the _runner_ version, even if only for a single callback, like the start. At some point, we're going to have problems with non-compliant nodes popping up in the execution plane.

ansible_runner/runner.py
```diff
@@ -80,7 +81,10 @@ def event_callback(self, event_data):
 
     def status_callback(self, status):
         self.status = status
-        status_data = dict(status=status, runner_ident=str(self.config.ident))
+        status_data = {
+            'status': status, 'runner_ident': str(self.config.ident),
+            'command': self.config.command, 'env': self.config.env, 'cwd': self.config.cwd
```
AlanCoding: Getting command/env/cwd are important, and this is a perfectly fine mechanism to get them. However, it doesn't seem right to send those on every status callback.

We need them on the "starting" status, I would prefer to not have them on running/successful/canceled.

ansible_runner/streaming.py
```diff
@@ -0,0 +1,227 @@
...
+        if self.status_handler is not None:
+            self.status_handler(status_data, runner_config=self.config)
+
+    def event_callback(self, event_data):
+        full_filename = os.path.join(self.artifact_dir,
+                                     'job_events',
+                                     '{}-{}.json'.format(event_data['counter'],
+                                                         event_data['uuid']))
+        if 'stdout' in event_data:
+            print(event_data['stdout'])
```
AlanCoding, line 187: Since `process` is taking in kwargs and private_data_dir, you could just assign the `quiet` config to an instance variable, and check that to determine if this should/shouldn't be printed. And the worker could always set quiet=True.

ansible_runner/runner.py
jbradberry: You are right, I forgot to make this conditional.

ansible_runner/streaming.py
jbradberry: The `worker` _does_ always set `quiet=True`.  Your point is well taken about `process`, though.

ansible_runner/streaming.py
jbradberry: I don't believe that that is the way it works.  I think the unconditional call of `finished_callback` below correctly parallels how it currently works.

ansible_runner/streaming.py
jbradberry: I'm now sending the kwargs as a separate message to be parsed by `Worker`.

ansible_runner/utils.py
jbradberry: I'm afraid that's as intended.  How is the worker going to execute a job if it doesn't have a copy of the project?

ansible_runner/streaming.py
AlanCoding: It's hard to talk precisely about this since this change isn't the full picture, but yes, it's true that someone (AWX) using this pragmatically could pass a `finished_callback` that re-implements the construction of the EOF dict at:

https://github.com/ansible/ansible-runner/blob/e6020b7d33627a5d6656dca64d7ae09dd95cc75c/ansible_runner/utils.py#L325

The point I was trying to make is that AWX makes decisions upon detection of that EOF event, concretely:

https://github.com/ansible/awx/blob/0c7bfa543b9e1b5637b69c97696e9ea145a0b98f/awx/main/dispatch/worker/callback.py#L158

To have that continue to work, it seems pretty cheap to me to add

```python
self.event_callback(data)
```

before `break`. It's not that you couldn't resolve that in the `finished_callback`... or you could do something more dramatic, like move logic currently in the callback receiver to `tasks.py` where it inputs the callback methods. But there's a lot invested in this AWX EOF logic, and it would be good to leave that untouched as much as possible.

jbradberry: That's not what I'm saying at all, though the issue is confounded by our inconsistent use of the terms 'callback' and 'handler'.

jbradberry: Ok, so, here's the deal.  We have _two_ different things that are an EOF.  We have `{'event': 'EOF'}` sent from `utils.OutputEventFilter` unconditionally to `Runner.event_callback`.  None of my work here intercepts or does anything different with this event.  Secondly, we have `{'eof': True}` as issued by `Worker.finished_callback`.  This is only intended to be a message used to break out of the loop in `Process.run`.

ansible_runner/streaming.py
jbradberry: `Worker` unconditionally gets the 'quiet' parameter set.  Hopefully everything that would leak output to stdout respects that parameter, but so far it looks like it does.

ansible_runner/utils.py
AlanCoding: The concern isn't the size of `project/` on disk, the concern is loading that into memory - risking OOM. Here, the 2 calls are:

```python
self._output.write(utils.stream_dir(private_data_dir))
self._output.write(utils.stream_dir(artifact_dir))
```

Instead of returning this value, why not pass `self._output` as an argument instead of making a new `buf = BytesIO()`.

```diff
diff --git a/ansible_runner/streaming.py b/ansible_runner/streaming.py
index 2a1ff0c..0a373a6 100644
--- a/ansible_runner/streaming.py
+++ b/ansible_runner/streaming.py
@@ -49,7 +49,7 @@ class Transmitter(object):
         self._output.flush()
 
         private_data_dir = self.kwargs.get('private_data_dir')
-        self._output.write(utils.stream_dir(private_data_dir))
+        utils.stream_dir(private_data_dir, self._output)
         self._output.write(json.dumps({'eof': True}).encode('utf-8'))
         self._output.write(b'\n')
         self._output.flush()
@@ -120,7 +120,7 @@ class Worker(object):
         self._output.flush()
 
     def artifacts_handler(self, artifact_dir):
-        self._output.write(utils.stream_dir(artifact_dir))
+        utils.stream_dir(artifact_dir, self._output)
         self._output.flush()
 
     def finished_callback(self, runner_obj):
diff --git a/ansible_runner/utils.py b/ansible_runner/utils.py
index 71cbc38..1818a33 100644
--- a/ansible_runner/utils.py
+++ b/ansible_runner/utils.py
@@ -83,8 +83,8 @@ def check_isolation_executable_installed(isolation_executable):
         return False
 
 
-def stream_dir(directory):
-    buf = BytesIO()
+def stream_dir(directory, buf):
+    buf.write(b'\n'.join((json.dumps({'zipfile': True}).encode('utf-8'), b'')))
     with zipfile.ZipFile(buf, 'w', compression=zipfile.ZIP_DEFLATED, allowZip64=True) as archive:
         if directory:
             for dirpath, dirs, files in os.walk(directory):
@@ -95,9 +95,6 @@ def stream_dir(directory):
                     archive.write(os.path.join(dirpath, fname), arcname=os.path.join(relpath, fname))
         archive.close()
 
-    payload = buf.getvalue()
-    return b'\n'.join((json.dumps({'zipfile': len(payload)}).encode('utf-8'), payload))
-
 
 def dump_artifact(obj, path, filename=None):
     '''
```

jbradberry: Oh, I see what you mean.  Good idea, if it can be made to work.

jbradberry: There's no way to do that and still write the header object with the zip file size.

AlanCoding: That's true, I don't know what the python ZipFile stuff needs exactly.

As a last ditch, it could also work to send it to a temporary file.

shanemcd (approved):
I'm good with merging this if @AlanCoding and @ghjm are. We should create issues for things we dont want to lose sight of.

ghjm (approved):
I tested this and it worked as expected:

```
ansible-runner transmit demo/ -p test.yml | receptorctl work submit runner -p - --follow --rm | ansible-runner process demo/
```

The Receptor worker was configured as:
```
 - work-command:
      worktype: runner
      command: ansible-runner
      params: worker
```

I'm OK with merging this and opening issues for any follow-on work that needs to happen.

AlanCoding:
Link https://github.com/jbradberry/ansible-runner/pull/1

that did wind up having 1 attempted bug fix, so this is not a completely trivial patch

AlanCoding (approved):

ansible_runner/utils.py
AlanCoding: I'm pretty late coming back around to this, but filed an issue for this: https://github.com/ansible/ansible-runner/issues/604

PR #507, opened 2020-08-27
Pass any envvars content into containerized runs

AlanCoding:


AlanCoding:
well

```
2020-08-27 18:03:36.714961 | centos-8 | test/integration/containerized/test_cli_containerized.py::test_provide_env_var SKIPPED [ 75%]
```

that's why it's passing.

shanemcd:
I like the idea of whitelisting any environment variables provided in `env/envvars`. One thing I want to avoid is passing through _all_ environment variables set on the host. This would be likely to cause problems in multiple ways.

AlanCoding:
recheck

AlanCoding:
Passing now with the patch, but need to see if it breaks any other tests

```
$ py.test test/integration/ -k test_provide_env_var -vv
============================================================================ test session starts =============================================================================
platform darwin -- Python 3.7.7, pytest-6.0.0, py-1.9.0, pluggy-0.13.1 -- /Users/alancoding/.virtualenvs/ansible-builder/bin/python
cachedir: .pytest_cache
rootdir: /Users/alancoding/Documents/repos/ansible-runner
collected 106 items / 105 deselected / 1 selected                                                                                                                            

test/integration/containerized/test_cli_containerized.py::test_provide_env_var PASSED                                                                                  [100%]

```

test/integration/containerized/test_cli_containerized.py
```diff
@@ -8,7 +8,6 @@
 
 @pytest.fixture
 def skip_if_no_podman(container_runtime_installed):
-    return
```
AlanCoding, line 11: whoops

AlanCoding:
@shanemcd I think it's productive to try and run these in Zuul, but it digs up lots of problems

https://github.com/ansible/ansible-runner/blob/devel/Dockerfile#L30

we give permission to `/home/runner/.ansible/tmp` there, but the test hits a permission denied error on `/runner/.ansible/tmp/ansible-local-12_wk5y1l7`. I'm not sure about the order of operations here... try to add that to the Dockerfile, build & push new image (from branch, merge separately?), then maybe we can re-run this and get it to pass.

ansible_runner/runner_config.py
```diff
@@ -145,6 +145,8 @@ def __init__(self,
         self.ssh_key_data = ssh_key
         self.execution_mode = ExecutionMode.NONE
         self.envvars = envvars
+        # for strict container isolation, keep a rolling list of what env vars should be exposed to job
+        self.env_var_whitelist = ['PROJECT_UPDATE_ID']
```
matburt: lets not use `whitelist` here. In the ansible project itself, we've made a conscious push towards `allowlist`

ansible_runner/runner_config.py
```diff
@@ -314,13 +316,17 @@ def prepare_env(self):
...
             else:
-                self.env.update({k:v for k, v in self.envvars.items()})
+                envvars_update = self.envvars
+            self.env_var_whitelist.extend(list(envvars_update.keys()))
+            self.env.update(envvars_update)
         try:
             envvars = self.loader.load_file('env/envvars', Mapping)
             if envvars:
-                self.env.update({six.text_type(k):six.text_type(v) for k, v in envvars.items()})
+                envvars_update = {six.text_type(k):six.text_type(v) for k, v in envvars.items()}
```
matburt: Given that we are moving `devel` to py3 exclusively... does it make sense to carry `six` and use these features? Or is that still necessary?

matburt: Should we try to pull `six` entirely?

ansible_runner/runner_config.py
```diff
@@ -734,11 +741,9 @@ def _parse_cli_execenv_cmd_playbook_args():
                 _ensure_path_safe_to_mount(host_path)
                 new_args.extend(["-v", "{}:{}".format(host_path, container_path)])
 
-        env_var_whitelist = ['PROJECT_UPDATE_ID']
-
         for k, v in self.env.items():
-            if k in env_var_whitelist:
-                new_args.extend(["-e", "{}={}".format(k, v)])
+            if k in self.env_var_whitelist:
+                new_args.extend(["-e", k])
```
matburt: From a cursory reading, what happens to `v` here?

AlanCoding: If you pass `-e FOO` then docker / podman will pass the value of that environment on the host machine into the container. This is so that we can pass secret values without having them show up in `job_args` in the API, or in the process list on the computer. Credit goes to Shane here, otherwise I was struggling for any way to manage secrets in env vars.

ansible_runner/runner_config.py
AlanCoding: Yeah, no reluctance to removing six, but we still need to force type here because of bugs we've hit in the past with pexpect (for say, integer types). If it's a simple swap of `six.text_type` -> `str` then that's not a problem.

ansible_runner/runner_config.py
matburt: Gotcha, okay I get it now.

Issue #508, opened 2020-08-28
ansible-runner playbook --help does not show arg for playbook


AlanCoding:
Docs give an example command:

https://github.com/ansible/ansible-runner/blob/180cda82381b99362a9c3dcb304a0076c4233d6c/docs/standalone.rst

```
$ ansible-runner playbook demo.yml -i inventory.ini
```

Here, `demo.yml` is given as a positional argument. Inspecting the help text:

```
usage: ansible-runner playbook [-h] [--private-data-dir PRIVATE_DATA_DIR]
                               [--container-runtime CONTAINER_RUNTIME]
                               [--keep-files] [--version] [--debug]
                               [--logfile LOGFILE]
                               [--container-image CONTAINER_IMAGE]
                               [--container-volume-mount CONTAINER_VOLUME_MOUNTS]
                               [--container-option CONTAINER_OPTIONS]

optional arguments:
  -h, --help            show this help message and exit
  --private-data-dir PRIVATE_DATA_DIR
                        base directory cotnaining the ansible-runner metadata
                        (project, inventory, env, etc)
  --container-runtime CONTAINER_RUNTIME
                        OCI Compliant container runtime to use. Examples:
                        podman, docker
  --keep-files          Keep temporary files persistent on disk instead of
                        cleaning them automatically. (Useful for debugging)
  --version             show program's version number and exit
  --debug               enable ansible-runner debug output logging
                        (default=False)
  --logfile LOGFILE     log output messages to a file (default=None)

Ansible Container Options:
  configuation options for executing Ansible playbooks

  --container-image CONTAINER_IMAGE
                        Container image to use when running an ansible task
  --container-volume-mount CONTAINER_VOLUME_MOUNTS
                        Bind mounts (in the form 'host_dir:/container_dir)'.
                        Can be used more than once to create multiple bind
                        mounts.
  --container-option CONTAINER_OPTIONS
                        Container options to pass to execution engine. Can be
                        used more than once to send multiple options.
```

I don't see a positional playbook option anywhere in there. This is in disagreement with the docs. An argparse-style argument should show up in "usage:" with a blurb about it, just like all of the options. Ping @maxamillion

maxamillion:
[...]

PR #509, opened 2020-08-28
Get integration tests to run content instead of returning help text

AlanCoding:
Prior behavior was that it ran just `ansible-runner` and gave help text - see comments on `subprocess.run` expected call pattern.

----
@kdelee I think this playbook comes from you. I couldn't really do it justice with my assertions.

```
PLAY [localhost] ***************************************************************

TASK [Gathering Facts] *********************************************************
[DEPRECATION WARNING]: Distribution centos 8.1.1911 on host localhost should 
use /usr/libexec/platform-python, but is using /usr/bin/python for backward 
compatibility with prior Ansible releases. A future Ansible release will 
default to using the discovered platform python for this host. See https://docs
.ansible.com/ansible/2.9/reference_appendices/interpreter_discovery.html for 
more information. This feature will be removed in version 2.12. Deprecation 
warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
ok: [localhost]

TASK [debug] *******************************************************************
ok: [localhost] => {
    "msg": "Test if we are in a container!"
}

TASK [Look for container env var defined in /proc/1/environ] *******************
fatal: [localhost]: FAILED! => {"changed": true, "cmd": "grep 'container=' /proc/1/environ", "delta": "0:00:00.006559", "end": "2020-08-28 01:48:05.721700", "msg": "non-zero return code", "rc": 2, "start": "2020-08-28 01:48:05.715141", "stderr": "grep: /proc/1/environ: Permission denied", "stderr_lines": ["grep: /proc/1/environ: Permission denied"], "stdout": "", "stdout_lines": []}
...ignoring

TASK [Look for container info defined in /proc/1/cgroup] ***********************
changed: [localhost]

TASK [Found evidence that we are in docker container] **************************
ok: [localhost]

TASK [Found evidence that we are in podman container] **************************
skipping: [localhost]

TASK [Found evidence that we are in a container] *******************************
ok: [localhost]

TASK [fail] ********************************************************************
skipping: [localhost]

localhost                  : ok=6    changed=2    unreachable=0    failed=0    skipped=2    rescued=0    ignored=1 
```

I think this gives information about whether you are running in podman or docker. In which case, you caught me, I'm running docker. However, that's not a durable assertion and no one wants to really parse Ansible output, so this is the best I've got.

### DIFF
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -49,3 +49,4 @@ def container_runtime_installed():
             return runtime
         except FileNotFoundError:
             pass
+    pytest.skip('No container runtime is available.')
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -40,13 +40,12 @@ def container_runtime_available():
 # TODO: determine if we want to add docker / podman
 # to zuul instances in order to run these tests
 @pytest.fixture(scope="session", autouse=True)
-def container_runtime_installed(container_runtime_available):
+def container_runtime_installed():
     import subprocess
 
-    if container_runtime_available:
-        for runtime in ('podman', 'docker'):
-            try:
-                subprocess.run([runtime, '-v'])
-                return runtime
-            except FileNotFoundError:
-                pass
+    for runtime in ('podman', 'docker'):
+        try:
+            subprocess.run([runtime, '-v'])
+            return runtime
+        except FileNotFoundError:
+            pass
--- a/test/integration/containerized/conftest.py
+++ b/test/integration/containerized/conftest.py
@@ -60,7 +60,7 @@ def cli(request):
         })
 
         try:
-            ret = CompletedProcessProxy(subprocess.run(args, shell=True, *a, **kw))
+            ret = CompletedProcessProxy(subprocess.run(' '.join(args), shell=True, *a, **kw))
         except subprocess.CalledProcessError as err:
             pytest.fail(
                 f"Running {err.cmd} resulted in a non-zero return code: {err.returncode} - stdout: {err.stdout}, stderr: {err.stderr}"
--- a/test/integration/containerized/test_cli_containerized.py
+++ b/test/integration/containerized/test_cli_containerized.py
@@ -6,25 +6,29 @@ import pytest
 HERE = os.path.abspath(os.path.dirname(__file__))
 
 
-@pytest.mark.serial
-def test_module_run(cli, container_runtime_available):
-    if not container_runtime_available:
-        pytest.skip('container runtime(s) not available')
-    cli(['-m', 'ping','--hosts', 'localhost', 'run', os.path.join(HERE, 'priv_data')])
+@pytest.fixture
+def skip_if_no_podman(container_runtime_installed):
+    return
+    if container_runtime_installed != 'podman':
+        pytest.skip('podman container runtime(s) not available')
 
 
 @pytest.mark.serial
-def test_playbook_run(cli, container_runtime_available):
-    if not container_runtime_available:
-        pytest.skip('container runtime(s) not available')
-    cli(['run', os.path.join(HERE,'priv_data'), '-p', 'test-container.yml'])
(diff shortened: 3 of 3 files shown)


test/integration/conftest.py
```diff
@@ -40,13 +40,12 @@ def container_runtime_available():
...
-        for runtime in ('podman', 'docker'):
-            try:
-                subprocess.run([runtime, '-v'])
-                return runtime
-            except FileNotFoundError:
-                pass
+    for runtime in ('podman', 'docker'):
+        try:
+            subprocess.run([runtime, '-v'])
+            return runtime
```
AlanCoding, line 49: The fixture `container_runtime_available` returns True if, and only if, _both_ docker and podman are installed. Well if that's the case, you don't need to run this loop because I can tell you the answer. It's podman.

Assuming that this had a point other than returning either "podman" or None, I made it so that it can return "docker" if that is there but podman is not.

test/integration/containerized/conftest.py
```diff
@@ -60,7 +60,7 @@ def run(args, *a, **kw):
         })
 
         try:
-            ret = CompletedProcessProxy(subprocess.run(args, shell=True, *a, **kw))
+            ret = CompletedProcessProxy(subprocess.run(' '.join(args), shell=True, *a, **kw))
```
AlanCoding, line 63: Some people don't like to see `shell=True`. I do not expression a position on that. But if you are going to use shell...

https://docs.python.org/3/library/subprocess.html

```python
>>> subprocess.run(["ls", "-l"])  # doesn't capture output
CompletedProcess(args=['ls', '-l'], returncode=0)

>>> subprocess.run("exit 1", shell=True, check=True)
  ...
subprocess.CalledProcessError: Command 'exit 1' returned non-zero exit status 1
```

It takes a list if you don't pass `shell=True`. If you do use shell mode, docs say to pass a string, but if you pass a list anyway, then it runs only the first element of the list. In our case `ansible-runner` by itself exits with code 0, so the fact that this was happening appears to not be known by anyone.

test/integration/containerized/test_cli_containerized.py
```diff
@@ -6,43 +6,49 @@
...
 
 @pytest.mark.serial
-def test_playbook_run(cli, container_runtime_available):
-    if not container_runtime_available:
-        pytest.skip('container runtime(s) not available')
-    cli(['run', os.path.join(HERE,'priv_data'), '-p', 'test-container.yml'])
+def test_playbook_run(cli, skip_if_no_podman):
+    r = cli(['run', os.path.join(HERE,'priv_data'), '-p', 'test-container.yml'])
+    assert 'PLAY RECAP *******' in r.stdout
+    assert 'failed=0' in r.stdout
```
AlanCoding, line 26: Again, this is the best assertion I've got. As the prior state demonstrates, we need to make _some_ assertion or else the test can pass as a false-pass.

test/integration/containerized/test_cli_containerized.py
```diff
@@ -6,43 +6,49 @@
...
-        pytest.skip('container runtime(s) not available')
-    cli(
+def test_playbook_with_private_data_dir(cli, skip_if_no_podman, container_runtime_installed):
+    r = cli(
         [
             'playbook',
             '--private-data-dir', os.path.join(HERE,'priv_data'),
             '--container-runtime', container_runtime_installed,
-            'test-container.yml'
+            'test/integration/containerized/priv_data/project/test-container.yml'  # FIXME: this may be a bug
```
AlanCoding, line 50: @maxamillion related to https://github.com/ansible/ansible-runner/issues/508, it would _most logically_ follow that the playbook path here is relative to the project folder from the `--private-data-dir`, but in the absence of a doc string for the argument and no prior experience with this, I cannot completely ascertain intent.

matburt (approved):

test/integration/containerized/conftest.py
kdelee: woops

PR #510, opened 2020-09-02
Make roles work in AWX

AlanCoding:
This addresses a specific issue we saw testing with AWX. User puts collection requirements in `collections/requirements.yml`, AWX downloads them to these folders, and more-or-less, the location just isn't mounted. This mounts it.

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -734,14 +734,21 @@ class RunnerConfig(object):
 
         # These directories need to exist before they are mounted in the container,
         # or they will be owned by root.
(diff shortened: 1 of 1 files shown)


AlanCoding:
Now I'm thinking that a better solution might be to not have a loop like `['project', 'inventory', 'env']`, and instead, loop over all subdirectories in the private data dir and then mount them if they exist, as opposed to creating them if they don't exist. I don't know why we would ever want to create them.

AlanCoding:
that's going to take some testing, but I do think it's the right thing to do

AlanCoding:
On 3rd thought, I still really do not like this

PR #511, opened 2020-09-02
Use relative paths for containerized private_data_dir settings

AlanCoding:
Containers are ugly because you have to juggle two sets of paths because the host context has a different filesystem than what the container does. The private_data_dir is mounted to the static location of `/runner`.

This changes the path specification for things inside of the private data directory to a relative specification which relies on the fact that `--workdir` is set to the static location of `/runner/project`

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -328,10 +328,13 @@ class RunnerConfig(object):
             self.env = {}
             # Special flags to convey info to entrypoint or process in container
             self.env['LAUNCHED_BY_RUNNER'] = '1'
-            artifact_dir = os.path.join("/runner/artifacts", "{}".format(self.ident))
-            self.env['AWX_ISOLATED_DATA_DIR'] = artifact_dir
+            self.env['AWX_ISOLATED_DATA_DIR'] = os.path.join(
+                os.pardir, 'artifacts', str(self.ident)
+            )
             if self.fact_cache_type == 'jsonfile':
-                self.env['ANSIBLE_CACHE_PLUGIN_CONNECTION'] = os.path.join(artifact_dir, 'fact_cache')
+                self.env['ANSIBLE_CACHE_PLUGIN_CONNECTION'] = os.path.join(
+                    os.pardir, 'artifacts', str(self.ident), 'fact_cache'
+                )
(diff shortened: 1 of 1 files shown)


ansible_runner/runner_config.py
```diff
@@ -332,10 +332,13 @@ def prepare_env(self):
...
             self.env['LAUNCHED_BY_RUNNER'] = '1'
-            artifact_dir = os.path.join("/runner/artifacts", "{}".format(self.ident))
-            self.env['AWX_ISOLATED_DATA_DIR'] = artifact_dir
+            self.env['AWX_ISOLATED_DATA_DIR'] = os.path.join(
+                os.pardir, 'artifacts', str(self.ident)
+            )
             if self.fact_cache_type == 'jsonfile':
-                self.env['ANSIBLE_CACHE_PLUGIN_CONNECTION'] = os.path.join(artifact_dir, 'fact_cache')
+                self.env['ANSIBLE_CACHE_PLUGIN_CONNECTION'] = os.path.join(
+                    os.pardir, 'artifacts', str(self.ident), 'fact_cache'
```
matburt, line 336: There's been a feature that exists in Runner for a while that allows you specify a different location for the artifact directory. It seems like this undoes that? iirc... that's what `artifact_dir` became as a property in this context.

AlanCoding: If you see the line being deleted `artifact_dir = os.path.join("/runner/artifacts", "{}".format(self.ident))`, that is not a customizable path, and it is a path inside the container filesystem.

We could look into supporting a custom artifact path for containerized runs, that would probably involve designating a new location in the container filesystem, making Dockerfile updates, and using that as a new volume mount. I think there would need to be some conditionals specific to that scenario, in order to not break anything else. Before looking at any of these specifics, I would like to start out with an integration test.

matburt: I definitely just want to make sure that it's *possible* to specify an artifact location outside of the private data dir.

AlanCoding: That is definitely not possible for jobs in containers, and has never been.

I want to keep focused on the standard input structure relevant to AWX until _that_ works.

PR #513, opened 2020-09-03
Mount private data directory to share novel folders like requirements_roles

AlanCoding:
AWX makes use of a number of folders inside of the private directory, and enumerating them is going to miss some.

Not only does it make use of folders, but it lays down a `cp` file that is used for something related to ssh-agent. Unless we refactor that, this is the only option that works. See a prior attempt at #510.

Testing initially confirms that this gets roles to work in AWX again.

This also pulls in #512, and won't work without it due to folder permission errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,7 @@ clean:
 	rm -rf rpm-build
 	rm -rf deb-build
 	find . -type f -regex ".*\py[co]$$" -delete
+	rm -rf $(shell find test/ -type d -name "artifacts")
 
 dist:
 	poetry build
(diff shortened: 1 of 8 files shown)


AlanCoding:
Turns out this is giving some unanticipated permissions error, which I have no explanation for.

AlanCoding:
I have now identified that this permission error was pre-existing, from some other source. This is my preferred fix for the purpose of using folders like `roles_requirements`, which works today.

shanemcd (approved):

PR #514, opened 2020-09-04
Do not set callback in entrypoint, so ad hoc commands work

AlanCoding:
This is causing a behavior that causes ad hoc commands to look like playbook runs

Screen Shot 2020-09-04 at 8 45 42 AM

but it should look like

```
localhost | CHANGED | rc=0 >>
hello world
```


AlanCoding:
recheck

shanemcd (approved):

PR #516, opened 2020-09-08
Add test for containerized privledge escalation

AlanCoding:


AlanCoding:
Documenting what I know here:

This fails on this raised exception:

https://github.com/ansible/ansible/blob/3e9943bc5e7a9cd393757aa8100d7fed80bd316e/lib/ansible/plugins/action/__init__.py#L616

Why is it raising the exception? Because the current user is `root`, and that user is known to be an admin user.

The method was supposed to have fixed up permissions (in a hand-wavey sense, coming into the middle of a large code base) and exited before that line, in that method. According to the comment

> it means we failed to chown as root

I had to force Ansible to print out what _actually_ happened in the commands it ran in that method, and I got this:

```
TASK [Gathering Facts] **************************************************************************************************************************************************************
final try command
{'rc': 127, 'stdout': '', 'stdout_lines': [], 'stderr': '/bin/sh: setfacl: command not found\n', 'stderr_lines': ['/bin/sh: setfacl: command not found']}
final final try command
{'rc': 0, 'stdout': '', 'stdout_lines': [], 'stderr': '', 'stderr_lines': []}
finalx3 try command
{'rc': 1, 'stdout': '', 'stdout_lines': [], 'stderr': 'chown: invalid user: ‘runner’\n', 'stderr_lines': ['chown: invalid user: ‘runner’']}
```

those correspond to:

```python
        # Step 3a: Are we able to use setfacl to add user ACLs to the file?
        res = self._remote_set_user_facl(
            remote_paths,
            become_user,
            setfacl_mode)
# next one
            res = self._remote_chmod(remote_paths, 'u+x')
# third one
        # Step 3c: File system ACLs failed above; try falling back to chown.
        res = self._remote_chown(remote_paths, become_user)
```

Given that the comment is referencing a `chown` command, we can take the relevant error to be the:

> chown: invalid user: ‘runner’

Okay Ansible, why were you trying to give the runner user permission? I guess this is "my" fault in writing the test because I used `--become-user runner`. I am fully aware that "runner" is not an actual user unless the entrypoint hacks it in.

AlanCoding:
This PR was helpful for tests, but now I'm sufficiently satisfied with what we have on the AWX side

PR #518, opened 2020-09-10
Use a file in artifacts to enumerate podman env var allow list

AlanCoding:


AlanCoding:
See documentation for `--env-file` here:

https://docs.docker.com/engine/reference/commandline/run/

We are not putting the _values_ into the file. Because that would mean we are writing secrets to disk, and we should not do that. Only the keys are written, and these signal to podman/docker that those environment variables from the parent process should be applied to the job running inside the container.

AlanCoding:
looks like tests are good

shanemcd (approved):

AlanCoding:
I tested this along with AWX, and I'm really happy with how the results look so I'm going to go ahead with merge.

PR #519, opened 2020-09-11
Add test to catch collection install permission denied error

AlanCoding:
We are trying to hit a specific permission denied error when writing `MANIFEST.json`. I can't get it with docker or podman in a local container, but hoping that maybe Zuul might hit it.

If not, I might try to parameterize with a user id to see if maybe we can get something more similar to headless podman.


AlanCoding:
I think this is sufficiently covered in other places, I'm fine to kick it out of the queue

PR #520, opened 2020-09-15
Allow customizing Ansible version in Dockerfile, use source for runner install

AlanCoding:
For a demo of `make image_matrix` see:

https://quay.io/repository/alancoding/ansible-runner?tab=tags

This shows tags produced for series of Ansible versions, 2.9, 2.10, and devel. These are `pip` installed from the Ansible branches. We were already installing from `devel`, this just adds some stuff so we can do the others.

----

Ideally, we would build the image at the start of Zuul runs. This doesn't fix that problem, but it gets us closer by building `ansible-runner` from the source (where the Dockerfile exists).


matburt (approved):

matburt:
👍

maxamillion (approved):

AlanCoding:
with this latest CI round, it will be testing in conjunction with the image from Shane's latest changes

maxamillion (approved):

Dockerfile
```diff
@@ -2,24 +2,24 @@ ARG BASE_IMAGE=docker.io/fedora:32
...
+# Install system packages for use in all images
+RUN dnf install -y \
+    python3-pip \
+    python3-devel \
+    gcc \
+    rsync \
+    openssh-clients \
+    sshpass \
+    glibc-langpack-en \
+    git \
```
shanemcd, line 14: We might want to remove git from here now since it should live in the builder-side metadata.

shanemcd (approved):

Dockerfile
AlanCoding: you're right, but in this case, the git module is part of ansible-base.

Philosophically I would prefer to remove it, but it would push it to an `append` step type customization in the definition. Also, I suspect it's needed by the `pip install git+...` lines too.

So I want to, but these are the reasons we haven't.

shanemcd: Fair enough. Marging.

PR #521, opened 2020-09-15
handle automounts and ENV vars in a more generic way, add cloud credential automounts

maxamillion:

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -705,22 +705,88 @@ class RunnerConfig(object):
                                 )
                             ])
 
-            # volume mount ~/.ssh/ into the exec env container
-            new_args.extend(["-v", "{}/.ssh/:/home/runner/.ssh/".format(os.environ['HOME'])])
 
-            # volume mount system-wide ssh_known_hosts the exec env container
-            if os.path.exists('/etc/ssh/ssh_known_hosts'):
-                new_args.extend(["-v", "/etc/ssh/ssh_known_hosts:/etc/ssh/ssh_known_hosts"])
+            # Handle automounts
+            cli_automounts = [
+                {
+                    'ENVS': ['SSH_AUTH_SOCK'],
+                    'PATHS': [
+                        {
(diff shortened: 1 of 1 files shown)


ansible_runner/runner_config.py
```diff
@@ -705,22 +705,88 @@ def _parse_cli_execenv_cmd_playbook_args():
...
+                            'src': '{}/.kube/'.format(os.environ['HOME']),
+                            'dest': '/home/runner/.kube/'
+                        },
+                    ]
+                },
+                {
+                    "ENVS": [
+                        'AWS_URL', 'EC2_URL', 'AWS_ACCESS_KEY_ID', 'AWS_ACCESS_KEY',
+                        'EC2_ACCESS_KEY', 'AWS_SECRET_ACCESS_KEY', 'AWS_SECRET_KEY', 'EC2_SECRET_KEY',
+                        'AWS_SECURITY_TOKEN', 'EC2_SECURITY_TOKEN', 'AWS_REGION', 'EC2_REGION'
```
AlanCoding, line 737: Background question here, for the `ansible-runner playbook` command, does it copy all the env vars from the host into the container?

For environment keys that point to a path on the filesystem, couldn't you just loop over all the host environment variables, and if it points to a valid path, mount that. Unless it's something that shouldn't be allowed? I just think it would be easier to build out an exclude list, rather than building out an allow list.

maxamillion: @AlanCoding I hadn't considered that. The main reason I was trying the include-list approach was to make sure we're not mounting too many things into the container and/or accidentally adding a ENV var or mount point that has unexpected consequences.

shanemcd:
This makes sense to me but the `wrap_args_for_containerization` function is now 208 lines long. We really need to do some refactoring, either as part of this PR or as a follow-up task.

maxamillion:
@shanemcd I'll do the refactor in a separate PR just to keep the changes separate, but for sake of discussion should I make a function that's specific to the `adhoc`/`playbook` subcommands the gets called conditionally?

shanemcd (approved):

PR #522, opened 2020-09-16
Fix compat commands

shanemcd:


shanemcd:
recheck

maxamillion (approved):

ansible_runner/runner_config.py
```diff
@@ -724,12 +724,30 @@ def _parse_cli_execenv_cmd_playbook_args():
...
+            )
+        ]
+
+        if 'artifacts' not in private_subdirs:
+            private_subdirs += ['artifacts']
+
+        for d in private_subdirs:
+            if not os.path.exists(os.path.join(self.private_data_dir, d)):
+                if d == 'artifacts':
+                    os.mkdir(os.path.join(self.private_data_dir, d), 0o700)
```
AlanCoding: This should _only_ apply to artifacts, because `private_subdirs` only includes things which `os.path.isdir`

so you could write

```python
          if not os.path.exists(os.path.join(self.private_data_dir, 'artifacts')):
                os.mkdir(os.path.join(self.private_data_dir, 'artifacts'), 0o700)
```

This is what some of my closed PRs looked like

ansible_runner/runner_config.py
```diff
@@ -724,12 +724,30 @@ def _parse_cli_execenv_cmd_playbook_args():
...
+                else:
+                    continue
```
AlanCoding: Oh my gosh, I had no idea what was happening here.

The reason you `continue` when you hit a directory that already exists is because the "project" directory may _already be mounted_. By using the continue, you avoid the `new_args.extend`, that's how you avoid mounting it twice.

I had no idea about this feature of the compatibility commands - it is built to use relative paths, not just accept relative paths, but not even convert them into absolute paths when passing them into `ansible-playbook`.

Here:

https://github.com/ansible/ansible-runner/blob/c2a960193484ce57bfb462749b671088c2a3e97c/ansible_runner/runner_config.py#L702-L704

and here:

https://github.com/ansible/ansible-runner/blob/c2a960193484ce57bfb462749b671088c2a3e97c/ansible_runner/runner_config.py#L676-L678

It is counting on the fact that `--workdir` is exactly `/runner/project`

I kept banging my head against the wall asking "how the heck is it creating the project directory"? Because it wasn't created on the host machine, it was happening in the container somehow. It was created by another volume mount. The point of this is to not mount the same thing twice.

But given that information, this isn't fully what we would want. See the snippets above, inventory is stashed in the project directory, because `/runner/project` is the catch-all base for all relative paths. On the other hand, if the inventory directory exists, you're going to get in the situation of _not_ mounting that folder when you should. `os.path.exists(os.path.join(self.private_data_dir, d))` is not the heuristic we would want for this.

Assuming that you want to keep this (current working directory) == /runner/project global concept, the right way to do this is to track what directories have been mapped (really, only those working in the project folder umbrella), and add logic to either not map paths which are children of other paths, or duplicates of other paths. That is possible, but is certainly non-trivial.

utils/entrypoint.sh
```diff
@@ -4,11 +4,24 @@
...
+
+    # Only needed for RHEL 8. Try deleting this conditional (not the code)
+    # sometime in the future. Seems to be fixed on Fedora 32
+    # If we are running in rootless podman, this file cannot be overwritten
+    ROOTLESS_MODE=$(cat /proc/self/uid_map | head -n1 | awk '{ print $2; }')
+    if [[ "$ROOTLESS_MODE" -eq "0" ]]; then
 cat << EOF > /etc/passwd
 root:x:0:0:root:/root:/bin/bash
 runner:x:`id -u`:`id -g`:,,,:/home/runner:/bin/bash
 EOF
```
AlanCoding, line 17: the indentation of this is off.

ansible_runner/runner_config.py
```diff
@@ -724,12 +724,30 @@ def _parse_cli_execenv_cmd_playbook_args():
 
             if 'podman' in self.process_isolation_executable:
                 # container namespace stuff
+                new_args.extend(["--group-add=root"])
                 new_args.extend(["--userns=keep-id"])
                 new_args.extend(["--ipc=host"])
```
AlanCoding, line 729: This code is only reached if in the block `if self.cli_execenv_cmd`. Is that intended?

I can agree (by experimentation) that if `--userns=keep-id` is used, then `--group-add=root` is needed (in addition to the Dockerfile changes).

But why is `--group-add=root --userns=keep-id --ipc=host` added for `ansible-runner playbook/adhoc` but not added for `ansible-runner run`?

AlanCoding (approved):

utils/entrypoint.sh
AlanCoding: We talked about an alternative option of

```bash
cat 'root:x:0:0:root:/root:/bin/bash' >> /etc/passwd
```

But I didn't understand the `cat` command use here, and there are pitfalls, so we need to test image building in separate branch.

shanemcd:
recheck

PR #523, opened 2020-09-16
Parameterize integration tests with user and container runtime

AlanCoding:


AlanCoding:
After talking more, this is not the direction that people want after all

PR #524, opened 2020-09-16
[do not merge] running tests with change in private_data_dir volume mapping

AlanCoding:


AlanCoding:
so image change didn't have an effect, and the user id really does need to match the host, for reasons I did not understand before.

PR #529, opened 2020-09-22
Remove more serial markers

AlanCoding:
I believe this was fixed in a recent PR where the UUID4 method was not ran for each job, so those jobs ran by the python interface all shared the same identifier.

### DIFF
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -1,10 +1,13 @@
+import os
+import shutil
+
 import pytest
 import pexpect
 from ansible_runner.runner_config import RunnerConfig
 
 
 @pytest.fixture(scope='function')
-def rc(request, tmpdir):
+def rc(tmpdir):
     rc = RunnerConfig(str(tmpdir))
     rc.suppress_ansible_output = True
     rc.expect_passwords = {
@@ -13,9 +16,9 @@ def rc(request, tmpdir):
     }
     rc.cwd = str(tmpdir)
     rc.env = {}
-    rc.job_timeout = 2
+    rc.job_timeout = 10
     rc.idle_timeout = 0
-    rc.pexpect_timeout = .1
+    rc.pexpect_timeout = 2.
     rc.pexpect_use_poll = True
     return rc
 
@@ -39,7 +42,7 @@ def container_runtime_available():
 
 # TODO: determine if we want to add docker / podman
 # to zuul instances in order to run these tests
-@pytest.fixture(scope="session", autouse=True)
+@pytest.fixture(scope="session")
 def container_runtime_installed():
     import subprocess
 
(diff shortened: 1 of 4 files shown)


AlanCoding:
Ah, this is my problem https://github.com/pytest-dev/pytest-xdist/issues/271

test/integration/test_main.py
```diff
@@ -175,15 +170,13 @@ def test_role_bad_project_dir():
...
-@pytest.mark.serial
-def test_role_run_clean(skipif_pre_ansible28):
+def test_role_run_clean(skipif_pre_ansible28, clear_integration_artifacts):
 
     rc = main(['run', '-r', 'benthomasson.hello_role',
                '--hosts', 'localhost',
                '--roles-path', 'test/integration/roles',
                "test/integration"])
     assert rc == 0
-    ensure_removed("test/integration/artifacts")
```
AlanCoding, line 186: Some of these implemented it in try-except, and others didn't. You're never going to get consistency like this.

test/integration/conftest.py
```diff
@@ -50,3 +53,21 @@ def container_runtime_installed():
...
+    '''Fixture is session scoped to allow parallel runs without error
+    '''
+    if 'PYTEST_XDIST_WORKER' in os.environ:
+        # we never want to clean artifacts if running parallel tests
+        # because we cannot know when all processes are finished and it is
+        # safe to clean up
+        return
+
+    def rm_integration_artifacts():
+        path = "test/integration/artifacts"
```
AlanCoding, line 69: This isn't totally coherent, because this isn't the only private_data_dir in the tests. We don't bother cleaning up the artifacts from the other directories. Obviously I have preferences - I have been adding new private_data_dirs to `test/data` folder, because I don't want them polluting the test files. I get that someone else looked at this and also said "hey, I don't want these artifact files polluting the test files". So I get it, but this was a partial answer which is arguably worse than no answer. It complicated unrelated efforts to parallelize the tests, and the problem quickly re-emerged in `test/integration/containerized/priv_data`, if you could have called it solved for `test/integration/artifacts` in the first place.

My goal here is not to refactor, and thus I try to make this only slightly less bad. My goal here is to un-break the multiprocessing.

jbradberry (approved):

AlanCoding:
It looks like maybe the pid file is an actual problem in the flake here

```
>           assert os.path.exists(os.path.join(temp_dir, 'pid'))
E           AssertionError: assert False
E            +  where False = <function exists at 0x7fe35f4f45e0>('/tmp/tmp4cg14jf2/pid')
E            +    where <function exists at 0x7fe35f4f45e0> = <module 'posixpath' from '/home/zuul/src/github.com/ansible/ansible-runner/.tox/ansible28/lib64/python3.8/posixpath.py'>.exists
E            +      where <module 'posixpath' from '/home/zuul/src/github.com/ansible/ansible-runner/.tox/ansible28/lib64/python3.8/posixpath.py'> = os.path
E            +    and   '/tmp/tmp4cg14jf2/pid' = <function join at 0x7fe35f4f4e50>('/tmp/tmp4cg14jf2', 'pid')
E            +      where <function join at 0x7fe35f4f4e50> = <module 'posixpath' from '/home/zuul/src/github.com/ansible/ansible-runner/.tox/ansible28/lib64/python3.8/posixpath.py'>.join
E            +        where <module 'posixpath' from '/home/zuul/src/github.com/ansible/ansible-runner/.tox/ansible28/lib64/python3.8/posixpath.py'> = os.path
```

Issue #506, opened 2020-08-26
Canceling is not hooked up for jobs in containers


AlanCoding:
I am basing my example here off the `ansible-builder` pytz example:

https://github.com/ansible/ansible-builder/tree/devel/test/data/pytz

modify the `pytz.yml` playbook to sleep for 1 minute:

```yaml
---
- hosts: localhost
  gather_facts: false
  connection: local
  tasks:
  - name: Debug using the rrule plugin
    debug:
      msg: >
        {{ query(
          'awx.awx.tower_schedule_rrule',
          'minute',
          start_date='2020-4-16 03:45:07',
          end_on='2'
        ) }}

  - name: sleep
    command: sleep 60

```

Build your image an all that jazz, and run this for a reference:

```
ansible-runner run test/data/pytz --playbook=pytz.yml
```

Watch it finish.
Check `docker ps -a` and you find that the container is now gone (expected).

Now, run this bash script:

```
$ cat alan.sh 
ansible-runner start test/data/pytz --playbook=pytz.yml
echo $?
ansible-runner stop test/data/pytz
echo $?
```

Expected result: no running container after this finishes.

Actual result: running container persists after this finishes for roughly 1 minute until the process has finished.

In practice, I have experienced several jobs where they are completely blocked on something and should be canceled. Because canceling doesn't do anything but reports no error code, the containers are never cleaned up.

AlanCoding:
should we use `docker kill`?

https://docs.docker.com/engine/reference/commandline/kill/

I also have a really big question of reference-ability. How do we identify the container that needs to be killed?

AlanCoding:
The fix for this has been merged in #530

PR #530, opened 2020-09-23
Kill-on-cancel for containers launched from programatic interface

AlanCoding:
Connect https://github.com/ansible/ansible-runner/issues/506

This does _not_ solve the problem for using the CLI, which is specifically the steps I gave in the issue:

```
ansible-runner start test/data/pytz --playbook=pytz.yml
ansible-runner stop test/data/pytz
```

However, I realize that involves a redesign of the `ansible-runner` contract with the user.

Reason is that the pidfile is saved in `<private_data_dir>/pid`... as opposed to:

 - `<private_data_dir>/artifacts/pid`
 - `<private_data_dir>/artifacts/<ident_value>/pid`

This is because the `ansible-runner stop` kind of commands give the private data dir. Well, multiple jobs could be running using the same private data dir (although not entirely correctly). Putting that aside, how would we go from the pid value to the container it is running in? I don't think we would, we would probably write a new file to save the container name.

I am absolutely not going to introduce a new temporary file to the top-level of the private data dir without specific agreement on the broader architectural direction. If anything, I'd expect that we would move away from addressing with the private data dir and use user-provided `ident` on top of that.

----

This solution should probably solve it for the AWX use case... for now. We can come back to the CLI fix if in the receptor integration it loses the parent ansible-runner process (which I don't think it will).

Also, we do SIGKILL for ordinary Ansible processes, so I will not weep for the undignified demise of these containers.

### DIFF
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -218,7 +218,7 @@ class Runner(object):
             if password is not None:
                 child.sendline(password)
                 self.last_stdout_update = time.time()
-            if self.cancel_callback:
+            if (not self.canceled) and self.cancel_callback:
                 try:
                     self.canceled = self.cancel_callback()
                 except Exception as e:
@@ -231,8 +231,10 @@ class Runner(object):
                 # if isinstance(extra_update_fields, dict):
                 #     extra_update_fields['job_explanation'] = "Job terminated due to timeout"
             if self.canceled or self.timed_out or self.errored:
+                self.kill_container()
                 Runner.handle_termination(child.pid, is_cancel=self.canceled)
             if self.config.idle_timeout and (time.time() - self.last_stdout_update) > self.config.idle_timeout:
+                self.kill_container()
                 Runner.handle_termination(child.pid, is_cancel=False)
                 self.timed_out = True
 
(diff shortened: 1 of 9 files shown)


ansible_runner/utils.py
```diff
@@ -390,3 +390,12 @@ def ensure_str(s, encoding='utf-8', errors='strict'):
...
     return s
+
+
+def sanitize_container_name(original_name):
+    """
+    Docker and podman will only accept certain characters in container names
+    This takes a given name from user-specified values and replaces the
+    invalid characters so it can be used in docker/podman CLI commands
+    """
+    return re.sub('[^a-zA-Z0-9_-]', '_', original_name)
```
AlanCoding: https://github.com/moby/moby/pull/3197/files

AlanCoding:
Getting the test to work in Zuul was a little brutal, looks like the reason gets into gnarly interactions between podman & the pytest multiprocessing. While I don't like making it serial, I can't run it with the rest of the parallel tests without fairly extraordinary effort.

If one were to ask me how to later make this work in parallel tests, I would suggest a dead-band at the start. Don't look for the running container until, say, 0.5 seconds after the first time the cancel_callback was called. Otherwise, you'll catch it after the `podman run` command, but before it has time to get going. I didn't want to deal with variable scope issues and whatnot, in addition to making it slower.

ansible_runner/runner.py
```diff
@@ -218,7 +218,7 @@ def _decode(x):
             if password is not None:
                 child.sendline(password)
                 self.last_stdout_update = time.time()
-            if self.cancel_callback:
+            if (not self.canceled) and self.cancel_callback:
```
matburt: I think my concern here is... what if you need repeated calls to cancel?

AlanCoding: This doesn't affect how many times `Runner.handle_termination` is called. Do you mean to ask if a plugin needs multiple calls to `cancel_callback`?

shanemcd (approved):
Of course you also added some comprehensive tests.

ansible_runner/runner.py
AlanCoding: I pushed a commit to address this which has the test refactor I mentioned as a possibility. I'm worried that we're going too far being overly-conservative, to the point of protecting bad code practices.

PR #537, opened 2020-09-29
Try testing Ansible 2.10 and 2.11

AlanCoding:


AlanCoding:
Looks like this may need changes in Zuul config too...

PR #541, opened 2020-10-07
RFC: allow for environment driven env settings

mattp-:
expose ANSIBLE_RUNNER_{PASSWORDS,SETTINGS,ENVVARS,SSH_KEY} as an alternative means of driving ansible runner inputs vs serializing to file. I'm submitting this for feedback on the idea, if this feature is something that would be accepted I will polish/document/add tests. This could be useful in scenarios where you don't want to be serializing at least passwords to disk, where environ is ephemeral to the lifetime of the process.

### DIFF
--- a/ansible_runner/runner_config.py
+++ b/ansible_runner/runner_config.py
@@ -306,7 +306,10 @@ class RunnerConfig(object):
         with existing values so the :py:class:`ansible_runner.runner.Runner` object can read and use them easily
         """
         try:
-            passwords = self.loader.load_file('env/passwords', Mapping)
+            if 'ANSIBLE_RUNNER_PASSWORDS' in os.environ:
+                passwords = self.loader.load_env('ANSIBLE_RUNNER_PASSWORDS', Mapping)
+            else:
+                passwords = self.loader.load_file('env/passwords', Mapping)
             self.expect_passwords = {
                 re.compile(pattern, re.M): password
                 for pattern, password in iteritems(passwords)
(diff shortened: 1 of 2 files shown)


ansible_runner/runner_config.py
```diff
@@ -352,7 +361,10 @@ def prepare_env(self):
 
         try:
             if self.ssh_key_data is None:
-                self.ssh_key_data = self.loader.load_file('env/ssh_key', string_types)
+                if 'ANSIBLE_RUNNER_SSH_KEY' in os.environ:
+                    self.ssh_key_data = self.loader.load_env('ANSIBLE_RUNNER_SSH_KEY', Mapping)
```
AlanCoding, line 365: Wouldn't it be preferable to cut right to the Ansible settings?

https://github.com/ansible/ansible/blob/devel/lib/ansible/config/base.yml

You could pass `ANSIBLE_PRIVATE_KEY_FILE`, and ansible-runner should pass that to the subprocess, and then Ansible should respect it for the same function as this.

mattp-: hmm I didn't know that existed actually, yes for ssh key it would indeed make sense. we could either duplicate it for parity with other ANSIBLE_RUNNER_* or just remove it in lieu of using ANSIBLE_PRIVATE_KEY_FILE. my main motivation is really avoiding the passwords file actually :)

mattp-: actually looking a bit closer, ANSIBLE_PRIVATE_KEY_FILE is a path to a key, not the privkey contents itself, so a bit different.

mattp-: @AlanCoding any further thoughts on this functionality?

shanemcd: Hey @mattp-. Alan is out for a couple more weeks after his wife gave birth. Hang tight and he'll get to this eventually.

mattp-: ah had no idea. congratulations :) and no rush, thanks.

PR #543, opened 2020-10-13
Do not mangle roles_path when it is a string

aetherith:
Previously, when providing a single string as a parameter to `roles_path` the value would be mangled to insert colons between every character in the string. Now the assignment checks if the provided value is a single string or a list of strings and processes it into the environment variable accordingly.


shanemcd (approved):
Thank you 😄

shanemcd:
Would you mind rebasing and seeing if the failures go away?

aetherith:
Ok, so I *think* I did that rebase correctly. Let me know if you see anything shocking and I'll try to patch/correct it.

PR #545, opened 2020-10-18
Fix editable installations using `pip install -e .`

duckinator:
When doing `pip install -e .`, the build process needs setuptools before
reading setup.py, which means it needs to be specified in pyproject.toml's
build-system table.

This addresses the problem mentioned in pypa/pip#8986

cc @AlanCoding


shanemcd (approved):
Thanks again!

PR #551, opened 2020-10-30
Switch the streaming protocol to use lxma-compressed tar

jbradberry:
The zipfile library doesn't really do file permissions or symlinks.

### DIFF
--- a/ansible_runner/streaming.py
+++ b/ansible_runner/streaming.py
@@ -6,7 +6,7 @@ import stat
 import sys
 import tempfile
 import uuid
-import zipfile
+
 try:
     from collections.abc import Mapping
 except ImportError:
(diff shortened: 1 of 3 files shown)


shanemcd:
Closing and re-opening in hopes of catching Zuul's attention.

ghjm:
The reason we switched from tar to zip in the first place was that Python's tarfile class allows path traversal attacks.  The zipfile class has protections against this.  See https://bugs.python.org/issue21109#msg215656.  This was first pointed out by @chrismeyersfsu in my original PR for this: https://github.com/ansible/ansible-runner/pull/432

shanemcd:
I've opened https://github.com/ansible/ansible-runner/pull/552 as an alternative.

PR #552, opened 2020-10-30
Preserve permissions when unzipping worker payload

shanemcd:
Pulled from: https://www.burgundywall.com/post/preserving-file-perms-with-python-zipfile-module


ghjm:
Do we also need to preserve permissions in the artifact unzipping?

ghjm (approved):

PR #558, opened 2020-11-10
AnsibleRunner.run use dump_artifact for status/rc

mattp-:
a few lines less of redundant code. I've also moved the status updates
to the status_handler, so it gets updated as a run progresses, as
opposed to only when the run is complete.


pabelanger:
recheck

matburt:
I think this is good, probably needs to be re-based and re-submitted.

PR #564, opened 2020-11-12
Create tox -evenv entry point

pabelanger:
This allows a user to quickly install the tox requirement files, to then
source ~/.tox/venv/bin/avtivate, to run make dist.

Helpful to quickly create a tarball.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -56,3 +56,6 @@ commands=
     poetry env info
     poetry run py.test -v -n 4 -m "not serial" test
     poetry run py.test -v test -m serial
+
+[testenv:venv]
+commands = {posargs}


shanemcd:
Could you just use poetry here? You just need to run `poetry build` and it spits out a wheel and sdist to `./dist`.

pabelanger:
Yes, that is what we do with this,
```
tox -evenv --notest
source .tox/venv/bin/activate; make dist
```
The other option I thought of, was moving the direct requirements out of tox into (test-)requirement.txt files.  I'd like to avoid the need to manually managed the dependencies.

pabelanger:
Actually, the more I think about this, I can just call python setup.py directly.

PR #571, opened 2020-11-16
Don't try to promote ansible-runner image

pabelanger:
Right now, we do not have a promote pipeline job running for container
uploads. This means, we only want to tag 'devel' releases on upload.

Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/723


pabelanger:
recheck

PR #575, opened 2020-11-19
Attempt to deal gracefully with errors in streamed message parsing

jbradberry:


jbradberry:
recheck

ansible_runner/streaming.py
```diff
@@ -87,8 +87,12 @@ def update_paths(self, kwargs):
 
     def run(self):
         while True:
-            line = self._input.readline()
-            data = json.loads(line)
+            try:
+                line = self._input.readline()
+                data = json.loads(line)
+            except (json.decoder.JSONDecodeError, IOError):
+                self.status_handler({'status': 'error'}, None)
```
AlanCoding, line 94: can we give some more text somewhere that this was a JSON parser issue in the transmission/processing phase? Could we pass the job_explanation key, would that be ok?

ansible_runner/streaming.py
```diff
@@ -87,8 +87,12 @@ def update_paths(self, kwargs):
 
     def run(self):
         while True:
-            line = self._input.readline()
-            data = json.loads(line)
+            try:
+                line = self._input.readline()
```
AlanCoding, line 91: Wasn't the problem seen that `line` is empty string? If that _should_ be an error case, I'd like to have a special error message just for that case.

jbradberry: I think so.  This was coming about if I remember correctly because the sender was crashing, so we were receiving less data than we were expecting.

ansible_runner/streaming.py
AlanCoding: For this to work `finished_callback` needs to be called, because that sends the eof event. Even if Process gets the error status, it will hang waiting for the eof.

jbradberry: You are probably right.

jbradberry:
The commits for this one have been picked up by #616.  Closing.

PR #581, opened 2020-12-01
Switch to multi-stage builds using ansible/python-builder

pabelanger:
This is the first step towards supporting multi-stage builds in
ansible-runner. In this commit, we are simplifying our Dockerfile and
using a builder image to build python wheels. We then use the compiled
wheels to install into the base image, allowing us to remove compile
time dependencies.

Depends-On: https://github.com/ansible/python-builder-image/pull/12


pabelanger:
Pull-request updated, HEAD is now a76eaf3f8d16d5a99342b002c42c267a9b9876a3

pabelanger:
Pull-request updated, HEAD is now a97a566628299c8912ca2db4341abafddcf09db7

Dockerfile
```diff
@@ -1,23 +1,18 @@
-ARG BASE_IMAGE=docker.io/fedora:32
+FROM quay.io/ansible/python-builder:latest as builder
```
AlanCoding, line 1: The quay.io pages have no descriptions. Sorry for being so out of the loop here, but can I get the Containerfile/Dockerfile from which this is being built? This will be very important to us to maintain on our own team as a part of the pipelines for producing execution environments, and I know I have already seen a lot of fixes that impact functionality in the collection install process.

Dockerfile
```diff
@@ -1,23 +1,18 @@
...
-RUN dnf install -y \
-    python3-pip \
-    gcc \
-    rsync \
-    openssh-clients \
-    sshpass \
-    glibc-langpack-en \
-    git \
-    https://github.com/krallin/tini/releases/download/v0.19.0/tini_0.19.0-amd64.rpm && \
-    rm -rf /var/cache/dnf
```
AlanCoding, line 15: I'm totally behind this, but we're going to need to get it tested fully, and we don't yet have automation for it.

I'm a little skeptical of the resultant image not having pip.

I'm also skeptical off the bat seeing tini being removed. I tried removing it earlier and it was very wrong.

Dockerfile
pabelanger: Yah, I just started on publishing it, I need to better understand how the readme for that works.  https://github.com/ansible/python-builder-image is the repo for now, I do agree adding the right people is the way to go.

Right now it is setup to just deal with bindep / requirements files. The next step for me, would be to grow ansible-collections support. But I also think talking more about it is good.

Dockerfile
pabelanger: I only drop tini, since I couldn't find an official rpm package or python package. So, I have no idea the difference between dumb-init and it.  I don't have a preference except, I don't feel using random package from web is a good idea

piip will be installed, that will come from python-base, which is not published yet. I hope to get that on monday

pabelanger: Oh, as for testing. I've created https://github.com/ansible/ansible-builder/pull/130 in ansible-runner. My goal is to get integration tests running in zuul, then we can use depends-on on a ansible-builder PR, to reference this PR, and zuul will do the work to use the speculative container for quay.io/ansible/ansible-runner:devel.

Once we confirm ansible-builder, and any another project is great, we can merge with confidence

pabelanger:
Pull-request updated, HEAD is now 4805d26eadaff6a88562929afefa5ddb066a8fa3

pabelanger:
Pull-request updated, HEAD is now c36f10e102805a0f814bc40607ed05e21b8ceb15

pabelanger:
Pull-request updated, HEAD is now 0cce3e2466ba9b24f8dabbac73bfc971c0c2d126

pabelanger:
Pull-request updated, HEAD is now 1df5cbcda9ee9afa047bd528e7619932f81d4ced

pabelanger:
Pull-request updated, HEAD is now 6809678070c1bc542a9d99a0fbd265ae361d8661

pabelanger:
Pull-request updated, HEAD is now e3f0c3cc63a273895bd027961094af140eb0f7ce

.zuul.d/jobs.yaml
```diff
@@ -3,14 +3,20 @@
...
     parent: ansible-build-container-image
     description: Build ansible-runner container image
-    pre-run: playbooks/ansible-runner-build-container-image/pre.yaml
+    required-projects:
+      - github.com/ansible/ansible
     timeout: 2700
     provides: ansible-runner-container-image
+    requires:
+      - python-builder-container-image
+      - python-base-container-image
```
AlanCoding, line 13: what's the need for this image, since it seems like it's a subset of `python-builder-container-image` after a quick look at the Containerfiles

requirements.txt
```diff
@@ -0,0 +1 @@
+ansible-core
```
AlanCoding: I was asking about this a lot in Slack, but the PyPI version of `ansible-core` doesn't seem to be good. But given the purpose of this file, I'm not sure if it would be a problem.

.zuul.d/jobs.yaml
pabelanger: python-builder is for compile time, and python-base is the runtime stage.  The OS for each is the same, just different dependencies, and build scripts installed.

requirements.txt
pabelanger: I can demo more, but the way our build system works now in zuul with multi-stage builds, we use requirement files to pull in the needed bits.  I'll link build logs shortly.

pabelanger:
Pull-request updated, HEAD is now b02033d18cae4cbc67c6f0b4393a17e0f65a2fbd

pabelanger:
recheck

pabelanger:
recheck

pabelanger:
Pull-request updated, HEAD is now 5ab41bfc490825e21d3a0012964e9bfc8bfe0b58

pabelanger:
recheck

shanemcd:
This is looking great. Thanks for updating the entrypoint stuff we talked about. What else is left here before you'd be ready to talk about merging?

requirements.txt
shanemcd: Should we just change this to a git url until this actually exists?

pabelanger: we might, or we update makefile to git clone the repo. I can test tomorrow to see what happen if we use git+https://github here.

Dockerfile
```diff
@@ -1,23 +1,16 @@
-ARG BASE_IMAGE=docker.io/fedora:32
+FROM quay.io/ansible/python-builder:latest as builder
+# =============================================================================
 
-FROM ${BASE_IMAGE}
```
AlanCoding, line 3: @shanemcd do we need to support customizing the base image? There was some very important reason we did this...

pabelanger:
Pull-request updated, HEAD is now 2f88f67cc40706ed7372739ec97181fe1776d35b

Dockerfile
shanemcd: I _think_ I just added this back when I was unsure what it would be, and got tired of updating the file. We can iterate more on this in the future if we need to, but I think for now this is fine.

requirements.txt
AlanCoding: Previously the thing we were doing was passing an Ansible branch into `make image` and it would install from source, from the `stable-2.9` or `devel` branch, or whatever.

We expect to do some version of that again, which may argue for the "git clone the repo" option, but I don't want to be opinionated beyond acknowledging the requirement.

pabelanger: Once landed, we can start rolling 2.9 / 2.10 images again, adding the jobs for that is no issue.  Can also add the logic into Makefile to deal with it too.

pabelanger:
recheck

pabelanger:
recheck

pabelanger:
recheck

pabelanger:
recheck

AlanCoding:
I want to link things where appropriate, as I understand, the status here is that we're trying to work through https://github.com/ansible/ansible-builder/pull/148, because it's deeply tied to this PR

pabelanger:
Pull-request updated, HEAD is now 814e00f4f96b68dc4d6a32c2836f9814d387986c

pabelanger:
Pull-request updated, HEAD is now 73d6692f0a8e8c54d4a9c5d3e75e2c6a6b62323b

shanemcd (approved):
Amazing work. It's been fun collaborating on this.

PR #582, opened 2020-12-01
Switch to centos:8 for BASE container image

pabelanger:
This is the first step towards multi-stage builds. We'll be basing our
images off centos-8, which gets us one step closer to using UBI8
container images.  Given that fedora is fast moving, every 6 month we'd
have to update this image or hit EOL.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,23 @@
-ARG BASE_IMAGE=docker.io/fedora:32
+ARG BASE_IMAGE=quay.io/centos/centos:8
 
 FROM ${BASE_IMAGE}
 
 # Install system packages for use in all images
 RUN dnf install -y \
+    epel-release \
+    dnf-plugins-core && \
+    dnf config-manager --set-disabled epel && \
+    dnf install -y --enablerepo epel \
+    sshpass && \
+    dnf install -y \
     python3-pip \
     gcc \
     rsync \
     openssh-clients \
-    sshpass \
     glibc-langpack-en \
     git \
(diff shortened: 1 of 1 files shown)


pabelanger:
Pull-request updated, HEAD is now 2e2882b85d9e4c0f229635da266bf00be2a3c0c8

pabelanger:
recheck

pabelanger:
Pull-request updated, HEAD is now 6a2cc0883211ba2441d5a5b71fccd827fc927e66

AlanCoding:
This is my fault, but I worry that what I did in https://github.com/ansible-collections/community.kubernetes/pull/211 will come back to bite us

> kubernetes-client [platform:fedora]

pabelanger:
> This is my fault, but I worry that what I did in [ansible-collections/community.kubernetes#211](https://github.com/ansible-collections/community.kubernetes/pull/211) will come back to bite us
> 
> > kubernetes-client [platform:fedora]

I feel this is always going to be an issue, the only really way to solve it, is provided multiple ansible-runner images, for all the distro.

That doesn't mean, the community.k8s collection cannot package their own kubernetes-client rpm.  Which I think is something we could discuss.

At the end of the day, this also because an issue for downstream UBI images, if kubernetes-client is missing, this collection would now work.

shanemcd:
We talked about this earlier in chat, but the features that require `oc` / `kubectl` in Ansible have been deprecated. We don't need them for anything on AWX anymore either, so I think either leaving the metadata as-is, or removing the bindep.txt altogether would be fine.

pabelanger:
I'm going to close this one, given the progress we are making in https://github.com/ansible/ansible-runner/pull/581

PR #586, opened 2020-12-08
[WIP] Change ansible-base to ansible-core

AlanCoding:
Ansible changed its package name from ansible-base to ansible-core.

However, I think this may leave us in a bit of a catch 22.

https://pypi.org/project/ansible-base/   last release 1 hour ago

https://pypi.org/project/ansible-core/ last release Oct 27

So this may still not be right. I know that if you pip install from source, you get ansible-core.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -56,5 +56,5 @@ pathlib2 = "*"
 scandir = "*"
 
 [build-system]
-requires = ["setuptools", "poetry>=1.0.5", "ansible-base"]
+requires = ["setuptools", "poetry>=1.0.5", "ansible-core"]
 build-backend = "poetry.masonry.api"
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -15,7 +15,7 @@ def is_pre_ansible28():
         if LooseVersion(pkg_resources.get_distribution('ansible').version) < LooseVersion('2.8'):
             return True
     except pkg_resources.DistributionNotFound:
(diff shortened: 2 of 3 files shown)


pabelanger:
I think ansible-core is right way to do this, we solve this in zuul, by building the tarball for ansible-core, then pip installing it, not from pypi.  

naming is kind of a mess now

AlanCoding:
So the definition goes to:

https://github.com/ansible/project-config/blob/eed9880044e7b78a6e08741b6331fa1f840053ef/zuul.d/projects.yaml#L714

which goes to

https://github.com/ansible/ansible-zuul-jobs/blob/61711792e44c640deff02a51a34372a13f2089c6/zuul.d/jobs.yaml#L2417

So if I'm renaming that here, it has to be renamed there. Or maybe I could keep the name, there could be other issues beyond that.

pabelanger:
Yah, we'll need a 2 stage change for this.  update ansible-zuul-jobs, then project-config. Merge this, then remove ansible-base jobs.

pabelanger:
I think we can close this now, we handle it in multi-stage builds

AlanCoding:
Maybe a solution would be something along the lines of https://github.com/ansible/ansible-builder/pull/158

Those entries for `ansible-base` are still in there, and if we could just get rid of those (instead of changing them) that would be the best. Anywhere we have "ansible-base" is either going to do nothing or break things.

PR #594, opened 2020-12-17
python3.8 container image

pabelanger:
This rebuilds ansible-runner container images using python3.8 as the
base version.

Depends-On: https://github.com/ansible/python-builder-image/pull/17

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,6 @@ Ansible Runner
 [![Ansible Mailing lists](https://img.shields.io/badge/Mailing%20lists-Ansible-orange.svg)](https://docs.ansible.com/ansible/latest/community/communication.html#mailing-list-information)
 
 
-
 Ansible Runner is a tool and python library that helps when interfacing with Ansible directly or as part of another system whether that be through a container image interface, as a standalone tool, or as a Python module that can be imported. The goal is to provide a stable and consistent interface abstraction to Ansible.
 
 For the latest documentation see: [https://ansible-runner.readthedocs.io](https://ansible-runner.readthedocs.io/en/latest/)


AlanCoding:
Does this not depend on https://github.com/ansible/python-base-image/pull/7? I guess I don't have the full picture of the graph Zuul uses. The python-base image is used directly by both python-builder-image _and_ ansible-runner. It's non-obvious to me that it would be getting the python-base image from the Zuul dependencies, as opposed to the version on quay.

pabelanger:
recheck

pabelanger:
recheck

pabelanger:
> Does this not depend on [ansible/python-base-image#7](https://github.com/ansible/python-base-image/pull/7)? I guess I don't have the full picture of the graph Zuul uses. The python-base image is used directly by both python-builder-image _and_ ansible-runner. It's non-obvious to me that it would be getting the python-base image from the Zuul dependencies, as opposed to the version on quay.

yes, that is right. How we express that, is in the zuul job it self: https://github.com/ansible/ansible-runner/blob/devel/.zuul.d/jobs.yaml#L11

This means we always pull from both upstream quay.io and our intermediate zuul-registry, if the sha are the same, that is good. We are using the released version. If using depends-on, we pull the layer that only exists in zuul-registry, but tagged as quay.io/ansible/anbsible-runner:devel.

Confusing, but basically if there is a depends-on in the commit message, we look for a pre-build container image from that PR.

pabelanger:
recheck

pabelanger:
@AlanCoding @shanemcd If we do this, we are going to have issues with selinux / dnf, given they are still compiled against python36 on centos IIRC.

How do we plan to solve that?  cc @maxamillion

shanemcd:
Ansible itself will need to address these issues. I have talked briefly to @nitzmahone about it and it sounded like they had a plan in place. Perhaps we cant do this in Runner until a version of core exists that solves these problems.

AlanCoding (approved):
this PR doesn't change any code, but I would like to update the image

shanemcd (changes requested):
I would like to hold off on this until we have answers for: https://github.com/ansible/ansible-runner/pull/594#issuecomment-747647955

shanemcd (approved):
After discussion yesterday we have decided to #shipit.

pabelanger:
Pull-request updated, HEAD is now f0c620465db77bf1c90bcfe8e2b1eb9a84ee9195

pabelanger:
Pull-request updated, HEAD is now fce786417e5dc557bf040c6dce6d1ddb592c14f7

PR #602, opened 2021-01-05
Fix passwords regex in docs

616b2f:
For me regex from docs for SSH and BECOME in `env/passwords` file didn't work. I used docker container `ansible/ansible-runner:1.4.6`.

BECOME looked like this:
```
BECOME password[defaults to SSH password]:
```

It was necessary to add a space after the semicolon for both.
Took some time to figure that out because I am new to ansible and if you use wrong regex you don't see any useful output, it's just stops printing.

I hope it helps others.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -102,8 +102,8 @@ In order for **Runner** to respond with the correct password, it needs to be abl
 by providing a yaml or json formatted file with a regular expression and a value to emit, for example::
 
   ---
-  "^SSH [pP]assword:$": "some_password"
-  "^BECOME [pP]assword:$": "become_password"
+  "^SSH [pP]assword: $": "some_password"
+  "^BECOME [pP]assword.*: $": "become_password"
 
 ``env/cmdline``
 ---------------
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -102,8 +102,8 @@ In order for **Runner** to respond with the correct password, it needs to be abl
 by providing a yaml or json formatted file with a regular expression and a value to emit, for example::
 
   ---
-  "^SSH [pP]assword: $": "some_password"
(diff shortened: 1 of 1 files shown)


AlanCoding:
AWX has to make use of this, and has regular expressions in code here:

https://github.com/ansible/awx/blob/96a7fe0035427774f4eac93983cbb887ab275a42/awx/main/tasks.py#L1843

For these cases:

```python
        d[r'BECOME password.*:\s*?$'] = 'become_password'
        d[r'SSH password:\s*?$'] = 'ssh_password'
```

It's preferable to keep regex as non-greedy as possible with how pexpect is used. It looks like this text happens in a fairly specific case where both `--ask-pass` and `--ask-become-pass` are used.

https://github.com/ansible/ansible/blob/f569d80fde32889b8952fa13f66e615cbd5b4419/lib/ansible/cli/__init__.py#L246

Can you confirm that's the case for you?

If you want to sync the AWX regex with these in runner docs, I'm okay with that. I would prefer for them to match exactly.

616b2f:
Thank you for the suggestion and detailed infos.

Yes that's the case I use both `--ask-pass` and `--ask-become-pass`.

I agree on the non-greedy part, it's a better expression in awx. The only thing I would add for the regex is the `^` at the beginning. It would not be exactly like in awx but less greedy and as far as i know it's a better practice to add it at the beginning when we expect to match the whole string. What do you think? 

The file would look like this:
```yaml
---
"^SSH password:\\s*?$": "some_password"
"^BECOME password.*:\\s*?$": "become_password"
```

Note the extra backslashes because we use YAML file here. I already tested it and it worked.

AlanCoding:
Yeah, that looks good to me

616b2f:
@AlanCoding It's updated now.

AlanCoding (approved):

PR #605, opened 2021-01-06
POC for making stream_dir work without loading into memory

AlanCoding:
This is still incomplete because it only address `stream_dir` and not `unstream_dir`, but I want to open it up to feedback about the general method, and confirm tests are passing.

I will also prefer to break this up into multiple functions and test them separate from the full integration test.

### DIFF
--- a/ansible_runner/utils.py
+++ b/ansible_runner/utils.py
@@ -15,6 +15,7 @@ import pipes
 import uuid
 import codecs
 import zipfile
+from pathlib import Path
 
 try:
     from collections.abc import Iterable, Mapping
@@ -84,8 +85,9 @@ def check_isolation_executable_installed(isolation_executable):
 
 
 def stream_dir(directory, buf):
-    local_buf = BytesIO()
-    with zipfile.ZipFile(local_buf, 'w', compression=zipfile.ZIP_DEFLATED, allowZip64=True) as archive:
+    fd, zip_file = tempfile.mkstemp(suffix='.zip')
+
+    with zipfile.ZipFile(zip_file, 'w', compression=zipfile.ZIP_DEFLATED, allowZip64=True) as archive:
         if directory:
             for dirpath, dirs, files in os.walk(directory):
                 relpath = os.path.relpath(dirpath, directory)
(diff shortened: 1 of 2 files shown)


ansible_runner/utils.py
```diff
@@ -95,8 +97,21 @@ def stream_dir(directory):
                     archive.write(os.path.join(dirpath, fname), arcname=os.path.join(relpath, fname))
         archive.close()
 
-    payload = buf.getvalue()
-    return b'\n'.join((json.dumps({'zipfile': len(payload)}).encode('utf-8'), payload))
+    zip_size = Path(zip_file).stat().st_size
```
AlanCoding, line 100: I definitely had to test that this gives the same answer as `len(f.read())`, because that was not at all obvious to me.

AlanCoding:
What I'm interested in doing from here:

 - [ ] Add a new option to `transmit` to save the zip file to, as an alternative to the temporary file
 - [ ] Make sure the temporary file gets deleted
 - [ ] Profile memory use for compressing a large repo like openshift origin
 - [ ] Develop a similar solution for the process of extracting a zip file

ansible_runner/utils.py
```diff
@@ -83,9 +84,10 @@ def check_isolation_executable_installed(isolation_executable):
         return False
 
 
-def stream_dir(directory):
-    buf = BytesIO()
-    with zipfile.ZipFile(buf, 'w', compression=zipfile.ZIP_DEFLATED, allowZip64=True) as archive:
+def stream_dir(directory, buf):
+    fd, zip_file = tempfile.mkstemp(suffix='.zip')
```
jbradberry, line 88: I'd suggest `tempfile.TemporaryFile` instead so that the zipfile is always deleted, even if something goes wrong before the end of this function.

jbradberry: (or maybe `NamedTemporaryFile`)

shanemcd:
I'm super curious what the implications are now that we know we need to base64 encode this data.

shanemcd:
Rebased this and resolved a conflict.

AlanCoding:
I won't be able to finish this, unless it waits until April. I just wanted to have all my thoughts out there, which they are now.

> I'm super curious what the implications are now that we know we need to base64 encode this data.

I was about to say that'd be easy because this PR is doing custom buffering and you'd just encode each chunk before you write it, and you'd reliably read the same chunks. Then I realized the encoding will change the length, and that be non-trivial, could cause boundary errors...

shanemcd:
Closing in favor of https://github.com/ansible/ansible-runner/pull/670

PR #607, opened 2021-01-07
Add test for adhoc via transmit command

AlanCoding:
<this does not work right now>

Still hunting for either a way to make this work, or a bug definition.

### DIFF
--- a/test/integration/test_transmit_worker_process.py
+++ b/test/integration/test_transmit_worker_process.py
@@ -2,12 +2,14 @@ import io
 import os
 
 import pytest
+import json
 
 from ansible_runner import run
 from ansible_runner.streaming import Transmitter, Worker, Processor
 
 
-def test_remote_job_interface(tmpdir, test_data_dir):
+@pytest.mark.parametrize("job_type", ['run', 'adhoc'])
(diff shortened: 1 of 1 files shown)


AlanCoding:
This is working now, so I submit it as simply additional test coverage.

Right now, you _cannot_ use `transmit` with the CLI. The normal playbook use looks like:

```
ansible-runner transmit demo -p test.yml > /tmp/output
ansible-runner worker --private-data-dir alan < /tmp/output > /tmp/output2
ansible-runner process demo < /tmp/output2
```

jbradberry:
As far as I'm aware, anything with adhoc should be doable using `ansible-runner run -m ...`, which can directly be converted to a `ansible-runner transmit ...` command.

jbradberry (approved):

AlanCoding:
I don't think a demo for that was ever written down in a public place, I couldn't figure it out, so this is @jbradberry's suggestions reformatted.

```
ansible-runner transmit demo -m command -a 'ls -l' --hosts localhost > /tmp/output
ansible-runner worker --private-data-dir alan < /tmp/output > /tmp/output2
ansible-runner process demo < /tmp/output2
```

PR #613, opened 2021-01-15
Add option for only streaming kwargs

shanemcd:
This is helpful when the private data dir exists on the same system. Without this, we unnecessarily zip up data, only to unpack it again on the same host.

### DIFF
--- a/ansible_runner/streaming.py
+++ b/ansible_runner/streaming.py
@@ -35,6 +35,7 @@ class Transmitter(object):
             _output = sys.stdout.buffer
         self._output = _output
         self.private_data_dir = os.path.abspath(kwargs.pop('private_data_dir'))
+        self.only_transmit_kwargs = kwargs.pop('only_transmit_kwargs', False)
         self.kwargs = kwargs
 
         self.status = "unstarted"
@@ -47,7 +48,9 @@ class Transmitter(object):
         self._output.write(b'\n')
         self._output.flush()
 
-        self._output.write(utils.stream_dir(self.private_data_dir))
+        if not self.only_transmit_kwargs:
+            self._output.write(utils.stream_dir(self.private_data_dir))
+
         self._output.write(json.dumps({'eof': True}).encode('utf-8'))
         self._output.write(b'\n')
         self._output.flush()


AlanCoding:
I do agree with this change.

However, I still want to know what the story will be for AWX setting the private data dir for the worker command.

With https://github.com/ansible/ansible-runner/pull/609 I wanted to start enshrining expectations for transmit / worker / process management of the private data dir.

For the solution to the AWX problem, this is what I had in mind:

```diff
diff --git a/ansible_runner/streaming.py b/ansible_runner/streaming.py
index 0ca39f4..b3146b6 100644
--- a/ansible_runner/streaming.py
+++ b/ansible_runner/streaming.py
@@ -34,7 +34,7 @@ class Transmitter(object):
         if _output is None:
             _output = sys.stdout.buffer
         self._output = _output
-        self.private_data_dir = os.path.abspath(kwargs.pop('private_data_dir'))
+        self.private_data_dir = os.path.abspath(kwargs.get('private_data_dir'))
         self.kwargs = kwargs
 
         self.status = "unstarted"
@@ -67,10 +67,9 @@ class Worker(object):
         self.kwargs = kwargs
         self.job_kwargs = None
 
-        private_data_dir = kwargs.get('private_data_dir')
-        if private_data_dir is None:
-            private_data_dir = tempfile.TemporaryDirectory().name
-        self.private_data_dir = private_data_dir
+        self.private_data_dir = None
+        if kwargs.get('private_data_dir'):
+            self.private_data_dir = kwargs.get('private_data_dir')
 
         self.status = "unstarted"
         self.rc = None
@@ -91,6 +90,11 @@ class Worker(object):
             data = json.loads(line)
 
             if 'kwargs' in data:
+                if data['kwargs'].get('private_data_dir') and not self.private_data_dir:
+                    self.private_data_dir = data['kwargs'].get('private_data_dir')
+                else:
+                    self.private_data_dir = tempfile.TemporaryDirectory().name
+
                 self.job_kwargs = self.update_paths(data['kwargs'])
             elif 'zipfile' in data:
                 zip_data = self._input.read(data['zipfile'])
```

That would involve changing the docs wording in that PR I linked - that `worker` will inherit the private data dir from `transmit` by default. This sounds reasonable to me.

This approach is an alternative to your solution here https://github.com/ansible/awx/pull/8598/commits/578f352b6f8c7683eb7ebc58be81f27b6b354b25#diff-6fb7bd98930ec6a05c1ae9296b41b63921e1813dcee53895f16769f910917c38R3106

I don't like tying these two things together with the receptor config. I have a feeling that won't age well as we support more types of deployments. Its value changes with every job, and I'd prefer for receptor to not need that info.

AlanCoding (approved):

PR #614, opened 2021-01-17
Ensure ssh key fifo is always cleaned up

shanemcd:
In cases where ssh-add failed, we were never cleaning up the fifo. This little trap trick fixes that.

### DIFF
--- a/test/unit/test_runner_config.py
+++ b/test/unit/test_runner_config.py
@@ -407,19 +407,31 @@ def test_prepare_with_ssh_key(open_fifo_write_mock):
 def test_wrap_args_with_ssh_agent_defaults():
     rc = RunnerConfig('/')
     res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], '/tmp/sshkey')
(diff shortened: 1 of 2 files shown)


AlanCoding (approved):

test/unit/test_runner_config.py
```diff
@@ -407,19 +407,31 @@ def test_prepare_with_ssh_key(open_fifo_write_mock):
 def test_wrap_args_with_ssh_agent_defaults():
     rc = RunnerConfig('/')
     res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], '/tmp/sshkey')
-    assert res == ['ssh-agent', 'sh', '-c', 'ssh-add /tmp/sshkey && rm -f /tmp/sshkey && ansible-playbook main.yaml']
+    assert res == [
+        'ssh-agent',
+        'sh', '-c',
+        "trap 'rm -f /tmp/sshkey' EXIT && ssh-add /tmp/sshkey && rm -f /tmp/sshkey && ansible-playbook main.yaml"
```
pabelanger, line 413: unrelated,

shred is a great util to zero out a file, then delete.

shanemcd: TIL! Thanks @pabelanger. I will need to test that out and make sure it works for fifos that have already been read. If it does, that may be a safer solution here.

PR #616, opened 2021-01-19
Add test for socket implementation of remote job contract, add error handling

AlanCoding:
This is modeled after https://github.com/ansible/awx/pull/8598

~still very preliminary~

### DIFF
--- a/ansible_runner/interface.py
+++ b/ansible_runner/interface.py
@@ -43,7 +43,9 @@ def init_runner(**kwargs):
 
     See parameters given to :py:func:`ansible_runner.interface.run`
     '''
-    if not kwargs.get('cli_execenv_cmd'):
+    # If running via the transmit-worker-process method, we must only extract things as read-only
+    # inside of one of these commands. That could be either transmit or worker.
+    if not kwargs.get('cli_execenv_cmd') and (kwargs.get('streamer') not in ('worker', 'process')):
         dump_artifacts(kwargs)
 
     if kwargs.get('streamer'):
--- a/ansible_runner/streaming.py
+++ b/ansible_runner/streaming.py
@@ -5,6 +5,7 @@ import stat
 import sys
 import tempfile
 import uuid
+import traceback
 try:
     from collections.abc import Mapping
 except ImportError:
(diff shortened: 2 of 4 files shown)


AlanCoding:
I have cherry-picked https://github.com/ansible/ansible-runner/pull/575 into this PR, because it was impossible to move forward without it.

ansible_runner/interface.py
```diff
@@ -43,7 +43,9 @@ def init_runner(**kwargs):
 
     See parameters given to :py:func:`ansible_runner.interface.run`
     '''
-    if not kwargs.get('cli_execenv_cmd'):
+    # If running via the transmit-worker-process method, we must only extract things as read-only
+    # inside of one of these commands. That could be either transmit or worker.
+    if not kwargs.get('cli_execenv_cmd') and (kwargs.get('streamer') not in ('worker', 'process')):
```
AlanCoding, line 48: This is an important fix here. Summary:

```
$ ls -la test/data/debug/env
total 12
drwx------. 2 alancoding alancoding 4096 Jan 20 14:16 .
drwxrwxr-x. 5 alancoding alancoding 4096 Jan 20 14:16 ..
-r--------. 1 alancoding alancoding   23 Jan 20 14:16 envvars
```

When using `ansible_runner.interface.run`, it will write things to a file. Here, it writes the contents of the key "envvars" to the file `env/envvars`. It also wants file permissions for that to be `700`.

It's not a problem for `ansible-runner transmit` to write this `700` file. It's not a problem for it to zip it, send it, and to unpack it elsewhere.

However, `ansible-runner worker` would hit `dump_artifacts` here _before_ extracting the zip file. That means that it would be trying to extract the contents _overriding_ the `700` file that it just wrote. This is a problem.

```
E           Exception: Traceback (most recent call last):
E             File "/home/alancoding/repos/ansible-runner/ansible_runner/streaming.py", line 111, in run
E               utils.unstream_dir(zip_data, self.private_data_dir)
E             File "/home/alancoding/repos/ansible-runner/ansible_runner/utils.py", line 109, in unstream_dir
E               archive.extract(info.filename, path=directory)
E             File "/usr/lib64/python3.9/zipfile.py", line 1616, in extract
E               return self._extract_member(member, path, pwd)
E             File "/usr/lib64/python3.9/zipfile.py", line 1687, in _extract_member
E               open(targetpath, "wb") as target:
E           PermissionError: [Errno 13] Permission denied: '/tmp/pytest-of-alancoding/pytest-91/test_remote_job_by_sockets_run0/for_worker/env/envvars'
```

It's fine for the commands transmit / worker / process to re-run a lot of stuff in the normal `ansible-runner run` code path, but not this. Only 1 command out of the set should dump artifacts.

This was discovered by the test added in this PR.

AlanCoding:
Changes to the code itself are fairly minor here. Zuul tests look to be passing and stable enough. I'm happy with how the tests look.

Marking ready for review.

shanemcd (approved):

Issue #611, opened 2021-01-13
Run Execution Environments as non-root user, even in rootless containers


shanemcd:
@kdelee reached out and expressed some concerns after reading https://www.redhat.com/sysadmin/rootless-podman-makes-sense.

When running as root inside of a rootless container, if a user manages to break out of the container, they will be able to read things on the host owned by the user that started the container.

To prevent this, a non-root user can be used inside of the container:

```
FROM centos:8

RUN useradd -u 1000 runner

USER runner
```

The UID here must be predictable / known so that you can do:

```
$ mkdir -p test
$ podman unshare chown 1000:1000 $PWD/test
```

Once that is done, the host-side directory will be owned by the mapped UID:

```
$ ll
total 0
drwxr-xr-x. 2 100999 100999 6 Jan 13 13:45 test
```

Demonstrating of using that directory inside of container as non-root user:

```
$ podman run -ti -v $PWD/test/:/runner:Z test bash
[runner@31b9ad8dfe15 /]$ touch /runner/test
[runner@31b9ad8dfe15 /]$ ls -la /runner/
total 0
drwxr-xr-x.  2 runner runner 18 Jan 13 18:47 .
dr-xr-xr-x. 18 root   root   42 Jan 13 18:47 ..
-rw-rw-r--.  1 runner runner  0 Jan 13 18:47 test
[runner@31b9ad8dfe15 /]$ exit
shanemcd@fedora:~/podman
 $ ls -la test
total 0
drwxr-xr-x. 2   100999   100999 18 Jan 13 13:47 .
drwxr-xr-x. 3 shanemcd shanemcd 39 Jan 13 13:47 ..
-rw-rw-r--. 1   100999   100999  0 Jan 13 13:47 test
```

To get back to something you can delete as the host user:

```
$ podman unshare chown -R 0:0 $PWD/test
```

.... or just run:

```
$ podman unshare rm -rf $PWD/test
````

**Comment by shanemcd** [comment]
This will have several touch points:

This will need to go in the base `ansible-runner` image:

```
RUN useradd -u 1000 runner
```

This will need to be the final line of any `Containerfile` written by `ansible-builder`:

```
USER runner
```

In AWX, we currently clean up private data directories like this: https://github.com/ansible/awx/blob/74daa49e6f6ad97cbd57691a6ee41fd4709386fb/awx/main/tasks.py#L899

We will need to do something like `podman unshare rm -rf $private_data_dir` instead.

[...]

PR #617, opened 2021-01-20
Allow for running EEs as non-root user in rootless podman

shanemcd:
Potential solution for https://github.com/ansible/ansible-runner/issues/611

#### Podman

```
$ CONTAINER_ENGINE=podman make image
$ ansible-runner run --process-isolation demo -p test.yml
$ ansible-runner playbook demo/project/test.yml -i 'localhost,' -c local
```

#### Docker

```
$ make image
$ ansible-runner run --process-isolation --process-isolation-executable=docker demo -p test.yml
$ ansible-runner playbook --keep-files --container-runtime=docker demo/project/test.yml -i 'localhost,' -c local
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,6 +25,9 @@ RUN /output/install-from-bindep \
 # Prepare the /runner folder, seed the folder with demo data
 ADD demo /runner
 
+RUN groupadd -g 555 runner && useradd -u 555 -g 555 runner && \
+    usermod -aG root runner
+
 # In OpenShift, container will run as a random uid number and gid 0. Make sure things
 # are writeable by the root group.
 RUN for dir in \
@@ -52,6 +55,8 @@ ENV HOME=/home/runner
 
 ADD utils/entrypoint.sh /bin/entrypoint
 RUN chmod +x /bin/entrypoint
+
+USER runner
 
 ENTRYPOINT ["entrypoint"]
 CMD ["ansible-runner", "run", "/runner"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,8 +25,7 @@ RUN /output/install-from-bindep \
 # Prepare the /runner folder, seed the folder with demo data
 ADD demo /runner
 
-RUN groupadd -g 555 runner && useradd -u 555 -g 555 runner && \
-    usermod -aG root runner
+RUN useradd runner && usermod -aG root runner
 
 # In OpenShift, container will run as a random uid number and gid 0. Make sure things
 # are writeable by the root group.
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
 ARG ANSIBLE_CORE_IMAGE=quay.io/ansible/ansible-core:latest
 ARG PYTHON_BUILDER_IMAGE=quay.io/ansible/python-builder:latest
+ARG ANSIBLE_BRANCH=""
+ARG ZUUL_SIBLINGS=""
 
 FROM $PYTHON_BUILDER_IMAGE as builder
 # =============================================================================
 
-ARG ANSIBLE_BRANCH=""
-ARG ZUUL_SIBLINGS=""
 COPY . /tmp/src
 RUN if [ "$ANSIBLE_BRANCH" != "" ] ; then \
       echo "Installing requirements.txt / upper-constraints.txt for Ansible $ANSIBLE_BRANCH" ; \
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -157,6 +157,22 @@ class Runner(object):
             env_file_host = os.path.join(self.config.artifact_dir, 'env.list')
             with open(env_file_host, 'w') as f:
                 f.write('\n'.join(list(self.config.env.keys())))
+
+            if 'podman' in self.config.process_isolation_executable:
+                for root, dirs, files in os.walk(self.config.artifact_dir):
+                    dir_perms = (stat.S_IRUSR |
+                                 stat.S_IWUSR |
+                                 stat.S_IXUSR |
+                                 stat.S_IRGRP |
+                                 stat.S_IWGRP |
+                                 stat.S_IXGRP |
+                                 stat.S_ISGID)
+
+                    os.chmod(root, dir_perms)
+
+                    for d in dirs:
+                        os.chmod(os.path.join(root, d), dir_perms)
+
         else:
             cwd = self.config.cwd
             pexpect_env = self.config.env
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -158,7 +158,7 @@ class Runner(object):
             env_file_host = os.path.join(env_dir, 'env.list')
             with open(env_file_host, 'w') as f:
                 f.write('\n'.join(
-                     [f"{k}={v}" for k, v in self.config.env.items()]
+                    [f"{k}={v}" for k, v in self.config.env.items()]
                 ))
 
             # This here is the magic sauce. The user inside of the container is in the root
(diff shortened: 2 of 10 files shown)


ansible_runner/runner_config.py
```diff
@@ -186,7 +186,9 @@ def prepare(self):
         if self.module and self.playbook:
             raise ConfigurationError("Only one of playbook and module options are allowed")
         if not os.path.exists(self.artifact_dir):
-            os.makedirs(self.artifact_dir, mode=0o700)
+            original_umask = os.umask(0)
```
shanemcd, line 189: I was surprised that the permissions passed to `os.makedirs` seemed to be overridden by the system's default umask without this. Would it be better to do an explicit chmod after creating the directory instead?

Dockerfile
```diff
@@ -24,6 +24,9 @@ RUN /output/install-from-bindep \
 # Prepare the /runner folder, seed the folder with demo data
 ADD demo /runner
 
+RUN groupadd -g 555 runner && useradd -u 555 -g 555 runner && \
```
pabelanger: Should we consider making this configurable? I am thinking of downstream, and could be possible they may be different.

shanemcd: Come to think of it, I don't think we need to specifically set this at all anymore. My original approach was using `podman unshare`, which I promptly noped out of. Let me test with it.

Dockerfile
```diff
@@ -52,5 +55,7 @@ ENV HOME=/home/runner
 ADD utils/entrypoint.sh /bin/entrypoint
 RUN chmod +x /bin/entrypoint
 
+USER runner
```
pabelanger, line 58: bikeshed

should we consider an 'ansible' user over runner? I don't have a preference myself, but I could see some folks maybe wanting it.

ansible_runner/runner.py
```diff
@@ -149,9 +149,33 @@ def run(self):
             pexpect_env.update(self.config.env)
```
AlanCoding: Remove this line because it's no longer needed please.

ansible_runner/runner.py
```diff
@@ -145,13 +145,35 @@ def run(self):
...
+                     [f"{k}={v}" for k, v in self.config.env.items()]
+                ))
+
+            # This here is the magic sauce. The user inside of the container is in the root
+            # group, which is mapped to the host user's gid. By setting setgid on the artifacts
+            # dir + any subdirectories, any files created inside them will belong to the
+            # (in-container) root / host user's group, rather than the container user's gid.
+            # This allows runner's event_handler function to interact with these files outside
+            # the container.
+            if 'podman' in self.config.process_isolation_executable:
```
AlanCoding, line 174: Does this stuff really apply to podman but not docker?

ansible_runner/runner.py
```diff
@@ -145,13 +145,35 @@ def run(self):
...
+                for root, dirs, files in os.walk(self.config.artifact_dir):
+                    dir_perms = (stat.S_IRUSR |
+                                 stat.S_IWUSR |
+                                 stat.S_IXUSR |
+                                 stat.S_IRGRP |
+                                 stat.S_IWGRP |
+                                 stat.S_IXGRP |
+                                 stat.S_ISGID)
+
+                    os.chmod(root, dir_perms)
```
AlanCoding: it would be cleaner to put this before the `os.walk` loop, unless I'm missing something.

ansible_runner/runner.py
```diff
@@ -145,13 +145,35 @@ def run(self):
...
+                ))
+
+            # This here is the magic sauce. The user inside of the container is in the root
+            # group, which is mapped to the host user's gid. By setting setgid on the artifacts
+            # dir + any subdirectories, any files created inside them will belong to the
+            # (in-container) root / host user's group, rather than the container user's gid.
+            # This allows runner's event_handler function to interact with these files outside
+            # the container.
+            if 'podman' in self.config.process_isolation_executable:
+                for root, dirs, files in os.walk(self.config.artifact_dir):
```
AlanCoding: You `chmod` the root folder separately, which makes me think that you could make this more targeted by replacing `self.config.artifact_dir` with `os.path.join(self.config.artifact_dir, str(self.config.ident))`

If not, you're modifying permissions of artifacts from all previous jobs, which may not have even been ran in a container.

ansible_runner/runner.py
shanemcd: It's possible it might apply also to rootless docker. But haven't tested that. For normal docker, this is not needed.

ansible_runner/runner.py
shanemcd: I don't think it really matters, but sure.

ansible_runner/runner.py
shanemcd: `self.config.artifact_dir` contains the part `ident` already.

AlanCoding:
:+1:  I feel great about how the code looks now, but I haven't had the time to fully run through the image build and testing.

shanemcd:
@AlanCoding One thing we'll need to change in `ansible-builder` will be set `USER root` at the beginning of the build and flip back to `USER runner` at the end.

AlanCoding:
I wonder how that will work with prepend / append steps...

shanemcd:
Would it be better to:

a) put `USER root` before the prepend steps
b) leave `USER runner` out of the ansible-runner image altogether?

AlanCoding:
Option (a) sounds perfectly fine to me, I just wanted to have a plan out there.

AlanCoding:
I'm trying to test this according to the instructions I put together in https://gist.github.com/AlanCoding/a61ec24e35b980dca27f5e2ddb553a1d

Then I hit a wall in the part of the bash script where I built the awx-ee image.

```
podman build -f awx-ee/Dockerfile awx-ee -t quay.io/ansible/awx-ee:latest
```

output:

```
STEP 1: FROM quay.io/ansible/ansible-runner:devel AS galaxy
STEP 2: ARG ANSIBLE_GALAXY_CLI_COLLECTION_OPTS=
--> a69020302a9
STEP 3: ADD _build /build
--> 4e73ccd5322
STEP 4: WORKDIR /build
--> e16320264a0
STEP 5: RUN ansible-galaxy role install -r requirements.yml --roles-path /usr/share/ansible/roles
[WARNING]: You are running the development version of Ansible. You should only
run Ansible from "devel" if you are modifying the Ansible engine, or trying out
features under development. This is a rapidly changing source of code and can
become unstable at any point.
Skipping install, no requirements found
--> 9da2f712ed6
STEP 6: RUN ansible-galaxy collection install $ANSIBLE_GALAXY_CLI_COLLECTION_OPTS -r requirements.yml --collections-path /usr/share/ansible/collections
[WARNING]: You are running the development version of Ansible. You should only
run Ansible from "devel" if you are modifying the Ansible engine, or trying out
features under development. This is a rapidly changing source of code and can
become unstable at any point.
Starting galaxy collection install process
ERROR! Unexpected Exception, this is probably a bug: [Errno 13] Permission denied: b'/usr/share/ansible'
to see the full traceback, use -vvv
STEP 7: FROM quay.io/ansible/python-builder:latest AS builder
Error: error building at STEP "RUN ansible-galaxy collection install $ANSIBLE_GALAXY_CLI_COLLECTION_OPTS -r requirements.yml --collections-path /usr/share/ansible/collections": error while running runtime: exit status 250
```

At a quick glance, it looks like that might be the exact thing we just discussed.

> One thing we'll need to change in ansible-builder will be set USER root at the beginning of the build and flip back to USER runner at the end.

Obviously, that means that we need an ansible-builder branch which corresponds to this. I'll go try to make one. It will be a good chance to use the "depends-on" magic.

AlanCoding:
I started to make a little progress on this, but I have realized that we need changes to _both_ the `builder` and `galaxy` stages of the EE Containerfile. So this will still take some more time.

shanemcd:
recheck

shanemcd:
Idea has been abandoned.

PR #626, opened 2021-01-29
Allow for work to continue even if a bad utf-8 byte sequence is hit

jbradberry:


ryanpetrello (approved):
@jbradberry were you ever able to come up with some contrived way to reproduce this (if not, that's fine - the PR seems reasonable either way)?

jbradberry:
@ryanpetrello still working on it, it's tricky.  It looks like you have to have an invalid byte sequence right at the boundary of the read.  My naive reproducer, where I output a string encoded in Shift-JIS, works fine even before this change.

**Edit:** I couldn't make that trigger it either.

jbradberry:
regate

AlanCoding:
We supposedly have a file from the logs that created this issue. Worse comes to worse, we should be able to add that file as a data file in tests, and then create a mock worker that emits a stream from reading that file and dumping it to a buffer. I do think we need to get tests for this, however difficult that is.

jbradberry:
> We supposedly have a file from the logs that created this issue.

Not really.  What we have is the output _after_ it has already been decoded and the bad sequences replaced.

PR #628, opened 2021-01-29
Base64 encode zip data sent over streaming protocol

shanemcd:
We can argue about how to optimize this code later. But this solves our errors for now.


shanemcd:
Perhaps we should consider using https://pypi.org/project/base64io/

ryanpetrello (approved):

ryanpetrello:
We might also consider https://docs.python.org/3/library/base64.html#base64.b85encode, which is what git uses for binary patches, as it's a bit *more* efficient than base64.

PR #629, opened 2021-02-04
Record Information Related to Installed Collections + Collections Version

beeankha:
Merging blocked by: https://github.com/ansible/ansible/pull/73474
~~Test with: https://github.com/ansible/ansible-core-image/pull/3~~

### DIFF
--- a/utils/entrypoint.sh
+++ b/utils/entrypoint.sh
@@ -32,9 +32,9 @@ if [[ -n "${LAUNCHED_BY_RUNNER}" ]]; then
     export ANSIBLE_CALLBACK_PLUGINS="$(dirname $RUNNER_CALLBACKS)"
 fi
 
-if [[ -n "${AWX_ISOLATED_DATA_DIR}" ]]; then
+if [[ -d ${AWX_ISOLATED_DATA_DIR} ]]; then
     if output=$(ansible-galaxy collection list --format json 2> /dev/null); then
-        echo $output > "${AWX_ISOLATED_DATA_DIR}/collections.json"
+        echo $output > ${AWX_ISOLATED_DATA_DIR}/collections.json
     fi
 fi
 
--- a/utils/entrypoint.sh
+++ b/utils/entrypoint.sh
@@ -32,4 +32,10 @@ if [[ -n "${LAUNCHED_BY_RUNNER}" ]]; then
     export ANSIBLE_CALLBACK_PLUGINS="$(dirname $RUNNER_CALLBACKS)"
 fi
 
+if [[ -n "${AWX_ISOLATED_DATA_DIR}" ]]; then
+    if output=$(ansible-galaxy collection list --format json 2> /dev/null); then
+        echo $output > "${AWX_ISOLATED_DATA_DIR}/collections.json"
+    fi
+fi
+
 exec /usr/local/bin/dumb-init -- "${@}"


ansible_runner/runner.py
```diff
@@ -286,6 +286,20 @@ def _delete(retries=15):
                         raise
                 return True
             _delete()
+
+        # Record installed Collections + Collections version info
+        cmd = 'ansible-galaxy collection list -o json'
+        proc = Popen(cmd, stdout=PIPE, stderr=PIPE, shell=True)
```
ryanpetrello: We control these inputs *now*, so it's not a huge deal, but unless we _need it_ (and I suspect we don't), we should consider *not* doing `shell=True` here (and below):

https://docs.python.org/3/library/subprocess.html#security-considerations

ansible_runner/runner.py
```diff
@@ -286,6 +286,20 @@ def _delete(retries=15):
...
+        stdout, stderr = proc.communicate()
+        if proc.returncode:
+            # Unable to create cgroup
```
ryanpetrello: This comment is copy-pasta

ansible_runner/runner.py
```diff
@@ -286,6 +286,20 @@ def _delete(retries=15):
...
+            logger.error('Unable to list Collections: {}'.format(stderr))
+            raise RuntimeError('Unable to list Collections: {}'.format(stderr))
+        else:
+            logger.info("Listing Collections '{}'".format(stdout))
+        with open(os.path.join(self.config.artifact_dir, "collections.json"), "wb") as f:
```
ryanpetrello: Instead of running popen and managing the data via a pipe, you could just have Popen write directly to your file handle, and skip some of this:

```python
>>> subprocess.Popen(['ls', '-la'], stdout=open('testing123', 'wb')).communicate()
(None, None)
>>>
~ cat testing123
```

beeankha: We want to keep this as-is so we can detect if there's a return code, and only create the `collections.json` file if the command was successful.

ansible_runner/display_callback/module.py
```diff
@@ -167,6 +170,17 @@ def clear_task(self, local=False):
 
     def v2_playbook_on_start(self, playbook):
         self.set_playbook(playbook)
+
+        # Record installed Collections + Collections version info
+        cmd = "ansible-galaxy collection list --format json"
```
shanemcd: @matburt @ryanpetrello I didn't know where else to call this command to record the data we need. On a scale from 1-10, how dumb is this? We need to do it somewhere inside of the callback system so that we execute it inside of the EE. Halp.

shanemcd: If this _is_ an ok place to handle this, but the issue is mostly with the subprocess, maybe we need to reconsider writing out this metadata at build time. This would cause us to lose insight into collections installed via project updates.

ryanpetrello: Yea, the notion to me of spawning another process *inside* of this one is a little gross to me 😞

matburt: oof yeah, I don't think we can do this.

@ryanpetrello and I discussed this brad in the context of some other tooling yesterday.... both embedding this information in the EE OR potentially calling it as a pre-run command before other tasks.

shanemcd: I actually had another idea last night after thinking about this more... We'll probably just end up putting this in Runner's entrypoint. Unless one of you have a better idea @ryanpetrello @matburt

ryanpetrello: Yea, I think runner's entry point is a valid option.

shanemcd:
recheck

ryanpetrello (approved):

shanemcd (approved):

shanemcd:
regate

beeankha:
regate

Issue #637, opened 2021-03-01
Ansible runner does not write event files atomically


adamruzicka:
There is a small time window where the actual final event json file can be incomplete and therefore contain invalid json, tripping up tools which rely on those files always being complete. It should write to a temporary location and then atomically move the now-complete file to the final destination the same way it is done here[1].

[1] - https://github.com/ansible/ansible-runner/blob/devel/ansible_runner/display_callback/events.py#L69

adamruzicka:
Linked PR was merged, closing.

PR #638, opened 2021-03-01
Write event files atomically

adamruzicka:
There used to be a small time window where the actual final event json file
could be incomplete and therefore contain invalid json, tripping up tools which
rely on those files always being complete.

#637


shanemcd:
recheck

shanemcd (approved):

PR #652, opened 2021-03-26
Remove demo directory from container image

shanemcd:
This is proving to be problematic for us in AWX 18. I'm trying to use /runner as
the private data dir location, but since stuff is already there, the new
ansible-runner streaming interface is blowing up.


matburt (approved):

pabelanger:
This timed out, but another PR bumps the jobs to a little longer.

PR #657, opened 2021-03-31
Make sure we import from non-deprecated module

AlanCoding:
https://github.com/ansible/ansible-runner/blob/830c82809fbb37085c09bdfc1f51fe6c2158e826/ansible_runner/runner_config.py#L20-L21

Just as a matter of consistent messaging I'd prefer to nuke the old import pattern.

Ping @ganeshrn

### DIFF
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -3,7 +3,7 @@ import shutil
 
 import pytest
 import pexpect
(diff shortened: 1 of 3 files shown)


shanemcd (approved):

ganeshrn:
recheck

PR #662, opened 2021-04-06
Add missing ARG ANSIBLE_BRANCH

pabelanger:
We were referecing a missing ARG, which breaks on newer versions of
podman.


pabelanger:
Pull-request updated, HEAD is now 1979c85af77cf02f645a375c40f801cb82d6da9d

Issue #604, opened 2021-01-06
Do not load private_data_dir into memory for receptor transmission


AlanCoding:
This is a followup from the conversation in:

https://github.com/ansible/ansible-runner/pull/505#discussion_r500518372

As of today, this method still works with an in-memory buffer.

https://github.com/ansible/ansible-runner/blob/ae4cd96a2568d0e31695bf98f5a374e5efa660dc/ansible_runner/utils.py#L87

This is a non-starter for many users of AWX, because they have projects that run into the 20 GB territory and this will crash. This seems to me like a hard release blocker.

#### Fix option - send directly to standard out

I attempted a fix in the branch here:

https://github.com/ansible/ansible-runner/compare/devel...AlanCoding:direct_to_pipe

The idea here is that we completely do away with the in-memory buffer and write directly to the output stream. This means that we don't have the full length of the message before we send it. I don't know if that can be worked around. This also had problems grappling with the general way that the zipfile library worked, and it might not be workable.

#### Fix option - use a file on disk an intermediary

We know pretty well that this alternative will work - write the data to a file instead of a `BytesIO()` object. Then we can obtain the size of it without reading the entire contents, and we can write it to standard out line-by-line without storing the entire thing in memory.

kdelee:
Using this in awx, sending a large project (8.something GB of data on a machine that only has 8GB of memory) to a container group, we saw the disk usage go up indicating the file was being staged and then watched the network traffic as the file got sent. Then in the job in the container group we saw  that the files were indeed present..


Keep in mind the tower node needs to have more than 2x of that the project consumes. So if the project is 20GB, they need to have at least 21GB of free disk or they will run out of disk.

![Screenshot from 2021-05-26 16-25-44](https://user-images.githubusercontent.com/29239279/119726816-8c3da780-be3f-11eb-81f6-804cc38b08b9.png)

PR #670, opened 2021-04-14
Avoid loading zip data into memory

shanemcd:
Link: https://github.com/ansible/ansible-runner/issues/604
Replaces: https://github.com/ansible/ansible-runner/pull/605

### DIFF
--- a/ansible_runner/utils/__init__.py
+++ b/ansible_runner/utils/__init__.py
@@ -14,14 +14,12 @@ import threading
 import pipes
 import uuid
 import codecs
-import zipfile
-import math
 
 try:
     from collections.abc import Iterable, Mapping
 except ImportError:
     from collections import Iterable, Mapping
-from io import BytesIO, StringIO
+from io import StringIO
 from six import string_types, PY2, PY3, text_type, binary_type
 
 
@@ -82,6 +80,7 @@ def check_isolation_executable_installed(isolation_executable):
         if isinstance(e, ValueError) or getattr(e, 'errno', 1) != 2:  # ENOENT, no such file or directory
             raise RuntimeError(f'{isolation_executable} unavailable for unexpected reason.')
         return False
+
 
 def dump_artifact(obj, path, filename=None):
     '''
--- a/ansible_runner/utils/__init__.py
+++ b/ansible_runner/utils/__init__.py
@@ -15,6 +15,7 @@ import pipes
 import uuid
 import codecs
 import zipfile
+import math
 
 try:
     from collections.abc import Iterable, Mapping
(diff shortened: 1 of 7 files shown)


ansible_runner/utils/__init__.py
```diff
@@ -15,6 +15,7 @@
 import uuid
 import codecs
 import zipfile
+import math
```
jbradberry: ???

ansible_runner/utils/streaming.py
```diff
@@ -0,0 +1,59 @@
...
+                    encoded_target.write(line)
+
+
+def unstream_dir(stream, length, target_directory):            
+    # NOTE: caller needs to process exceptions
+    with tempfile.NamedTemporaryFile() as tmp:
+        with open(tmp.name, "wb") as target:
+            with Base64IO(stream) as source:                
+                while True:
+                    remaining = length
```
jbradberry: this is in the loop, so it's basically always going to be getting reset to the full size

ansible_runner/utils/streaming.py
```diff
@@ -0,0 +1,59 @@
...
+                    chunk_size = 1024 * 1000  # 1 MB
+                    if chunk_size >= remaining:
+                        chunk_size = remaining
+                    data = source.read(chunk_size)
+                    target.write(data)
+                    remaining -= chunk_size
+
+                    if remaining == 0:
```
jbradberry: the end condition can be folded into the `while` statement

ansible_runner/utils/streaming.py
```diff
@@ -0,0 +1,59 @@
...
+            if source_directory:
+                for dirpath, dirs, files in os.walk(source_directory):
+                    relpath = os.path.relpath(dirpath, source_directory)
+                    if relpath == ".":
+                        relpath = ""
+                    for fname in files:
+                        archive.write(os.path.join(dirpath, fname), arcname=os.path.join(relpath, fname))
+            archive.close()
+
+        zip_size = Path(tmp.name).stat().st_size
```
jbradberry, line 23: wait, did you lose the calculation of tarball -> base64 size?

ansible_runner/utils/streaming.py
shanemcd: Nice catch!

ansible_runner/utils/streaming.py
shanemcd: Yep! It turns out that `base64io` is smart enough that when you say you need to read `n` number of bytes, it gives you that many bytes of _unencoded_ data. I thought that was pretty neat.

ansible_runner/utils/__init__.py
shanemcd: Ooops.  Left over from a failed experiment.

shanemcd:
recheck

shanemcd:
recheck

jbradberry (approved):

PR #675, opened 2021-04-24
Fix ZUUL_SIBLINGS and ANSIBLE_BRANCH ARGs

pabelanger:
We didn't correctly expose these 2 ARGs for our Dockerfile to work
properly. This should mean depends-on again for Zuul CI works, along
with the proper version of ansible in our images.


pabelanger:
recheck

shanemcd (approved):

PR #680, opened 2021-04-28
fix duplicate volume mounts for containers, simplify conditional

maxamillion:
Uniformally handle dir paths that do have a trailing / and don't, but
are in fact the same path which are passed in or detected as duplicates
when constructing the list of volume mounts for containerized process
 isolation.

Previously the BaseConfig._update_volume_mount_paths() method was using
this expression to check for a potentially duplicate path but the extra
processing of the map() and join() function calls are not needed since
the format src_mount_path:dest_mount_path stored in volume_mount_path is
already an element of arg_list and no transformation is necessary.

```
-        if ', '.join(map(str, ['-v', volume_mount_path])) not in ', '.join(map(str, args_list)):
+        if volume_mount_path not in args_list:
```

### DIFF
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -314,6 +314,14 @@ class BaseConfig(object):
 
         self._ensure_path_safe_to_mount(src_mount_path)
 
+        def _add_trailing_slash_if_neeeded(some_path):
+            if os.path.isdir(src_mount_path):
+                return some_path + '/' if (some_path[-1] != '/') else some_path
+            else:
+                return some_path
+
+        src_mount_path = _add_trailing_slash_if_neeeded(src_mount_path)
+
         if os.path.isabs(src_mount_path):
             if os.path.isdir(src_mount_path):
                 volume_mount_path = "{}:{}".format(src_mount_path, dest_mount_path)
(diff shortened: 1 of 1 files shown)


ansible_runner/config/_base.py
```diff
@@ -314,6 +314,14 @@ def _update_volume_mount_paths(self, args_list, src_mount_path, dest_mount_path=
 
         self._ensure_path_safe_to_mount(src_mount_path)
 
+        def _add_trailing_slash_if_neeeded(some_path):
```
ganeshrn: Can we move this method as a level class method? IMO defining a method within a method makes code less readable.

ansible_runner/config/_base.py
```diff
@@ -314,6 +314,14 @@ def _update_volume_mount_paths(self, args_list, src_mount_path, dest_mount_path=
...
+            if os.path.isdir(src_mount_path):
```
ganeshrn: Should this be `some_path` instead of `src_mount_path`

ansible_runner/config/_base.py
```diff
@@ -314,6 +314,14 @@ def _update_volume_mount_paths(self, args_list, src_mount_path, dest_mount_path=
...
+                return some_path + '/' if (some_path[-1] != '/') else some_path
```
shanemcd, line 319: Would this break cases where we might actually want to mount a file instead of a directory?

shanemcd:
@maxamillion it looks like theres a teeny little test failure on your PR but otherwise wfm

ansible_runner/config/_base.py
maxamillion: it sure should! thanks :)

ansible_runner/config/_base.py
maxamillion: Sure thing!

ganeshrn (approved):
Thank you!

ganeshrn:
recheck

ganeshrn:
Fixed by PR https://github.com/ansible/ansible-runner/pull/691

PR #682, opened 2021-05-04
Str runner config

didib:
Add __str__ to RunnerConfig.

This was useful for me when working on [some oVirt patch](https://gerrit.ovirt.org/c/ovirt-system-tests/+/114423)

### DIFF
--- a/ansible_runner/config/runner.py
+++ b/ansible_runner/config/runner.py
@@ -170,6 +170,38 @@ class RunnerConfig(object):
     def containerized(self):
         return self.process_isolation and self.process_isolation_executable in self._CONTAINER_ENGINES
 
+    def __str__(self):
+        attrs = (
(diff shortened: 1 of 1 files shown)


didib:
Sorry, this was on top of 'master' branch. I understand development is now on 'devel' branch, will rebase.

didib:
Ok, I think this should work. I only verified the first version I pushed, which I applied on top of the version included in RHEL8 (python3-ansible-runner-1.4.6-1.el8.noarch). Changes since then:
1. Rebase on devel branch
2. Fix the string to point at the new location
3. Add some attrs that were added since then

matburt:
I realize this has been hanging around for a while. My intuition is that this isn't going to scale as parameters are added. I like to see more direct introspection of the config as a better mechanism.

PR #689, opened 2021-05-11
Refactoring and purging playbook and ad-hoc commands

matburt:
We'll ship these over to navigator and Runner will stay exclusively as an interface

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -550,11 +550,6 @@ def print_common_usage():
                 ansible-runner run . -r role_name --hosts myhost
                 ansible-runner run . -m command -a "ls -l" --hosts myhost
 
-            run ansible execution environments:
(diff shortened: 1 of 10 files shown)


pabelanger:
Can we hold off on doing this, until we know that ansible-navigator will work properly taking a playbook via the CLI?

I am dealing with an issue locally (can reproduce in CI) which navigator dying using playbooks with stdout.

matburt:
What's the issue? I think there's an inevitability that we will shift this over there and it's currently unused.

ganeshrn:
@pabelanger Can you please try with the latest `ansible-navigtor` code? The issue is most likely fixed by PR https://github.com/ansible/ansible-navigator/pull/340
Also confirmed with @nilashish, he is not able to see the issue with latest navigator code.

PR #690, opened 2021-05-11
surrounded ZipFile write() method in try/except block which raises a custom Exception

sarabrajsingh:
custom error raised when an invalid symlink is detected

### DIFF
--- a/ansible_runner/utils/streaming.py
+++ b/ansible_runner/utils/streaming.py
@@ -10,34 +10,39 @@ from pathlib import Path
 
 def stream_dir(source_directory, stream):
     with tempfile.NamedTemporaryFile() as tmp:
-        with zipfile.ZipFile(tmp.name, 'w', compression=zipfile.ZIP_DEFLATED, allowZip64=True) as archive:
+        with zipfile.ZipFile(
+            tmp.name, "w", compression=zipfile.ZIP_DEFLATED, allowZip64=True
+        ) as archive:
             if source_directory:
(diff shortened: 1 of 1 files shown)


ansible_runner/utils/streaming.py
```diff
@@ -17,7 +17,11 @@ def stream_dir(source_directory, stream):
                     if relpath == ".":
                         relpath = ""
                     for fname in files:
-                        archive.write(os.path.join(dirpath, fname), arcname=os.path.join(relpath, fname))
+                        try:
+                            archive.write(os.path.join(dirpath, fname), arcname=os.path.join(relpath, fname))
+                        except FileNotFoundError as e:
+                            e.strerror = "invalid symlink detected"
```
jbradberry: I think it would be better to raise a new exception with its own message than modifying an existing exception.

That said, perhaps we're going about this wrong.  We should probably just do an `os.path.isfile()` or something like that before doing the `archive.write()`.

sarabrajsingh: removed try/except for `if conditional` in latest commit `d9285f6`

jbradberry (commented):
It looks like you've got a couple of problems with whitespace, but I think this is the right idea.

jbradberry:
I'm kind of surprised that black didn't automatically fix these.

jbradberry (approved):

shanemcd (approved):

PR #696, opened 2021-05-18
Add return code for run_command interface

ganeshrn:
*  Update the return value to run_command
   interface API to return the return code
   from the process that excecutes the command

### DIFF
--- a/ansible_runner/interface.py
+++ b/ansible_runner/interface.py
@@ -366,14 +366,15 @@ def run_command(executable_cmd, cmdline_args=None, **kwargs):
     :type status_handler: function
     :type artifacts_handler: function
 
-    :returns: Returns a tuple of return code, response and error string. In case if ``runner_mode`` is set to ``pexpect`` the error value is empty as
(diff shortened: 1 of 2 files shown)


ansible_runner/interface.py
```diff
@@ -366,14 +366,14 @@ def run_command(executable_cmd, cmdline_args=None, **kwargs):
     :type status_handler: function
     :type artifacts_handler: function
 
-    :returns: Retunes a tuple of response and error string. In case if ``runner_mode`` is set to ``pexpect`` the error value is empty as
+    :returns: Retunes a tuple of return code, response and error string. In case if ``runner_mode`` is set to ``pexpect`` the error value is empty as
```
relrod:
```suggestion
    :returns: Returns a tuple of return code, response and error string. In case if ``runner_mode`` is set to ``pexpect`` the error value is empty as
```

matburt (approved):
I *might* suggest putting rc on the end to be additive but I'm mostly indifferent.

PR #700, opened 2021-05-19
bumped release version to 2.0.0a3

sarabrajsingh:


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "ansible-runner"
-version = "2.0.0a2"
+version = "2.0.0a3"
 description = "Consistent Ansible Pythin API and CLI with container and process isoluation runtime capabilities"
```
jbradberry: I know this isn't your fault, but: `Pythin`, `isoluation`

jbradberry:
We should probably also bring the changelog up to date, since 1.4.7 failed to make it in there :face_with_head_bandage:

PR #704, opened 2021-05-21
Allow no event data to fix test

AlanCoding:


AlanCoding:
thinking about this more, I wonder if I should do a git bisect to see where this was introduced...

but introducing it should have been blocked by the Zuul jobs, so I don't know.

AlanCoding:
Yeah, I couldn't reproduce the failure locally. Tests are running latest Ansible, and so am I.

ganeshrn (approved):

PR #710, opened 2021-05-31
Add timeout parameter in the interface

ganeshrn:
*  All timeout parameter for pexpect/subprocess to be passed through
   runner interface
*  Remove default value for subprocess timeout as it results
   in timeout for long running playbooks

### DIFF
--- a/ansible_runner/interface.py
+++ b/ansible_runner/interface.py
@@ -165,7 +165,8 @@ def run(**kwargs):
     :param project_dir: The path to the playbook content, this defaults to 'project' within the private data dir
     :param rotate_artifacts: Keep at most n artifact directories, disable with a value of 0 which is the default
     :param timeout: The timeout value in seconds that will be passed to either ``pexpect`` of ``subprocess`` invocation
-                    (based on ``runner_mode`` selected) while executing command.
+                    (based on ``runner_mode`` selected) while executing command. It the timeout is triggered it will force cancel the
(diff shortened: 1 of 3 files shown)


matburt (changes requested):
I'd say instead of re-using the pexpect timeout look at `job_timeout` instead.

ganeshrn:
@matburt Review comment fixed in latest commit

ansible_runner/interface.py
```diff
@@ -164,6 +164,8 @@ def run(**kwargs):
     :param artifact_dir: The path to the directory where artifacts should live, this defaults to 'artifacts' under the private data dir
     :param project_dir: The path to the playbook content, this defaults to 'project' within the private data dir
     :param rotate_artifacts: Keep at most n artifact directories, disable with a value of 0 which is the default
+    :param timeout: The timeout value in seconds that will be passed to either ``pexpect`` of ``subprocess`` invocation
```
matburt, line 167: This is a good chance, one small tweak to this in that I'd like to explain what the effect of this will be if set... force canceling the execution in this case. I just want to make sure we call that out.

matburt (approved):
This looks good, make that change and apply the `gate` to merge it (you don't have to wait for me)

ansible_runner/interface.py
ganeshrn: Fixed

PR #713, opened 2021-06-03
Preserve symlinks in streaming interface

shanemcd:


ansible_runner/utils/streaming.py
```diff
@@ -19,14 +20,21 @@ def stream_dir(source_directory, stream):
...
-                            raise FileNotFoundError(
-                                "invalid symlink found; file={0}.".format(fullpath)
+                        full_path = os.path.join(dirpath, fname)
+                        # Magic to preserve symlinks
+                        if os.path.islink(full_path):
+                            archive_relative_path = os.path.relpath(dirpath, source_directory)
+                            file_relative_path = os.path.join(archive_relative_path, fname)
+                            zip_info = zipfile.ZipInfo(file_relative_path)
+                            zip_info.create_system = 3
+                            permissions = 0o777
```
chrismeyersfsu, line 30: Do unzipped files result in permission `777` ?

ansible_runner/utils/streaming.py
```diff
@@ -64,5 +72,11 @@ def unstream_dir(stream, length, target_directory):
...
                 archive.extract(info.filename, path=target_directory)
                 out_path = os.path.join(target_directory, info.filename)
-                perm = info.external_attr >> 16
-                os.chmod(out_path, perm)
+                perms = info.external_attr >> 16
+                mode = stat.filemode(perms)
+                if mode[:1] == 'l':
+                    link = open(out_path).read()
+                    os.remove(out_path)
+                    os.symlink(link, out_path)
```
chrismeyersfsu, line 87: We want to retain the perms from the original project, even on the symlink.

ansible_runner/utils/streaming.py
```diff
@@ -64,5 +72,11 @@ def unstream_dir(stream, length, target_directory):
             for info in archive.infolist():
                 archive.extract(info.filename, path=target_directory)
                 out_path = os.path.join(target_directory, info.filename)
-                perm = info.external_attr >> 16
-                os.chmod(out_path, perm)
+                perms = info.external_attr >> 16
+                mode = stat.filemode(perms)
+                if mode[:1] == 'l':
+                    link = open(out_path).read()
```
chrismeyersfsu, line 85: could OOM if a malicious user puts a bunch of content in the symlink data portion. There are probably other/better way to OOM the system than that. Besides, the project SCM should be scrutinized via review process "why has our git checkout time jumped from 30 seconds to 3 hours?"

ansible_runner/utils/streaming.py
shanemcd: I think this is just how symlinks work:

```
$ cd /tmp/
$ ln -sf /etc/passwd /tmp/hello
$ sudo su - anotheruser
Last login: Thu Jun  3 15:34:33 EDT 2021 on pts/27
$ ls -la /tmp/hello
lrwxrwxrwx. 1 shanemcd shanemcd 11 Jun  3 15:35 /tmp/hello -> /etc/passwd
$ rm /tmp/hello
rm: cannot remove '/tmp/hello': Operation not permitted
```

ansible_runner/utils/streaming.py
shanemcd: See my comment below.

ansible_runner/utils/streaming.py
shanemcd: https://unix.stackexchange.com/questions/32795/what-is-the-maximum-allowed-filename-and-folder-size-with-ecryptfs

>Linux has a maximum filename length of 255 characters for most filesystems (including EXT4), and a maximum path of 4096 characters.

AlanCoding (approved):
I contributed the tests, but did not modify the main source tree here.

All the specific concerns that were raised have been given test coverage. @chrismeyersfsu raised some points about file permissions, which have been addressed and tested to my satisfaction. Other points made here seem to be addressed as well.

New tests are passing locally, I'm happy

ansible_runner/utils/streaming.py
```diff
@@ -62,7 +70,20 @@ def unstream_dir(stream, length, target_directory):
...
                 out_path = os.path.join(target_directory, info.filename)
-                perm = info.external_attr >> 16
-                os.chmod(out_path, perm)
+                
+                perms = info.external_attr >> 16
+                mode = stat.filemode(perms)
+                
+                is_symlink = mode[:1] == 'l'
+                if is_symlink and os.path.exists(out_path):
+                    os.remove(out_path)
```
chrismeyersfsu, line 80: I wouldn't think we need this remove since we remove below and I wouldn't expect `out_path` to ever exist in this case.

PR #716, opened 2021-06-04
Add the latest tag to 2.11 image

AlanCoding:
Alternatively, maybe we could add a `release:` job.

I have some issues either way. Here, we are using `devel` of ansible-runner and `stable-2.11` branch of ansible/ansible.

It might be preferable, for the latest tag, to use the ansible-core on PyPI and the latest tagged ansible-runner. As of now, we don't have any ansible-runner tags that work for this, and I also don't know how to re-configure the job to Ansible in some way other than a checkout.

Maybe this works for now.

### DIFF
--- a/.zuul.d/jobs.yaml
+++ b/.zuul.d/jobs.yaml
@@ -122,7 +122,9 @@
           repository: quay.io/ansible/ansible-runner
           siblings:
             - github.com/ansible/ansible
-          tags: ['stable-2.11-devel']
+          tags:
+            - stable-2.11-devel
+            - latest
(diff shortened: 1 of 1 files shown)


AlanCoding:
Example of a release job

https://github.com/ansible/ansible-zuul-jobs/blob/9148c9a5e96e3760a8261956b9e7bf2e065aeea0/zuul.d/project-templates.yaml#L1640-L1642

shanemcd (approved):

AlanCoding:
Tag is showing up here https://quay.io/repository/ansible/ansible-runner?tab=tags

PR #717, opened 2021-06-04
Fix container image builds

shanemcd:
This is a complete hack. We now have a requirements.txt at the root of the repo, and the Dockerfile was overwriting it. We can clean this up later.


simaishi (approved):

shanemcd:
regate

PR #719, opened 2021-06-07
Raise exception if not able to read partial event data

### LINKED ISSUE #718: runner should fail if permission denied for reading event data from file
Related issue https://github.com/ansible/ansible-builder/issues/247

ansible-runner should not fail silently https://github.com/ansible/ansible-runner/blob/devel/ansible_runner/runner.py#L76 if in case the read permission is not present for reading event data from the partial temporary file.

ganeshrn:
Fixes https://github.com/ansible/ansible-runner/issues/718

*  Raise execption if user does not have permissions to read
   the partial event data file instead of silently ignoring it.


AlanCoding:
I thought it was the case that, for some event types, there is no partial data to combine with. For instance, normally runner_on_start events don't have any stdout so maybe they don't generate the partial data file. I'm not sure if it hits this in that scenario.

shanemcd (approved):

PR #722, opened 2021-06-10
If private_data_dir not passed default to temp location

### LINKED ISSUE #699: home directory .ansible-runner
Right now we make use of `~/.ansible-runner` : https://github.com/ansible/ansible-runner/blob/a151df7fc43c1ebfe9e72c1193603a3d3e50b046/ansible_runner/config/_base.py#L91

I think we might want to shift this to `/tmp/.ansible-runner`

and then add an `@property` to https://github.com/ansible/ansible-runner/blob/a151df7fc43c1ebfe9e72c1193603a3d3e50b046/ansible_runner/runner.py#L26 that points to the location of the private data dir.

ganeshrn:
Fixes https://github.com/ansible/ansible-runner/issues/699

*  Use temp location for `private_data_dir` if not passed
   in the interface API.

### DIFF
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -21,6 +21,7 @@ import os
 import pexpect
 import re
 
+from tempfile import gettempdir
 from uuid import uuid4
 try:
     from collections.abc import Mapping
@@ -87,7 +88,7 @@ class BaseConfig(object):
         if private_data_dir:
             self.private_data_dir = os.path.abspath(private_data_dir)
         else:
-            self.private_data_dir = os.path.abspath(os.path.expanduser('~/.ansible-runner'))
+            self.private_data_dir = os.path.join(gettempdir(), ".ansible-runner")
 
         if artifact_dir is None:
             artifact_dir = os.path.join(self.private_data_dir, 'artifacts')
--- a/test/unit/config/test__base.py
+++ b/test/unit/config/test__base.py
@@ -8,6 +8,8 @@ import six
 from pexpect import TIMEOUT, EOF
 
 import pytest
+
+from tempfile import gettempdir
 from unittest.mock import (Mock, patch, PropertyMock)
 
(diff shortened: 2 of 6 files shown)


ansible_runner/config/_base.py
```diff
@@ -87,7 +88,7 @@ def __init__(self,
         if private_data_dir:
             self.private_data_dir = os.path.abspath(private_data_dir)
         else:
-            self.private_data_dir = os.path.abspath(os.path.expanduser('~/.ansible-runner'))
+            self.private_data_dir = os.path.join(gettempdir(), ".ansible-runner")
```
AlanCoding, line 91: okay, so this is `/tmp/.ansible-runner` @matburt suggested out in the issue.

AlanCoding: err... it would be layered under another tmpdir, so maybe `/tmp/xxxxx/.ansible-runner`

shanemcd: I'm a _little_ worried about the possibility of cross-device errors here when ansible tries to do atomic writes.

shanemcd (changes requested):
I think I'd like to hold this until after the 2.0 release. We don't have the capacity to do test this right now, and I am worried about the risk associated with this patch.

matburt:
> I think I'd like to hold this until after the 2.0 release. We don't have the capacity to do test this right now, and I am worried about the risk associated with this patch.

ansible_runner/config/_base.py
matburt: I'm not sure I'm too terribly worried about that relative to the current working directory... `home` could just as easily be off on another device also. The original code, whose provenance is 2.0 adds some new assumptions which, in this case, will require a home directory to be set.

matburt:
My concern with leaving this as-is is that we're already setting a default private data dir, which is different than 1.X I'd like for that *not* to be rooted at home by default. temp space makes the most sense here.

matburt (approved):

PR #743, opened 2021-06-29
fix(make): removed unused arg and fixing tools/req/branch

atbore-phx:
Hi,
a simple fix for Makefile to build docker images using their defaults and the right --build-arg` values.

- fixing `make image_matrix`

removing devel from `ANSIBLE_VERSIONS` defaults; https://github.com/ansible/ansible-runner/tree/devel/tools doesn't contain "devel" requirements files so `make image_matrix` fails with default values.

- fixing `make image`

`RUNNER_VERSION` is not a Dokerfile ARG  replaced with `ANSIBLE_BRANCH`

`ANSIBLE_BRANCH` default value is "" to fit Dokerfile tools/requirement condition

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ NAME = ansible-runner
 IMAGE_NAME ?= quay.io/ansible/ansible-runner
 IMAGE_NAME_STRIPPED := $(word 1,$(subst :, ,$(IMAGE_NAME)))
 GIT_BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
-ANSIBLE_BRANCH ?= ""
+ANSIBLE_BRANCH ?=
 ANSIBLE_VERSIONS ?= stable-2.9 stable-2.10 stable-2.11
 PIP_NAME = ansible_runner
 VERSION := $(shell python setup.py --version)
@@ -97,9 +97,9 @@ image: sdist
 		-t $(IMAGE_NAME) -f Dockerfile .
 	$(CONTAINER_ENGINE) tag $(IMAGE_NAME) $(IMAGE_NAME_STRIPPED):$(GIT_BRANCH)
 
-image_matrix:
+image_matrix: image
 	for version in $(ANSIBLE_VERSIONS) ; do \
-		ANSIBLE_BRANCH=$$version GIT_BRANCH=$$version.$(GIT_BRANCH) make image ; \
+		ANSIBLE_BRANCH=$$version GIT_BRANCH=$$version-$(GIT_BRANCH) make image ; \
 	done
 	$(CONTAINER_ENGINE) tag $(IMAGE_NAME) $(IMAGE_NAME_STRIPPED):$(GIT_BRANCH)
 
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,8 @@ NAME = ansible-runner
 IMAGE_NAME ?= quay.io/ansible/ansible-runner
 IMAGE_NAME_STRIPPED := $(word 1,$(subst :, ,$(IMAGE_NAME)))
 GIT_BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
-ANSIBLE_BRANCH ?= devel
-ANSIBLE_VERSIONS ?= stable-2.9 stable-2.10 devel
+ANSIBLE_BRANCH ?= ""
+ANSIBLE_VERSIONS ?= stable-2.9 stable-2.10 stable-2.11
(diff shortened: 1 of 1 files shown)


atbore-phx:
Hi @AlanCoding,
What do you think about this PR ?
i'm available for any request.
Best

Makefile
```diff
@@ -6,8 +6,8 @@ NAME = ansible-runner
 IMAGE_NAME ?= quay.io/ansible/ansible-runner
 IMAGE_NAME_STRIPPED := $(word 1,$(subst :, ,$(IMAGE_NAME)))
 GIT_BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
-ANSIBLE_BRANCH ?= devel
-ANSIBLE_VERSIONS ?= stable-2.9 stable-2.10 devel
+ANSIBLE_BRANCH ?= ""
+ANSIBLE_VERSIONS ?= stable-2.9 stable-2.10 stable-2.11
```
AlanCoding, line 10: Actually, could we add back `devel` to the end of this? If we want to keep the number of versions small, we can remove 2.10. But fore now, I think it's easiest to put them all in.

atbore-phx: hi @AlanCoding thx for the reply,  yes of course for adding back !,  just two questions 
Do you mean to remove 2.9 (not 2.10) ?

What about this condition ? https://github.com/ansible/ansible-runner/blob/dcdb62daf668a31754fc6fbf73374e7d14d5f52c/Dockerfile#L12

if devel is an `ANSIBLE_VERSIONS` and/or  `ANSIBLE_BRANCH` element when we execute `make image_matrix` or `make image `,
make fails due the file requirements-**devel**.txt doesn't exist

Should i create the file ?  copying the content from 2.11 ?

AlanCoding: yeah, as you can image, `make image_matrix` and `make image` are not being used to produce the actual images on quay. The real logic is in the Zuul configs. I wish that it were easy to reuse that same stuff for user-oriented commands, but it's not. I'm not sure what to do about the requirements. The best solution would be to emulate what Zuul does, which requires some digging.

I did mean 2.10. But I'd hold off on that. 2.9 remains important for the time being.

atbore-phx: Ok perfect.   I'll check what Zuul does and I'll commit the changes.

atbore-phx:
hi @AlanCoding,
I've checked zuul builds,

from my understanding zuul launches four builds using `podman build .`  setting the docker tag and the `ANSIBLE_BRANCH` so inside the Dockerfile the build can chose the right requirement file.

I've modified the `Makefile` accordingly for a closest emulation  (same output tags and same requirements files).

 In Makefile `devel` should not be defined in `ANSIBLE_BRANCH` or `ANSIBLE_VERSION` because in the first zuul build (devel)
ANSIBLE_BRANCH is undefined. So the docker tag will be set from the git branch and version.

have a nice weekend !

P.S.

zuul builds details:

1. ansible-runner-build-container-image (two times) [https://dashboard.zuul.ansible.com/t/ansible/build/4b1c7288db4d4bfaa219a266a08afeb4/console#3/0/9/ubuntu-bionic](url)
`ANSIBLE_BRANCH` **undefined** 
tag: devel
req: tools/requirements.txt
2. ansible-runner-build-container-image-stable-2.9 https://dashboard.zuul.ansible.com/t/ansible/build/4c026018ac944ce69d0d3d551196ac43/console#3/0/9/ubuntu-bionic
`ANSIBLE_BRANCH`=stable-2.9
tag: stable-2.9-devel
req: tools/requirements-stable-2.9.txt 

3. ansible-runner-build-container-image-stable-2.10 https://dashboard.zuul.ansible.com/t/ansible/build/d7cf592ee65f4a3880db681cd6cd6b5a/console#3/0/9/ubuntu-bionic
ANSIBLE_BRANCH=stable-2.10
tag: stable-2.10-devel
req: tools/requirements-stable-2.10.txt 

4. ansible-runner-build-container-image-stable-2.11 https://dashboard.zuul.ansible.com/t/ansible/build/8497d07812be48abbdb206648891af61/console#3/0/9/ubuntu-bionic
`ANSIBLE_BRANCH`=stable-2.11
tag: stable-2.11-devel
req: tools/requirements-stable-2.11.txt

AlanCoding (approved):
I think this looks good to me, I'll run it by @pabelanger before merge

thanks

pabelanger:
Looks good.

atbore-phx:
thx guys @AlanCoding  @pabelanger  for the review !

PR #754, opened 2021-07-15
Update run_command() examples in docs

Shrews:
The tuple returned from `run_command()` returns 3 items (response, error string and return code), not two, as used in the examples in the Python interface doc.


pabelanger:
recheck

shanemcd (approved):

PR #759, opened 2021-07-22
Remove usage of --userns=keep-id.

shanemcd:
When using an EE that has `USER root`, this was causing files created in the container to be owned by the mapped UID on the host. Not what we want!


shanemcd:
recheck

ganeshrn (approved):

PR #761, opened 2021-07-27
PEP8 and setuptools cleanup

samdoran:
All the PEP8 rules that were ignored have to do with whitespace formatting which help keep the code readable. Fix all the rule violations and remove the ignored rules.

### DIFF
--- a/ansible_runner/__init__.py
+++ b/ansible_runner/__init__.py
@@ -14,4 +14,3 @@ plugins = {
     for entry_point
     in pkg_resources.iter_entry_points('ansible_runner.plugins')
 }
-
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,11 +1,11 @@
 [metadata]
 name = ansible-runner
 author = Ansible, Inc.
-author_email = info@ansible.com
+author-email = info@ansible.com
 summary = "Consistent Ansible Python API and CLI with container and process isolation runtime capabilities"
-home_page = https://ansible-runner.readthedocs.io
-description_file = README.md
-description_content_type = text/markdown
+home-page = https://ansible-runner.readthedocs.io
+description-file = README.md
+description-content-type = text/markdown
 license_file = LICENSE.md
 
 [entry_points]
--- a/setup.cfg
+++ b/setup.cfg
@@ -19,8 +19,12 @@ data-files =
     share/ansible-runner/utils = utils/*
 
 [pep8]
+# W503 - Line break occurred before a binary operator
+ignore=W503
 exclude=.tox,venv
 
 [flake8]
+# W503 - Line break occurred before a binary operator
+ignore=W503
 max-line-length=160
 exclude=.tox,venv
--- a/setup.cfg
+++ b/setup.cfg
@@ -19,23 +19,8 @@ data-files =
     share/ansible-runner/utils = utils/*
 
 [pep8]
-# E201 - Whitespace after '('
-# E203 - Whitespace before ":"
-# E221 - Multiple spaces after operator
-# E225 - Missing whitespace around operator
-# E231 - Missing whitespace after ','
-# E241 - Multiple spaces after ','
-# E251 - Unexpected spaces around keyword / parameter equals
-# E261 - At least two spaces before inline comment
-# E302 - Expected 2 blank lines found 0
-# E303 - Too many blank lines
-# W291 - Trailing whitespace
-# W391 - Blank line at end of file
-# W293 - Blank line contains whitespace
-ignore=E201,E203,E221,E225,E231,E241,E251,E261,E265,E303,W291,W391,W293
 exclude=.tox,venv
 
 [flake8]
 max-line-length=160
-ignore=E201,E203,E221,E225,E231,E241,E251,E261,E265,E303,W291,W391,W293,E731,F405
 exclude=.tox,venv
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,11 +1,11 @@
 [metadata]
 name = ansible-runner
 author = Ansible, Inc.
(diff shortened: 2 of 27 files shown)


ansible_runner/__init__.py
```diff
@@ -14,4 +14,3 @@
     for entry_point
     in pkg_resources.iter_entry_points('ansible_runner.plugins')
```
webknjaz, line 15: As a side note, I strongly encourage migrating to importlib in place of `pkg_resources`

setup.cfg
```diff
@@ -1,11 +1,11 @@
 [metadata]
 name = ansible-runner
 author = Ansible, Inc.
-author-email = info@ansible.com
+author_email = info@ansible.com
```
webknjaz: I'm not sure if this is right. It's not a setuptools option, it's coming from `pbr`: https://docs.openstack.org/pbr/latest/user/using.html#setup-cfg.

If you want to use native setuptools settings, see https://setuptools.readthedocs.io/en/latest/userguide/declarative_config.html#metadata / https://packaging.python.org/guides/distributing-packages-using-setuptools/

FWIW I'd split this change into a separate PR

webknjaz (commented):
Also, while you're at it, I recommend you integrate 
https://github.com/pre-commit/pre-commit-hooks#trailing-whitespace
into `.pre-commit-config.yaml`. And use pre-commit in the linters tox env so that it's actually being executed. This small change would help keep the project free of trailing whitespaces.

eqrx (changes requested):
Thanks you very much for doign the linting first, I find it important :) I agree with @webknjaz that the underscore changes in setup.py should be split into another PR

samdoran:
I didn't get to this today, but I will tomorrow.

ansible_runner/__init__.py
samdoran: I will have to do that in a separate PR.

Shrews:
recheck

setup.cfg
```diff
@@ -19,23 +19,12 @@ data-files =
...
-# W391 - Blank line at end of file
-# W293 - Blank line contains whitespace
-ignore=E201,E203,E221,E225,E231,E241,E251,E261,E265,E303,W291,W391,W293
+# W503 - Line break occurred before a binary operator
+ignore=W503
 exclude=.tox,venv
 
 [flake8]
+# W503 - Line break occurred before a binary operator
+ignore=W503
```
webknjaz, line 28: This shouldn't be necessary because this rule is already listed in the default `ignore` setting — https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-ignore:
```suggestion
```

Moreover, it's best to avoid using the `ignore` setting because it replaces what flake8 has set there by default (`E121,E123,E126,E226,E24,E704,W503,W504`).

Instead, if you want to add new ignore entries w/o replacing that underlying ignore list, use `extend-ignore` — https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-extend-ignore

ansible_runner/__init__.py
webknjaz: Sure, that's the way :)

setup.cfg
```diff
@@ -19,23 +19,12 @@ data-files =
...
 max-line-length=160
-ignore=E201,E203,E221,E225,E231,E241,E251,E261,E265,E303,W291,W391,W293,E731,F405
 exclude=.tox,venv
```
webknjaz, line 30: While I'm on it, let me inform you that https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-extend-exclude exists. Of course, it's best to change it in a separate PR too.

eqrx (approved):

setup.cfg
Shrews: Personally, I do not like the default ignore list and would prefer we override `ignore` here, as this change is doing.

setup.cfg
```diff
@@ -19,23 +19,12 @@ data-files =
...
-# E251 - Unexpected spaces around keyword / parameter equals
-# E261 - At least two spaces before inline comment
-# E302 - Expected 2 blank lines found 0
-# E303 - Too many blank lines
-# W291 - Trailing whitespace
-# W391 - Blank line at end of file
-# W293 - Blank line contains whitespace
-ignore=E201,E203,E221,E225,E231,E241,E251,E261,E265,E303,W291,W391,W293
+# W503 - Line break occurred before a binary operator
+ignore=W503
```
Shrews, line 23: Do we even need this `[pep8]` section? We only run flake8 during the tests.

Shrews (commented):
Fine with the changes here. If we can determine we don't need the [pep8] section, I'd rather see that removed in this change.

shanemcd (approved):

setup.cfg
samdoran: Looks like that is only used my `autopep8`. If we have no plans to use that, we can remove it.

setup.cfg
samdoran: I didn't realize that `ignore` overwrote the default list. I thought the behavior of `extend-ignore` was the behavior of `ignore`. That explains why I've been confused by these two rules in the past.

samdoran: After looking at the rules in the default ignore list, I'm inclined to enable them all and only ignore W503 as we have here. I accidentally did what I like. :)

setup.cfg
samdoran: I agree it would probably be better to use `extend-exclude`. The default list of `.svn,CVS,.bzr,.hg,.git,__pycache__,.tox,.eggs,*.egg`

PR #555, opened 2020-11-03
Remove usage of psutil

wenottingham:
Just killpg when we're killing processes, and remove the cleanup of controlmasters in the callback.

Re: that cleanup, the cleanup code has:

```
    cp_path = os.environ.get('ANSIBLE_SSH_CONTROL_PATH', '')
    if not cp_path:
        return
```

AWX hasn't set ANSIBLE_SSH_CONTROL_PATH for playbook runs in 2+ years: https://github.com/ansible/awx/pull/1601

PR #762, opened 2021-07-29
Allow Users to Obtain Execution Node CPU and Memory Capacities

beeankha:
Connecting AWX Issue https://github.com/ansible/awx/issues/10693

* * *

The CPU/memory capacity checks implemented in this PR are based off of the `get_cpu_capacity` and `get_mem_capacity` functions found in `awx/main/utils/common.py`:
https://github.com/ansible/awx/blob/devel/awx/main/utils/common.py#L702-L749

Since [`psutil` was removed from Ansible Runner](https://github.com/ansible/ansible-runner/pull/555) last year, `multiprocessing` and `resource` were utilized in order to surface the relevant information. 

Now, the command `ansible-runner worker capacity` can be run, with the resulting output as shown below:

Screenshot from 2021-07-28 21-39-50

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -34,6 +34,7 @@ import tempfile
 import atexit
 
 from contextlib import contextmanager
+from pathlib import Path
 from uuid import uuid4
 
 from yaml import safe_load
@@ -41,7 +42,7 @@ from yaml import safe_load
 from ansible_runner import run
 from ansible_runner import output
 from ansible_runner.utils import dump_artifact, Bunch
-from ansible_runner.utils.capacity import get_cpu_capacity, get_mem_capacity
+from ansible_runner.utils.capacity import get_cpu_count, get_mem_info
 from ansible_runner.runner import Runner
 from ansible_runner.exceptions import AnsibleRunnerException
 
@@ -629,17 +630,18 @@ def main(sys_args=None):
         'worker',
         help="Execute work streamed from a controlling instance"
     )
-    info = worker_subparser.add_subparsers(
-        dest='worker_info',
-        help="Show the execution node's Ansible Runner version along with its memory and CPU capacities"
-    )
-    info.add_parser('capacity')
     worker_subparser.add_argument(
         "--private-data-dir",
         help="base directory containing the ansible-runner metadata "
              "(project, inventory, env, etc)",
     )
 
+    worker_subparser.add_argument(
+        "--worker-info",
+        dest="worker_info",
+        action="store_true",
+        help="show the execution node's Ansible Runner version along with its memory and CPU capacities"
+    )
     process_subparser = subparser.add_parser(
         'process',
         help="Receive the output of remote ansible-runner work and distribute the results"
@@ -763,12 +765,15 @@ def main(sys_args=None):
 
     if vargs.get('command') in ('worker', 'process'):
         if vargs.get('worker_info'):
-            cpu = get_cpu_capacity()
-            mem = get_mem_capacity()
+            cpu = get_cpu_count()
+            mem = get_mem_info()
             if cpu or mem > 0:
-                print("\nExecution Node Info\n"
-                      "-------------------\n"
-                      f"Ansible-Runner Version: {VERSION}\nCPU Capacity: {cpu}\nMemory Capacity: {mem}\n")
+                base = Path('worker_info')
+                info_file = str(uuid4().hex)
+                jsonpath = base / info_file
+                base.mkdir(exist_ok=True)
+                info = [VERSION, cpu, mem]
+                jsonpath.write_text(json.dumps(info))
                 parser.exit(0)
             else:
                 sys.tracebacklimit = 0
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -41,6 +41,7 @@ from yaml import safe_load
 from ansible_runner import run
 from ansible_runner import output
 from ansible_runner.utils import dump_artifact, Bunch
+from ansible_runner.utils.capacity import get_cpu_capacity, get_mem_capacity
 from ansible_runner.runner import Runner
 from ansible_runner.exceptions import AnsibleRunnerException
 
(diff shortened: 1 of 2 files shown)


ansible_runner/__main__.py
```diff
@@ -756,6 +762,17 @@ def main(sys_args=None):
     vargs = vars(args)
 
     if vargs.get('command') in ('worker', 'process'):
+        if vargs.get('worker_info'):
+            cpu = get_cpu_capacity()
+            mem = get_mem_capacity()
+            if cpu or mem > 0:
+                print("\nExecution Node Info\n"
```
shanemcd: How are you planning to parse this data? It might be easier to just make this print json.

ansible_runner/utils/capacity.py
```diff
@@ -0,0 +1,17 @@
+import multiprocessing
+import resource
+
+
+def get_cpu_capacity():
+    # `multiprocessing` info: https://docs.python.org/3/library/multiprocessing.html
+    forkcpu = 4
```
fosterseth: should we return raw cpu count here, and then on the control plane side calculate the capacity based on the user settings for `SYSTEM_TASK_FORKS_CPU`?

https://github.com/ansible/awx/blob/d89719c7409528006a65378c3e6ae1a4b2666ec0/awx/main/utils/common.py#L719

ansible_runner/utils/capacity.py
```diff
@@ -0,0 +1,17 @@
...
+    cpu_capacity = multiprocessing.cpu_count() * forkcpu
```
samdoran: This isn't necessarily the number of CPUs available to the worker. It may be more accurate to use `os.sched_getaffinity(0)` especially If the goal is reporting resources available to the worker. Depending on the hypervisor or container platform, it may display the host resources not the guest/container resources.

ansible_runner/utils/capacity.py
```diff
@@ -0,0 +1,17 @@
...
+    return cpu_capacity
+
+
+def get_mem_capacity():
+    # `resource` info: https://docs.python.org/3/library/resource.html
+    byte_denom = 1024
+    mem = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
```
samdoran: I think this is probably an accurate way to get the memory available. I'm doing some testing to confirm.

samdoran (commented):
It would be nice to have unit tests for the new functions.

ansible_runner/utils/capacity.py
samdoran: I confirmed that it's more accurate to use `os.sched_getaffinity(0)`.

```
> docker run --rm -it \
    --cpuset-cpus 0-1 \
    quay.io/samdoran/fedora34-ansible:latest \
    python3 -c 'import os, multiprocessing, resource; \
    print("Affinity: %s" % len(os.sched_getaffinity(0))); \
    print("Multiprocessing: %s" % multiprocessing.cpu_count())'

Affinity: 2
Multiprocessing: 40
```

samdoran: Granted, the accuracy depends on how the container was invoked. If the `--cpus` flag was used, then it reports the number of CPUs on the container host and not the units available to the container. Trying to get accurate information inside containers is always challenging.

ansible_runner/utils/capacity.py
samdoran: This one is more tricky. I haven't been able to find a way to detect limited memory from within the container. I imagine the most accurate way would be inspecting the `cgroup` from `/proc/1/cgroup`, but I didn't find anything definitive.

AlanCoding: For some background, this would replace the logic we have in AWX:

for CPU:

https://github.com/ansible/awx/blob/747231d3500d86ba07006a6927498d8ec31bf930/awx/main/utils/common.py#L716

and for memory:

https://github.com/ansible/awx/blob/747231d3500d86ba07006a6927498d8ec31bf930/awx/main/utils/common.py#L748

I've dug enough into the psutil implementation to say that it just gives the value from `/proc/meminfo` (for Linux). We can't use the old logic here, because we _just_ removed psutil as a dependency of ansible-runner, with much enthusiasm for it.

I only mean to say - if our implementation is _as bad_ as the old psutil one, it's not a regression. Doesn't mean we can't improve, but that's where we're coming from, because this is for AWX capacity-based job allocation.

ansible_runner/utils/capacity.py
```diff
@@ -0,0 +1,17 @@
+import multiprocessing
+import resource
+
+
+def get_cpu_capacity():
```
AlanCoding: We are talking about having this return "raw data", as opposed to the `capacity` number as it will appear in AWX `/api/v2/instances/`

Since that's the case, I would prefer to use a method name other than `get_cpu_capacity`. Call it `get_cpu_count` or something like that to make it more clear. That will also help to allow us to import this freely in the AWX code.

beeankha:
For now, a `worker_info` directory gets created wherever the `ansible-runner worker --worker-info` command is run from, and the files that contain the `version`, `cpu` and `mem` capacity info are generated like so:

Screenshot from 2021-08-03 09-14-26

I can hardcode a filename instead if we want to only have one file in the `worker_info` directory at a time, to possibly make it easier to pick up the most current worker info text (or not generate a directory/file at all).

ansible_runner/__main__.py
```diff
@@ -756,6 +764,20 @@ def main(sys_args=None):
     vargs = vars(args)
 
     if vargs.get('command') in ('worker', 'process'):
+        if vargs.get('worker_info'):
+            cpu = get_cpu_count()
+            mem = get_mem_info()
+            if cpu or mem > 0:
+                base = Path('worker_info')
+                info_file = str(uuid4().hex)
+                jsonpath = base / info_file
```
shanemcd: I had no idea you could do this. 👍

shanemcd (approved):

beeankha:
Latest commit (https://github.com/ansible/ansible-runner/pull/762/commits/cdc3372359a96a01475d3d9865e08374f78ee965) makes it so that the data doesn't output to a file anymore, it instead prints to the terminal and is in YAML format:
```
$ ansible-runner worker --worker-info
{CPU Capacity: 12, Errors: [], Memory Capacity: 32615876, Version: 2.0.0.0a4.dev61}
```

If the `meminfo` file can't be found, an error will surface and get appended to a list that is under the key `error`:
```
$ ansible-runner worker --worker-info
{CPU Capacity: 12, Errors: ['The /proc/meminfo file could not found, memory capacity
      undiscoverable.'], Memory Capacity: null, Version: 2.0.0.0a4.dev61}
```

The changes in this commit also enable the _total_ memory capacity to be discovered, vs the _available_ memory capacity.

shanemcd (approved):

jladdjr (approved):

PR #763, opened 2021-07-30
Support custom stdout callbacks

shanemcd:


shanemcd:
Before:

image (4)

After:

image (5)

shanemcd:
Even works in AWX! 
image (6)

ansible_runner/display_callback/minimal.py
```diff
@@ -1,29 +0,0 @@
...
-# Python
-import os
-
-# Ansible
-import ansible
-
-# Because of the way Ansible loads plugins, it's not possible to import
-# ansible.plugins.callback.minimal when being loaded as the minimal plugin. Ugh.
-minimal_plugin = os.path.join(os.path.dirname(ansible.__file__), 'plugins', 'callback', 'minimal.py')
-exec(compile(open(minimal_plugin, "rb").read(), minimal_plugin, 'exec'))
```
shanemcd, line 29: 🔥

matburt (approved):
🔥

ansible_runner/display_callback/minimal.py
samdoran: Yeah, that some terribly scary code. Leaking a file description to boot.

ansible_runner/display_callback/module.py
```diff
@@ -21,18 +21,35 @@
...
+
+is_adhoc = os.getenv('AD_HOC_COMMAND_ID', False)
+
+# Dynamically construct base classes for our callback module, to support custom stdout callbacks.
+if is_adhoc:
+    default_stdout_callback = 'minimal'
+else:
+    default_stdout_callback = C.config.get_config_value('DEFAULT_STDOUT_CALLBACK')
+
+DefaultCallbackModule = callback_loader.get(default_stdout_callback).__class__
```
AlanCoding, line 45: Huh, so compared to #387, it looks like the key difference is that you don't try to change the "awx_display" name. You swap out the super class instead. This may resolve some of my hesitations about this.

In unpredictable contexts, I got errors importing and using the Ansible plugin loaders inside of other plugins. Since this runs on the control machine, it might be less likely to be a problem.

ansible_runner/config/runner.py
```diff
@@ -243,6 +243,13 @@ def prepare_env(self):
         self.env["RUNNER_OMIT_EVENTS"] = str(self.omit_event_data)
         self.env["RUNNER_ONLY_FAILED_EVENTS"] = str(self.only_failed_event_data)
 
+        original_stdout_callback = os.getenv('ANSIBLE_STDOUT_CALLBACK')
+        if original_stdout_callback:
+            self.env['ORIGINAL_STDOUT_CALLBACK'] = original_stdout_callback
+
+        self.env['ANSIBLE_STDOUT_CALLBACK'] = 'awx_display'
+        self.env["ANSIBLE_LOAD_CALLBACK_PLUGINS"] = '1'
```
AlanCoding: I don't like that we ever modified `self.env` in the runner module to begin with. It's an oddball. Most of the environment is prepared in the config class, and this is an exception that I don't feel like we ever needed.

ansible_runner/display_callback/module.py
```diff
@@ -21,18 +21,35 @@
...
+if user_specified_callback:
+    os.environ['ANSIBLE_STDOUT_CALLBACK'] = user_specified_callback
+
+is_adhoc = os.getenv('AD_HOC_COMMAND_ID', False)
+
+# Dynamically construct base classes for our callback module, to support custom stdout callbacks.
+if is_adhoc:
+    default_stdout_callback = 'minimal'
+else:
+    default_stdout_callback = C.config.get_config_value('DEFAULT_STDOUT_CALLBACK')
```
AlanCoding: You're going through this process that modifies `os.environ`, which can have unforeseen side-effects for the user later, and I just don't see any need for it.

```python
>>> os.environ['ANSIBLE_STDOUT_CALLBACK'] = 'foo'
>>> C.config.get_config_value('DEFAULT_STDOUT_CALLBACK')
'foo'
```

So you have this value that the user set, you read it, you set the callback to it in `os.environ`, and then you read it back via `C.config.get_config_value`.

Instead, couldn't you just do this?

```python
if is_adhoc:
    default_stdout_callback = 'minimal'
else:
    default_stdout_callback = user_specified_callback
```

And drop the part about `os.environ['ANSIBLE_STDOUT_CALLBACK'] = user_specified_callback`

ansible_runner/display_callback/module.py
```diff
@@ -21,18 +21,35 @@
...
 from ansible.plugins.callback import CallbackBase
-from ansible.plugins.callback.default import CallbackModule as DefaultCallbackModule
+from ansible.plugins.loader import callback_loader
 
-# AWX Display Callback
 from .events import event_context
-from .minimal import CallbackModule as MinimalCallbackModule
+
+user_specified_callback = os.getenv('ORIGINAL_STDOUT_CALLBACK')
+runner_callback = os.environ.pop('ANSIBLE_STDOUT_CALLBACK', 'awx_display')
```
AlanCoding: I don't think there's any case where this would be something other than `'awx_display'`? If that's true, then I think it would be easier to follow to hard-code that.

AlanCoding:
I think we should be able to write some integration tests, right?

ansible_runner/config/runner.py
```diff
@@ -243,6 +243,13 @@ def prepare_env(self):
         self.env["RUNNER_OMIT_EVENTS"] = str(self.omit_event_data)
         self.env["RUNNER_ONLY_FAILED_EVENTS"] = str(self.only_failed_event_data)
 
+        original_stdout_callback = os.getenv('ANSIBLE_STDOUT_CALLBACK')
```
Shrews: I'm still trying to grasp everything this PR is doing, but shouldn't we check `self.env` here instead of using `os.getenv()`?

Looking at `BaseConfig._prepare_env()`, it populates `self.env` with the environment variables, but also populates with any custom env vars (e.g. from `env/envvars`). So we'd be missing the latter with the current line of code.

AlanCoding: I agree with you. In another comment, I argued that `ANSIBLE_STDOUT_CALLBACK` should still be set in the RunnerConfig class instead of the runner.py module. The typical practice there is to compare with `self.env` as opposed to `os.environ`.

> Looking at `BaseConfig._prepare_env()`, it populates `self.env` with the environment variables

That's only true if the run is not containerized. For reasons of repeatability and functionality, the environment for runs in container start from scratch. If the run is _not_ containerized, then `self.env` should already be updated with `os.environ`. Thus, we only need consider `self.env`.

AlanCoding:
Confirmed this works with `ANSIBLE_STDOUT_CALLBACK=yaml ansible-runner run demo/ -p test.yml` (the difference is subtle). I would intend to make it work to add in `demo/env/envvars` as well.

AlanCoding:
Here's the start of some work for tests:

```diff
diff --git a/test/integration/test_config.py b/test/integration/test_config.py
index b79eab1..a22847d 100644
--- a/test/integration/test_config.py
+++ b/test/integration/test_config.py
@@ -1,7 +1,28 @@
 from ansible_runner.config._base import BaseConfig
+from ansible_runner.interface import run
+
+import os
 
 
 def test_combine_python_and_file_settings(project_fixtures):
     rc = BaseConfig(private_data_dir=str(project_fixtures / 'job_env'), settings={'job_timeout': 40})
     rc._prepare_env()
     assert rc.settings == {'job_timeout': 40, 'process_isolation': True}
+
+
+def test_default_ansible_callback(project_fixtures):
+    """This is the reference case for stdout customization tests, assures default stdout callback is used"""
+    res = run(private_data_dir=str(project_fixtures / 'debug'), playbook='debug.yml')
+    stdout = res.stdout.read()
+    assert res.rc == 0, stdout
+
+    assert '"msg": "Hello world!"' in stdout, stdout
+
+
+def test_custom_stdout_callback(project_fixtures, mocker):
+    mocker.patch.dict(os.environ, {'ANSIBLE_STDOUT_CALLBACK': 'yaml'})
+    res = run(private_data_dir=str(project_fixtures / 'debug'), playbook='debug.yml')
+    stdout = res.stdout.read()
+    assert res.rc == 0, stdout
+
+    assert 'msg: Hello world!' in stdout, stdout
```

This patches `os.environ`, with the more obvious approach being to pass `envvars` to `run(`, which is the next case to build here. Currently these pass, echoing what is established to work.

AlanCoding:
I put up https://github.com/ansible/ansible-runner/pull/953 to test my suggested changes. I might still push 1 more commit, a nice-to-have sort of thing. Otherwise everything seems to be working.

AlanCoding:
Hopefully I've now addressed all the issues here. There were some hiccups with ad-hoc commands, but I _think_ I'm done with that.

Notes for the future about what we can do better:

 - Use something other than `AD_HOC_COMMAND_ID` to denote that it's an ad-hoc command, probably a new env var
 - Where `ANSIBLE_STDOUT_CALLBACK` is set in `BaseConfig`, move that logic to `RunnerConfig`
   - reasoning is that it only applies to commands covered by this config class
   - criteria then makes use of `ExecutionMode.ANSIBLE`... however, this requires moving around other code

I considered a lot of other potential changes, but abandoned them. I think these are solid suggestions, but go pretty far out of scope for this change set.

AlanCoding:
Remaining failures, like `test_module_run[docker]`, look like they are _probably_ failing because it's using the current devel image, as opposed to an image built from this branch. That shouldn't be the case, and I don't know what went wrong.

AlanCoding:
If there aren't any objections, I'd like to submit https://github.com/ansible/ansible-runner/pull/957 as a replacement for this PR. It has the improvements here, but I kept going with more.

PR #765, opened 2021-08-03
Work around kernel bug that forces fuse-overlayfs in podman

samdoran:
In the current version of RHEL there is a bug in the fuser-overlayfs driver that prevents extended attributes from being set. This PR uses the mechanism [provided](https://github.com/ansible/ansible/pull/73282) in `anisble-core` to work around this issue.

### DIFF
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -196,8 +196,8 @@ class BaseConfig(object):
 
             if self.process_isolation_executable == 'podman':
                 # A kernel bug in RHEL < 8.5 causes podman to use the fuse-overlayfs driver. This results in errors when
-                # trying to set extended file attributes. Setting this environment variables
-                # allows modules to take advantage of a fallback to work around this bug when failures are encountered.
+                # trying to set extended file attributes. Setting this environment variable allows modules to take advantage
+                # of a fallback to work around this bug when failures are encountered.
                 #
                 # See the following for more information:
                 #    https://github.com/ansible/ansible/pull/73282
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -194,15 +194,16 @@ class BaseConfig(object):
             # Special flags to convey info to entrypoint or process in container
             self.env['LAUNCHED_BY_RUNNER'] = '1'
 
-            # A kernel bug in RHEL < 8.5 causes podman to use the fuse-overlayfs driver. This results in errors when
-            # trying to set extended file attributes. Setting this environment variables
-            # allows modules to take advantage of a fallback to work around this bug when failures are encountered.
-            #
-            # See the following for more information:
-            #    https://github.com/ansible/ansible/pull/73282
-            #    https://github.com/ansible/ansible/issues/73310
-            #    https://issues.redhat.com/browse/AAP-476
-            self.env['ANSIBLE_UNSAFE_WRITES'] = '1'
+            if self.process_isolation_executable == 'podman':
+                # A kernel bug in RHEL < 8.5 causes podman to use the fuse-overlayfs driver. This results in errors when
+                # trying to set extended file attributes. Setting this environment variables
+                # allows modules to take advantage of a fallback to work around this bug when failures are encountered.
+                #
+                # See the following for more information:
+                #    https://github.com/ansible/ansible/pull/73282
+                #    https://github.com/ansible/ansible/issues/73310
+                #    https://issues.redhat.com/browse/AAP-476
+                self.env['ANSIBLE_UNSAFE_WRITES'] = '1'
 
             artifact_dir = os.path.join("/runner/artifacts", "{}".format(self.ident))
             self.env['AWX_ISOLATED_DATA_DIR'] = artifact_dir
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -194,7 +194,7 @@ class BaseConfig(object):
             # Special flags to convey info to entrypoint or process in container
             self.env['LAUNCHED_BY_RUNNER'] = '1'
 
-            # A bug in the fuse-overlayfs driver in RHEL < 5.8 causes errors when trying to set extended file attributes.
+            # A bug in the fuse-overlayfs driver in RHEL < 8.5 causes errors when trying to set extended file attributes.
             # Setting this environment variablesallows modules to take advantage of a fallback to work around
             # this bug only when failures are encountered.
             #
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -193,6 +193,17 @@ class BaseConfig(object):
             self.env = {}
             # Special flags to convey info to entrypoint or process in container
             self.env['LAUNCHED_BY_RUNNER'] = '1'
+
+            # A bug in the fuse-overlayfs driver in RHEL < 5.8 causes errors when trying to set extended file attributes.
+            # Setting this environment variablesallows modules to take advantage of a fallback to work around
+            # this bug only when failures are encountered.
+            #
+            # See the following for more information:
+            #    https://github.com/ansible/ansible/pull/73282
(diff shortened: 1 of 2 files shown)


ansible_runner/config/_base.py
```diff
@@ -193,6 +193,17 @@ def _prepare_env(self, runner_mode='pexpect'):
             self.env = {}
             # Special flags to convey info to entrypoint or process in container
             self.env['LAUNCHED_BY_RUNNER'] = '1'
+
+            # A bug in the fuse-overlayfs driver in RHEL < 5.8 causes errors when trying to set extended file attributes.
```
matburt: 5.8?!

matburt: Is there a way to make this conditional on system capabilities? Rhel version? Rather than make it a blanket change on any containerized execution?

samdoran: Doh! Too many numbers. It's the 5.x kernel and the RHEL 8.x version. 🤦‍♂️

samdoran: That was my first inclination too: probe the running image to determine if the setting is needed.

samdoran: I'm not sure if we would be able to get details on the filesystem driver or container host from within the running container, though. I'm open to suggestions. My endeavors with fact collection inside containers results in mountains of conditional hacks and workarounds.

matburt: This code in particular, though, exists outside of the container as it's in the environment preparation.

samdoran: Oh, ok. There is hope, then.

shanemcd:
I think we should only set this when running under Podman. This shouldn’t be an issue for Docker.

ansible_runner/config/_base.py
sivel: This may be complicated to discover, as there are several layers where users could have configured their default storage driver.  Setting it regardless for podman shouldn't be too problematic, it just allows 1 additional fallback to perform an unsafe write when `atomic_move` fails.  Currently `atomic_move` has a waterfall of 4 methods, and if one succeeds, even with unsafe writes enabled, it'll never make it to the unsafe write.

For ref, RHEL8.5 is supposed to include the backported kernel fix that is going 5.13 of the kernel.  RHEL8 is using a 4.18 kernel, so that detail is probably not worth mentioning unless we find the exact kernel version they will include the backport in.

https://www.redhat.com/sysadmin/podman-rootless-overlay

But as @shanemcd mentions in his comment, we should probably just make it conditional to running under podman.

ansible_runner/config/_base.py
```diff
@@ -193,6 +193,17 @@ def _prepare_env(self, runner_mode='pexpect'):
             self.env = {}
             # Special flags to convey info to entrypoint or process in container
             self.env['LAUNCHED_BY_RUNNER'] = '1'
+
+            # A bug in the fuse-overlayfs driver in RHEL < 8.5 causes errors when trying to set extended file attributes.
```
sivel: I don't want to be too pedantic here, but it's that there is a kernel bug, that results in RHEL8 defaulting to `fuse-overlayfs` and then that `fuse-overlayfs` doesn't support xattrs, which breaks `atomic_move`.

Most details are available at https://www.redhat.com/sysadmin/podman-rootless-overlay

ansible_runner/config/_base.py
```diff
@@ -193,6 +193,17 @@ def _prepare_env(self, runner_mode='pexpect'):
...
+            # Setting this environment variablesallows modules to take advantage of a fallback to work around
```
sivel:
```suggestion
            # Setting this environment variable allows modules to take advantage of a fallback to work around
```

ansible_runner/config/_base.py
samdoran: No worries. I was trying to explain it succinctly and was not 100% sure I had it right. Thank you for clarifying.

ansible_runner/config/_base.py
samdoran: I will work on setting this only for `podman`.

ansible_runner/config/_base.py
```diff
@@ -193,6 +193,18 @@ def _prepare_env(self, runner_mode='pexpect'):
             self.env = {}
             # Special flags to convey info to entrypoint or process in container
             self.env['LAUNCHED_BY_RUNNER'] = '1'
+
+            if self.process_isolation_executable == 'podman':
+                # A kernel bug in RHEL < 8.5 causes podman to use the fuse-overlayfs driver. This results in errors when
```
rlopez133, line 198: Since this is a RHEL < 8.5 issue, should there be a conditional check when to enable ANSIBLE_UNSAFE_WRITES? I'm assuming if a user is using RHEL8.5+ you'd want to continue using the original behavior. 

And to add to that, would there be a plan to look at other distros regarding that conditional check?

sivel: That assumption implies that we would know whether people who have upgraded have changed their storage driver.  Which may also require that they have reset all prior podman content.  Also, since 8.5 hasn't been released and the blog post says:

> For the RHEL/CentOS Stream, we plan on backporting the feature for the RHEL8.5 release in the fall.

It would probably be wise to ensure the backport does indeed make it into 8.5 first.

shanemcd: What do you all think about checking the values of NAME and VERSION_ID in `/etc/os-release`? It would be nice have this auto-disable when 8.5 is out.

samdoran: This is really a kernel bug that will be fixed in RHEL 8.5. From [this article](https://www.redhat.com/sysadmin/podman-rootless-overlay) it sounds like Linux >= 5.11 _without_ SELinux or >=5.13 _with_ SELinux will not have this problem. 

> The kernel gained rootless support in the 5.11 kernel, but a bug prevented SELinux use with the file system; this bug was fixed in 5.13.

> As far as other distributions are concerned, this support will show up when kernel 5.13 is released. For the RHEL/CentOS Stream, we plan on backporting the feature for the RHEL8.5 release in the fall.

sivel: I think ideally, the check should check for functionality instead of a version.  `/etc/containers/storage.conf` sets the driver to `fuse-overlayfs`.  So I believe someone would have to change that still after the upgrade.  Additionally, users could configure `fuse-overlayfs` as the default even afterwards.  It might require running `podman info` and inspecting `store.graphDriverName` or `store.graphOptions.overlay.mount_program.Executable`

shanemcd: We currently only support EEs that are based on UBI, so I think we can make _some_ assumptions about the environment for this scenario. 

Since this only affects local copy operations inside the EE, and will not affect remote host.. I dont care all that much. If we want to move forward with things as-is, we can remove it when our EEs are based on UBI 8.5.

samdoran: Identifying the distribution got a lot more complicated since `platform.linux_distribution()` was removed in Python 3.7. The `distro` library is a good alternative, which Ansible Core vendors. But there isn't anything equivalent available to Ansible Runner, so accurately identifying the distribution and version information, then making decisions based on that would bring with it a lot of added complexity for very little gain.

samdoran: Feature probing would be far more accurate than relying on distribution information. But I still don't think it's worth the extra effort, and even feature probing may prove unreliable. I prefer a simpler and more reliable solution.

wenottingham: > We currently only support EEs that are based on UBI, so I think we can make some assumptions about the environment for this scenario.

Isn't the issue the *host* OS, not the RHEL 8 version in the EE?

shanemcd: Yeah, good point, not sure what I was thinking there. 

I think we can just set this for all podman invocations until we're only supporting 8.5+ for control plane and execution nodes. So this PR LGTM.

shanemcd (approved):

PR #769, opened 2021-08-05
[release_2.0] Work around kernel bug that forces fuse-overlayfs in podman (#765)

samdoran:
Backport of #765 for Ansible Runner 2.0.

### DIFF
--- a/test/unit/config/test__base.py
+++ b/test/unit/config/test__base.py
@@ -314,3 +314,28 @@ def test_containerization_settings(tmpdir, container_runtime):
             assert '--user=' in rc.command[index]
         else:
             assert rc.command[index] == element
+
+
+@pytest.mark.parametrize(
+    'container_runtime, expected', (
+        ('docker', None),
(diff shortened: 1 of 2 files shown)


samdoran:
recheck

Shrews:
recheck

test/unit/config/test__base.py
```diff
@@ -314,3 +314,28 @@ def test_containerization_settings(tmpdir, container_runtime):
             assert '--user=' in rc.command[index]
         else:
             assert rc.command[index] == element
+
+
+@pytest.mark.parametrize(
+    'container_runtime, expected', (
+        ('docker', None),
+        ('podman', '1')
```
Shrews, line 322: We really need to just fix it so that these tests are actually run rather than doing this awkward mock.

samdoran: Or refactor the code so we can pass in a parameter for testing use rather than having to do all this setup.

PR #772, opened 2021-08-05
Update linter configuration

samdoran:
- remove section for `autopep8` since we don't use it
- Change `flake8` settings to exclude excludes since the default list of excluded files is good

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,4 +22,4 @@ data-files =
 # W503 - Line break occurred before a binary operator
 ignore=W503
 max-line-length=160
-exclude=.tox,venv
+extend-exclude=venv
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -18,13 +18,8 @@ packages =
...
-# W503 - Line break occurred before a binary operator
-ignore=W503
-exclude=.tox,venv
-
 [flake8]
 # W503 - Line break occurred before a binary operator
 ignore=W503
 max-line-length=160
-exclude=.tox,venv
+extend-exclude=venv
```
Shrews: Why do we need `venv` here? That isn't part of the repo so seems like we can just remove this line altogether.

Shrews (dismissed):

setup.cfg
samdoran: No idea. Maybe we can just remove this altogether.

shanemcd (approved):

PR #773, opened 2021-08-05
Remove unused tox test environments

Shrews:
The `testenv:py*` entries are not necessary (especially since we no longer run the `py36` and `py37` jobs). We just need the `py38` factor conditional dependency for the Zuul tests.

To make testing locally easier and not require a specific python version, a developer may use `py` or `py3` as the tox target. E.g., `tox -e py3` will run the tests with whichever Python 3 version you may have installed.

Also, we now use `{posargs}` so we can specify a specific test to run. Example:

`tox -e py3 -- test/integration/test_interface.py::test_env_accuracy`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,8 +6,7 @@ deps = ansible27: ansible<2.8
        ansible28: ansible<2.9
        ansible29: ansible<2.10
        ansible-base: ansible-base
-       py36: ansible-core
-       py37: ansible-core
+       py3: ansible-core
        py38: ansible-core
        -rrequirements.txt
        -rtest/requirements.txt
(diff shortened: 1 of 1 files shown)


pabelanger (approved):
LGTM

tox.ini
```diff
@@ -6,8 +6,7 @@ deps = ansible27: ansible<2.8
        ansible28: ansible<2.9
        ansible29: ansible<2.10
        ansible-base: ansible-base
-       py36: ansible-core
-       py37: ansible-core
+       py3: ansible-core
        py38: ansible-core
```
webknjaz: Would this work?
```suggestion
       py{,3,38}: ansible-core
```

It's a common pattern to run `tox -e python` or `tox -e py` which corresponds to just the `py` factor — this uses the current Python interpreter.

Shrews: I've not encountered those patterns, tbh, and I honestly don't know if that expression works, but `py3` associates the need for a python 3.x version, where as the other two (`python` and `py`) do not. I think we should stick with `py3` here.

webknjaz: I think that should be an alias. A popular way of setting up the CI is usually choosing the Python interpreter on the job matrix level and then running `tox -e py` or even having `TOXENV=python tox`. w/o needing to calculate the exact python factor to put as an argument for `-e`.

shanemcd (approved):

PR #774, opened 2021-08-06
Add python37-devel for tox testing

Shrews:
The `ansible-tox-py37` job was changed from using a `centos` node to using a `fedora-latest` node. In AWS, some of these nodes will not have Python 3.7, and some will, causing random test failures. This makes sure that it is installed before running the tests.

And because the `py36` and `py37` jobs were removed from our project def in `project-config`, we re-add them here in-repo.

### DIFF
--- a/.zuul.d/project.yaml
+++ b/.zuul.d/project.yaml
@@ -6,6 +6,8 @@
         - ansible-runner-build-container-image-stable-2.9
         - ansible-runner-build-container-image-stable-2.10
         - ansible-runner-build-container-image-stable-2.11
+        - ansible-tox-py36
+        - ansible-tox-py37
     gate:
       jobs: *id001
     post:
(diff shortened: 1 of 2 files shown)


pabelanger:
recheck

.zuul.d/project.yaml
```diff
@@ -6,6 +6,8 @@
         - ansible-runner-build-container-image-stable-2.9
         - ansible-runner-build-container-image-stable-2.10
         - ansible-runner-build-container-image-stable-2.11
+        - ansible-tox-py36
+        - ansible-tox-py37
     gate:
       jobs: *id001
     post:
```
pabelanger: So we should likely delete post / periodic jobs for now. Given we need to think about it a little.

I don't think we need to provide images build of runner 2.0, but just use the devel version of all testing. These are the images that end up on quay for others to use

Shrews: Removed those two sections.

Shrews:
Abandoning in favor of https://github.com/ansible/ansible-runner/pull/776

PR #781, opened 2021-08-10
Fix integration test for proper run_command() return

Shrews:
`test_run_ansible_command_within_container` test was failing because of bad order of return tuple values.


Shrews:
recheck

shanemcd (approved):

PR #782, opened 2021-08-10
DNM: Install docker for integration tests

Shrews:
DNM.... testing

Depends-On: https://github.com/ansible/ansible-runner/pull/781


Shrews:
recheck

Shrews:
This won't work as-is b/c ubuntu does not start the docker service on installation. This will require writing a new custom job with a run playbook that uses the `ensure-docker` role to install and start docker... and then cross fingers that works.

PR #784, opened 2021-08-11
Check-in AUTHORS and ChangeLog files generated by PBR

shanemcd:
I'll admit the changelog aint very pretty, but it's better than not being accurate!


Shrews:
Do we really want to commit these? Doing so would require them changing on every PR, wouldn't it? I was considering adding them to `.gitignore` soon.

shanemcd:
Closing this per discussion in Slack

PR #790, opened 2021-08-12
Add custom integration python job

Shrews:
Changes the tox py38 job to also parent to `ansible-buildset-registry-consumer` which will get us a container environment (either docker or podman) on the test node.  This will cause several more tests, which are skipped if neither docker nor podman are installed, to actually run.

Removes the following (now failing) tests that were left in when `adhoc` and `playbook` commands were removed:

- test_adhoc_localhost_setup
- test_playbook_with_private_data_dir
- test_playbook_with_relative_path

The following tests were being skipped:

- test/integration/test_events.py::test_async_events
- test/integration/test_events.py::test_profile_data
- test/integration/test_display_callback.py::test_callback_plugin_records_notify_events[playbook0]
- test/integration/test_interface.py::test_run_script_within_container
- test/integration/test_interface.py::test_get_plugin_docs_within_container
- test/integration/test_events.py::test_basic_events[True]
- test/integration/test_interface.py::test_run_ansible_command_within_container
- test/integration/test_interface.py::test_get_inventory_within_container
- test/integration/test_interface.py::test_get_plugin_docs_list_within_container
- test/integration/test_transmit_worker_process.py::TestStreamingUsage::test_remote_job_by_sockets[adhoc]
- test/integration/test_transmit_worker_process.py::TestStreamingUsage::test_remote_job_by_sockets[run]
- test/integration/test_interface.py::test_env_accuracy_inside_container
- test/integration/containerized/test_cli_containerized.py::test_module_run
- test/integration/containerized/test_cli_containerized.py::test_playbook_run
- test/integration/containerized/test_cli_containerized.py::test_provide_env_var
- test/integration/containerized/test_cli_containerized.py::test_adhoc_localhost_setup
- test/integration/containerized/test_cli_containerized.py::test_playbook_with_private_data_dir
- test/integration/containerized/test_cli_containerized.py::test_playbook_with_relative_path
- test/integration/containerized/test_container_management.py::test_cancel_will_remove_container

After this change, this is the remaining set of skipped tests:

- test/integration/test_events.py::test_async_events[True]
- test/integration/test_events.py::test_profile_data
- test/integration/test_display_callback.py::test_callback_plugin_records_notify_events[playbook0]
- test/integration/test_events.py::test_basic_events[True]
- test/integration/containerized/test_cli_containerized.py::test_module_run (requires podman)
- test/integration/containerized/test_cli_containerized.py::test_playbook_run (requires podman)
- test/integration/containerized/test_cli_containerized.py::test_provide_env_var (requires podman)


pabelanger:
recheck

.zuul.d/jobs.yaml
```diff
@@ -193,3 +193,8 @@
 - job:
     name: ansible-runner-upload-container-image-stable-2.9
     parent: ansible-runner-container-image-base
+
+- job:
```
pabelanger, line 198: https://github.com/ansible/ansible-zuul-jobs/blob/master/zuul.d/jobs.yaml#L441 is an example of how we do integration testing with ansible-runner.  You could do the same here.

Then adding a dependency on the ansible-runner-build-container-image like:

https://github.com/ansible/ansible-zuul-jobs/blob/master/zuul.d/network-jobs.yaml#L9

should get you the container on to the node to run tests against.

Shrews: @pabelanger First link seems out-of-date now. Any chance you could re-point me to the proper job?

pabelanger: If you parent to ansible-buildset-registry-consumer (like we did for ansible-ee-tests-base) you'll get for free the setting up of a node to use containers. We'll also pull down the proper image from the container-build jobs too

https://github.com/ansible/ansible-zuul-jobs/blob/master/zuul.d/jobs.yaml#L351

https://zuul-ci.org/docs/zuul-jobs/docker-image.html gives some information on the flow

Shrews: I see what you're driving at, but the issue here is that we really need two parents: one for the ansible-tox-py38, and another for image re-use. Zuul doesn't allow for that.

Going to have to put this work on hold for a bit since I think these integration tests that use a container for testing is fundamentally flawed anyway. They are pulling the `ansible-runner:devel` image from quay, rather than building a new image based on the current code.

pabelanger: actually, zuul does support 2 parents. You can do something like: https://github.com/ansible/network-ee/blob/main/.zuul.d/jobs.yaml#L45 and https://github.com/ansible/network-ee/blob/main/.zuul.d/jobs.yaml#L80.

We can do it because of how job variants work in zuul.

Shrews:
`test/integration/containerized/test_container_management.py::test_cancel_will_remove_container` is the lone test failure here.

Seems `podman ps -aq --filter name=ansible_runner_foo_bar` is failing on the test node, returning a rc=125. Likely because the container does not exist, which is what the test is expecting. This is not reproducible on a newer OS (I get rc=0), so I'm guessing an older version of podman is in use here and it's been changed so that no longer gives an error on unknown containers?

@pabelanger I could not find the version of podman in use in the test logs. Any idea how I could figure that out?

Shrews:
recheck

Shrews:
recheck

test/integration/test_main.py
```diff
@@ -285,7 +295,7 @@ def test_role_start():
         p.join()
 
 
-def test_playbook_start(skipif_pre_ansible28):
+def ttest_playbook_start(skipif_pre_ansible28):
```
shanemcd: ttypo?

shanemcd (approved):

Shrews:
recheck

test/integration/test_main.py
Shrews: yup, fixed

PR #796, opened 2021-08-14
Migrate container jobs to centos-8-stream nodeset

pabelanger:
This should result in faster builds, and closer testing to RHEL.

Depends-On: https://github.com/ansible/ansible-zuul-jobs/pull/1050
Depends-On: https://github.com/ansible/ansible-runner/pull/797
Signed-off-by: Paul Belanger <pabelanger@redhat.com>

PR #798, opened 2021-08-15
Rework bindep.txt files

pabelanger:
When we run CI jobs, we are actually install packages for our
containers.  Given both use bindep.txt files, we need to be a little
more dynamic.

This pattern is already used for requirements.txt files and should
result in some faster CI run times.

Depends-On: https://github.com/ansible/ansible-runner/pull/796


pabelanger:
recheck

pabelanger:
recheck

shanemcd (approved):

Shrews (approved):

PR #805, opened 2021-08-27
Additionally mount `~/.ssh/` to `/root/.ssh` inside EEs

shanemcd:
Currently when running inside an EE as root, only `SSH_AUTH_SOCK` is working. 

This enables the usage of default key names under `~/.ssh/`on the host.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,6 @@
+# For more info on CODEOWNERS file, see:
+#   - https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners
+#   - https://github.blog/2017-07-06-introducing-code-owners/
+
+# Default owners for everything in the repo.
+*    @ansible/executor


cidrblock:
Some background:

paramiko looks for keys in $HOME/.ssh, $HOME in the EE is set to /home/runner
openssh pulls home from /etc/passwd, which is /root

this change will allow for both the use of paramiko and openssh within the EE, and keys being located in each of those default search paths

(this all w/o ssh-agent)

Shrews:
LGTM, except for needing test updates.

samdoran (approved):
Change looks good. The tests are failing due to a mismatch in expected order of options:
```
test/unit/config/test_ansible_cfg.py:92: AssertionError

AssertionError: assert '-v' == '--group-add=root'
```

Do we need to add new tests for this to ensure it doesn't break in the future?

samdoran:
recheck

samdoran:
Do we not need this anymore? It looks like the tests will need to be updated since a new mount is being added. I plan to refactor the command and object generation into common functions that can be shared. That will minimize the locations we need to change the tests when the code changes.

Issue #809, opened 2021-08-31
[Question] How to disable env/passwords to be created


FloLaco:
Hi,

I run `ansible-runner` by providing password with a dict. It run with success. If I re-run the script. It's not working anymore.
The issue is, when I run the first time the script, `ansible-runner` create a the file `env/passwords` with the content of what I provided with my dict.
If I remove manually this folder, my script is working, but again, it re create the file ...

How to disable the creation of this sensitive file ?

```
import ansible_runner
passwords = {"SSH password:": "MYPASSWORD"}
r = ansible_runner.run(private_data_dir='/home/florian/ansible-runner-test', playbook='get_facts.yaml', inventory="myhosts", passwords=passwords)
print("{}: {}".format(r.status, r.rc))
# successful: 0
for each_host_event in r.events:
    print(each_host_event['event'])
print("Final status:")
print(r.stats)
```

Here's the error on my second run : 
```
(ansible-runner) florian@dsrauto-dev1:~/ansible-runner-test$ python main.py
Traceback (most recent call last):
  File "main.py", line 6, in <module>
    r = ansible_runner.run(private_data_dir='/home/florian/ansible-runner-test', playbook='get_facts.yaml', inventory="myhosts", passwords=passwords)
  File "/home/florian/.pyenv/versions/ansible-runner/lib/python3.8/site-packages/ansible_runner/interface.py", line 254, in run
    r = init_runner(**kwargs)
  File "/home/florian/.pyenv/versions/ansible-runner/lib/python3.8/site-packages/ansible_runner/interface.py", line 119, in init_runner
    rc.prepare()
  File "/home/florian/.pyenv/versions/ansible-runner/lib/python3.8/site-packages/ansible_runner/config/runner.py", line 144, in prepare
    self.prepare_env()
  File "/home/florian/.pyenv/versions/ansible-runner/lib/python3.8/site-packages/ansible_runner/config/runner.py", line 182, in prepare_env
    super(RunnerConfig, self)._prepare_env()
  File "/home/florian/.pyenv/versions/ansible-runner/lib/python3.8/site-packages/ansible_runner/config/_base.py", line 158, in _prepare_env
    for pattern, password in iteritems(self.passwords)
  File "/home/florian/.pyenv/versions/ansible-runner/lib/python3.8/site-packages/six.py", line 605, in iteritems
    return iter(d.items(**kw))
AttributeError: 'NoneType' object has no attribute 'items'
```

FloLaco:
[...]

PR #811, opened 2021-09-06
Fix issue with passwords when passwords is provided by dict and by file

FloLaco:
Issue: #809 

Fix the issue about `dict.update()` which return `None`. It's a bad usage of this method.

### DIFF
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -150,16 +150,24 @@ class BaseConfig(object):
         if self.runner_mode == 'pexpect':
             try:
                 if self.passwords and isinstance(self.passwords, dict):
-                    self.passwords.update(self.loader.load_file('env/passwords', Mapping))
+                    # Load password file and then the passwords dict argument for precedence
+                    temporary_passwords = dict() or self.loader.load_file('env/passwords', Mapping)
+                    temporary_passwords.update(self.passwords)
+                    self.passwords = temporary_passwords
+                    del temporary_passwords
                 else:
                     self.passwords = self.passwords or self.loader.load_file('env/passwords', Mapping)
-                self.expect_passwords = {
-                    re.compile(pattern, re.M): password
-                    for pattern, password in iteritems(self.passwords)
-                }
             except ConfigurationError:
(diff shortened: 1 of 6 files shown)


Shrews:
That certainly does look like problematic code. Could you add a test that validates the fix?

FloLaco:
@Shrews I've added some tests

Shrews:
recheck

ansible_runner/config/_base.py
```diff
@@ -150,16 +150,24 @@ def _prepare_env(self, runner_mode='pexpect'):
         if self.runner_mode == 'pexpect':
             try:
                 if self.passwords and isinstance(self.passwords, dict):
-                    self.passwords = self.passwords.update(self.loader.load_file('env/passwords', Mapping))
+                    # Load password file and then the passwords dict argument for precedence
```
Shrews, line 153: What's the reason for the change in precedence?

FloLaco: Because it's useless to provide password as dictionnary if a file is already present if we don't not modify precedence. 
Argument provided by "user" in python code should always be more important as config file.

AlanCoding: @Shrews made a good point to me

```python
self.settings.update(self.loader.load_file('env/settings', Mapping))
```

differs from your precedence rules for passwords here.

I agree with you sentiment

> Argument provided by "user" in python code should always be more important as config file.

But that would apply to settings, just the same that it applies to passwords.

FloLaco: @AlanCoding you're right.
I didn't change settings because I don't not use it and the initial fix was not intend to change settings. I just changed the obvious bug about settings.
If you insist I change change precedence with settings to, but I just need to know how to test it.

Shrews: IMO, all entries in `env` should have the same precedence. I feel like this PR should just fix the current bug. Changing precedence for file vs. args should likely be a separate PR and make sure that precedence is treated the same way across all possible input files.

samdoran:
Superseded by #899

PR #812, opened 2021-09-07
New feature: support the choice to store or not passwords into file

FloLaco:
Hi

When we provide passwords with `passwords` dictionnary into `run()` function, `ansible-runner` create the `env/passwords` file with the `passwords` as content.
This new feature aim to disable the storage, because password is a sensitive information.

How to provide dictionnary password (feature already exists) : 
```
passwords = {"SSH password:": "PASS"}
r = ansible_runner.run(private_data_dir='/', playbook='get_facts.yaml', inventory="myhosts", passwords=passwords)
```

How to disable the storage : 
```
passwords = {"SSH password:": "PASS"}
r = ansible_runner.run(private_data_dir='/', playbook='get_facts.yaml', inventory="myhosts", passwords=passwords, store_passwords=False)
```

### DIFF
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -59,7 +59,8 @@ class BaseConfig(object):
                  project_dir=None, artifact_dir=None, fact_cache_type='jsonfile', fact_cache=None,
                  process_isolation=False, process_isolation_executable=None,
                  container_image=None, container_volume_mounts=None, container_options=None, container_workdir=None,
-                 ident=None, rotate_artifacts=0, timeout=None, ssh_key=None, quiet=False, json_mode=False, check_job_event_data=False):
+                 ident=None, rotate_artifacts=0, timeout=None, ssh_key=None, quiet=False, json_mode=False,
+                 check_job_event_data=False, store_passwords=True):
         # common params
         self.host_cwd = host_cwd
         self.envvars = envvars
@@ -84,6 +85,7 @@ class BaseConfig(object):
         self.settings = settings
         self.timeout = timeout
         self.check_job_event_data = check_job_event_data
+        self.store_passwords = store_passwords
 
         # setup initial environment
         if private_data_dir:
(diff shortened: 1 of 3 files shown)


ansible_runner/config/_base.py
```diff
@@ -153,13 +155,17 @@ def _prepare_env(self, runner_mode='pexpect'):
...
                 debug('Not loading passwords')
                 self.expect_passwords = dict()
+            finally:
+                if self.passwords:
+                    self.expect_passwords = {
+                        re.compile(pattern, re.M): password
+                        for pattern, password in iteritems(self.passwords)
+                    }
+                else:
+                    self.expect_passwords = dict()
```
AlanCoding, line 168: this looks like a real bug fix :+1:

ansible_runner/utils/__init__.py
```diff
@@ -198,11 +198,12 @@ def dump_artifacts(kwargs):
                 kwargs['inventory'] = dump_artifact(obj, path, 'hosts')
 
     for key in ('envvars', 'extravars', 'passwords', 'settings'):
-        obj = kwargs.get(key)
-        if obj and not os.path.exists(os.path.join(private_data_dir, 'env', key)):
-            path = os.path.join(private_data_dir, 'env')
-            dump_artifact(json.dumps(obj), path, key)
-            kwargs.pop(key)
+        if key == 'passwords' and kwargs.get('store_passwords', True) or key != 'passwords':
```
AlanCoding, line 201: why just passwords as opposed to all of these, envvars/extravars/passwords/settings?

I think a similar security argument can apply to all of these, and I don't know that functionality would be harmed by not dumping any of them. So I'm interested to ask if we could have 1 configurable that will tell it to not dump _any_ of these. So instead of `store_passwords`, something like `store_inputs`, name a WIP.

AlanCoding:
I want some eyes from my team on this, ping @john-westcott-iv

ansible_runner/config/_base.py
john-westcott-iv: Wouldn't this actually now cause an issue if the re.compile fails and raises an exception? Before you would fall into the except but now you will get a stack.

AlanCoding: Problems there wouldn't raise a `ConfigurationError` so it wouldn't be caught anyway.

john-westcott-iv:
Replaced by https://github.com/ansible/ansible-runner/pull/1000

Shrews:
Closing due to being superceded by PR #1000.

PR #814, opened 2021-09-08
Add constraints and update `pytest` options

samdoran:
Adding constraints ensures consistent test execution and avoids spontaneous breakage due to package updates.

Add `pytest` options to colorize output and show a summary at the end to show all tests expect those that passed.


CONTRIBUTING.md
```diff
@@ -17,7 +17,7 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
 ```bash
 (host)$ pip install virtualenvwrapper
 (host)$ mkvirtualenv ansible-runner
-(host)$ pip install -e .
+(host)$ pip install -e . -r test/requirements.txt -c test/constraints.txt
```
Shrews: I don't think this makes sense to do. Tox is the thing that will install these requirements into the virtual environments that _it_ creates. You shouldn't need to install the requirements into the outer environment. It's unnecessary since they won't even be used for the tests when using tox.

samdoran: I went back and forth on this. The reason I ended up adding it is because `test/requirements.txt` contains `tox` (and `constraints.txt` constrains the version).

We would need to at least replace `-r test/requirements.txt` with `tox`. That may make more sense.

samdoran: I also found that `tox` is smart enough not to install `tox` inside the virtual environments it creates.

Shrews: `tox` shouldn't be in `test/requirements.txt`. The fact that this makes the dev install the requirements into the outer environment unnecessarily is a non-starter for me. Seems like the constraints should be added to the `tox.ini` entries instead?

samdoran: > tox shouldn't be in test/requirements.txt.

I'll remove it.

> Seems like the constraints should be added to the tox.ini entries instead?

I thought I had done that, but apparently I only aded the requirements to `deps` and not the constraints.

Shrews: Oh, I think I see the confusion here. The original docs are suggesting a way to _use_ runner utilizing `mkvirtualenv`. The testing environments are separate, so your change is sort of conflating the two.  A user should be able to _test_ runner without having to activate the environment used to run `ansible-runner` from the command line.

samdoran: Interesting. I missed that.

samdoran: Hmm, it seems like this is in the "Setting up your development environment" section. I don't think it's talking about a virtual environment for using runner. 😕

samdoran: Oh, I see what you mean. the `-e .` is for running from source in addition to being able to run tests. I'm so focused on being able to run tests I forgot about actually running runner.

CONTRIBUTING.md
```diff
@@ -8,34 +8,39 @@ If you have questions about this document or anything not covered here? Come cha
...
+pip install tox -c test/constraints.txt
 ```
 
-When done making changes, run:
+Setup a test virtual environment and activate it:
 
 ```
-(host)$ deactivate
+tox -e dev
+source .tox/dev/bin/activate
```
webknjaz, line 25: Did you know that tox has a `--devenv` option? It basically allows to create a regular venv out of a testenv that is defined in `tox.ini`.

So you'd use this as
```suggestion
tox -e dev --devenv dev-venv
. dev-venv/bin/activate
```

No need to rely on internal tox's envs. Plus, you could get an env for things that are more specific to what you need, like
```console
$ tox -e py38-ansible29 --devenv ans29-venv
```

refs:
* https://tox.readthedocs.io/en/latest/config.html#cmdoption-tox-devenv
* https://tox.readthedocs.io/en/latest/changelog.html#v3-13-0-2019-06-24

CONTRIBUTING.md
```diff
@@ -8,34 +8,39 @@ If you have questions about this document or anything not covered here? Come cha
...
-(host)$ pip install tox
-(host)$ tox
+source .tox/dev/bin/activate
 ```
+
+## Testing
+
+`tox` is used to run tests. To run the default set of tests, just run `tox`.
+
+To list all available test targets, run `tox -a`.
```
webknjaz, line 44: It's better to be verbose, it'll show descriptions for every env, then:
```suggestion
To list all available test targets, run `tox -av`.
```

Example output (from a navigator checkout):
```console
$ tox -av                         
using tox.ini: ~/wk/src/github/ansible/ansible-navigator--bugfixes--503-diff-local-ansible-interpreter/tox.ini (pid 781911)
using tox-3.23.0 from ~/wk/.pyenv/versions/3.9.0/lib/python3.9/site-packages/tox/__init__.py (pid 781911)
default environments:
linters    -> Enforce quality standards under ~/wk/.pyenv/versions/3.9.0/bin/python3.9 (~/wk/src/github/ansible/ansible-navigator--bugfixes--503-diff-local-ansible-interpreter/.tox/linters/bin/python)
type       -> Verify static typing with MyPy under ~/wk/.pyenv/versions/3.9.0/bin/python3.9 (~/wk/src/github/ansible/ansible-navigator--bugfixes--503-diff-local-ansible-interpreter/.tox/type/bin/python)
py38       -> Run pytest under python3.8 (~/wk/src/github/ansible/ansible-navigator--bugfixes--503-diff-local-ansible-interpreter/.tox/py38/bin/python)
report     -> Produce coverage report
clean      -> Erase coverage data

additional environments:
smoke      -> Run smoke tests under ~/wk/.pyenv/versions/3.9.0/bin/python3.9 (~/wk/src/github/ansible/ansible-navigator--bugfixes--503-diff-local-ansible-interpreter/.tox/smoke/bin/python)
build-docs -> Build The Docs
```

test/constraints.txt
```diff
@@ -0,0 +1,62 @@
+alabaster==0.7.12
```
webknjaz, line 1: I recommend managing this with https://pip-tools.rtfd.io. And I'd suggest separating pins for different envs: why do you need pins for testing against three completely different Ansible setups *and* docs? I don't think they should influence each other.

tox.ini
```diff
@@ -3,20 +3,23 @@ envlist = linters, ansible{27, 28, 29, -base}
...
+    ansible29: ansible<2.10
+    ansible-base: ansible-base
+    py{,3,38}: ansible-core
+    -r {toxinidir}/requirements.txt
+    -r {toxinidir}/test/requirements.txt
+    -c {toxinidir}/test/constraints.txt
 passenv = HOME
 usedevelop = True
 commands=
     py.test -v -n 4 -m "not serial" {posargs:test}
```
webknjaz, line 18: The default `test/` could be moved to the config under `testpaths`.

CONTRIBUTING.md
```diff
@@ -8,34 +8,39 @@ If you have questions about this document or anything not covered here? Come cha
...
+
+Run a specific test with `tox -e [target]`
```
webknjaz, line 46: How about also showing how to pass args to pytest?
```console
$ tox -e py -- -k some_name
```

pytest.ini
```diff
@@ -2,3 +2,6 @@
 markers =
     serial: tests that cannot be reliably ran with pytest multiprocessing
     timeout: used with pytest-timeout
+addopts =
+    --color yes
+    -r a
```
webknjaz, line 7: I recommend making use of the strict options.
Here's a few annotated examples I have: https://github.com/ansible/pylibssh/blob/devel/pytest.ini / https://github.com/cherrypy/cheroot/blob/master/pytest.ini.
Pay special attention at `--strict-markers` / `xfail_strict` / `filterwarnings = error` and convenience settings like `--showlocals` / `--durations=10`

tox.ini
```diff
@@ -25,6 +28,11 @@ commands=
     yamllint -s .
 
 [testenv:docs]
+description = Build docs
 deps = -r{toxinidir}/docs/requirements.txt
 commands =
     sphinx-build -E -W -d docs/build/doctrees -b html docs docs/build/html
```
webknjaz, line 34: Unrelated to this PR, I also recommend adding
```suggestion
    sphinx-build -E -n -W --keep-going -d docs/build/doctrees -b html docs docs/build/html
```

tox.ini
```diff
@@ -25,6 +28,11 @@ commands=
     yamllint -s .
 
 [testenv:docs]
+description = Build docs
 deps = -r{toxinidir}/docs/requirements.txt
```
webknjaz, line 32: (since you have alabaster/sphinx in the constraints)
```suggestion
deps = -r{toxinidir}/docs/requirements.txt -c{toxinidir}/test/constraints.txt
```

test/constraints.txt
webknjaz: Also, consider using sphinx-ansible-theme or Furo. Alabaster is half-dead.

Shrews:
So, after considering this, I'm generally not in favor of capping every dependency with constraints. Yes, it makes our CI _super_ stable, but I think it does a disservice to us and our users.  Reasons being:

- I firmly believe one of the important jobs of a CI is to catch breakages with our dependencies before users.
- It requires us to chase our dependencies and manually change each version.
- It's not realistic to how it would be used in other environments.
- I'm not a packaging expert, but I'd imagine this would make packaging our product much more complicated.

In general, these are the scenarios where I believe constraints are appropriate (off the top of my head):

- Using a feature of a dependency's API that requires using a minimum version.
- Avoiding a broken release of a dependency.
- A dependency makes a breaking change to their API, so we must set a maximum version until we adapt.

webknjaz:
@Shrews

*(with my Python packaging expert hat on)*

The common workflow is to have something like Dependabot enabled to have the process _fully automated_. It supports pip-tools and knows to update the constraints files on schedule. So it opens a PR when a dependency gets updates and you see if it fails or not in each case.
This is super useful and automatic most of the time. You'll still see if specific dependencies break things, this will be very explicit and you'll see the logs of how each pin upgrade affects the CI.
With this workflow, you'll still have direct dependencies open-ended and separate from the constraints. But you'll gain more control over the investigation of how the CI gets broken, with specific updates, not at random points in time.

Many major projects in the ecosystem do this. Here's an example of how it's set up for the PyPI https://github.com/pypa/warehouse/tree/main/requirements / https://github.com/pypa/warehouse/pull/9877/files.

As a side note, I know some projects go further, they separate sets of requirements — one has the oldest possible dependency versions and the other one has the very recent versions of the packages in the dep tree.

Feel free to ask me more about this if you need more insight.

PR #815, opened 2021-09-08
Fix unit tests to run consistently in CI and locally

samdoran:
- use pytest-mock and `tmp_path` fixtures
- mock test environment variables to prevent inconsistent test results
- reformat test fixtures so they are easier to read

This will resolve the test failures in #805.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -2,3 +2,10 @@
 markers =
     serial: tests that cannot be reliably ran with pytest multiprocessing
     timeout: used with pytest-timeout
+addopts =
+    -r a
+    --color yes
+    --showlocals
+    --verbose
+    --numprocesses auto
+    --durations 10
--- a/test/unit/config/test__base.py
+++ b/test/unit/config/test__base.py
@@ -250,8 +250,8 @@ def test_wrap_args_with_ssh_agent_silent():
 @patch('os.path.isdir', return_value=False)
 @patch('os.path.exists', return_value=True)
 @patch('os.makedirs', return_value=True)
-def test_container_volume_mounting_with_Z(mock_isdir, mock_exists, mock_makedirs, tmpdir):
-    rc = BaseConfig(private_data_dir=str(tmpdir))
+def test_container_volume_mounting_with_Z(mock_isdir, mock_exists, mock_makedirs, tmp_path):
+    rc = BaseConfig(private_data_dir=str(tmp_path))
     os.path.isdir = Mock()
     rc.container_volume_mounts = ['project_path:project_path:Z']
     rc.container_name = 'foo'
@@ -275,14 +275,14 @@ def test_container_volume_mounting_with_Z(mock_isdir, mock_exists, mock_makedirs
 
 
 @pytest.mark.parametrize('container_runtime', ['docker', 'podman'])
-def test_containerization_settings(tmpdir, container_runtime, mocker):
-    mocker.patch.dict('os.environ', {'HOME': str(tmpdir)}, clear=True)
-    os.mkdir(os.path.join(tmpdir, '.ssh'))
+def test_containerization_settings(tmp_path, container_runtime, mocker):
+    mocker.patch.dict('os.environ', {'HOME': str(tmp_path)}, clear=True)
+    tmp_path.joinpath('.ssh').mkdir()
 
     mock_containerized = mocker.patch('ansible_runner.config._base.BaseConfig.containerized', new_callable=PropertyMock)
     mock_containerized.return_value = True
 
-    rc = BaseConfig(private_data_dir=tmpdir)
+    rc = BaseConfig(private_data_dir=tmp_path)
     rc.ident = 'foo'
     rc.cmdline_args = ['main.yaml', '-i', '/tmp/inventory']
     rc.command = ['ansible-playbook'] + rc.cmdline_args
@@ -309,7 +309,7 @@ def test_containerization_settings(tmpdir, container_runtime, mocker):
         '--interactive',
         '--workdir',
         '/runner/project',
-        '-v', '{}/.ssh/:/home/runner/.ssh/'.format(str(tmpdir))
+        '-v', '{}/.ssh/:/home/runner/.ssh/'.format(str(tmp_path))
     ]
 
     if container_runtime == 'podman':
@@ -341,9 +341,9 @@ def test_containerization_settings(tmpdir, container_runtime, mocker):
         ('podman', '1')
     )
 )
-def test_containerization_unsafe_write_setting(tmpdir, container_runtime, expected):
+def test_containerization_unsafe_write_setting(tmp_path, container_runtime, expected):
     with patch('ansible_runner.config._base.BaseConfig.containerized', new_callable=PropertyMock) as mock_containerized:
-        rc = BaseConfig(private_data_dir=tmpdir)
+        rc = BaseConfig(private_data_dir=tmp_path)
         rc.ident = 'foo'
         rc.cmdline_args = ['main.yaml', '-i', '/tmp/inventory']
         rc.command = ['ansible-playbook'] + rc.cmdline_args
--- a/test/unit/config/test__base.py
+++ b/test/unit/config/test__base.py
@@ -299,7 +299,7 @@ def test_containerization_settings(tmp_path, container_runtime, mocker):
     if container_runtime == 'podman':
         extra_container_args = ['--quiet']
     else:
-        extra_container_args = ['--user={os.getuid()}']
+        extra_container_args = [f'--user={os.getuid()}']
 
     expected_command_start = [
         container_runtime,
@@ -328,11 +328,7 @@ def test_containerization_settings(tmp_path, container_runtime, mocker):
         'my_container', 'ansible-playbook', 'main.yaml', '-i', '/tmp/inventory',
     ])
 
-    for index, element in enumerate(expected_command_start):
-        if '--user=' in element:
-            assert '--user=' in rc.command[index]
-        else:
-            assert rc.command[index] == element
+    assert expected_command_start == rc.command
 
 
 @pytest.mark.parametrize(
--- a/test/unit/config/test__base.py
+++ b/test/unit/config/test__base.py
@@ -275,21 +275,25 @@ def test_container_volume_mounting_with_Z(mock_isdir, mock_exists, mock_makedirs
 
 
 @pytest.mark.parametrize('container_runtime', ['docker', 'podman'])
-def test_containerization_settings(tmpdir, container_runtime):
-    with patch('ansible_runner.config._base.BaseConfig.containerized', new_callable=PropertyMock) as mock_containerized:
-        rc = BaseConfig(private_data_dir=tmpdir)
-        rc.ident = 'foo'
-        rc.cmdline_args = ['main.yaml', '-i', '/tmp/inventory']
-        rc.command = ['ansible-playbook'] + rc.cmdline_args
-        rc.process_isolation = True
-        rc.runner_mode = 'pexpect'
-        rc.process_isolation_executable = container_runtime
-        rc.container_image = 'my_container'
-        rc.container_volume_mounts = ['/host1:/container1', 'host2:/container2']
-        mock_containerized.return_value = True
-        rc.execution_mode = BaseExecutionMode.ANSIBLE_COMMANDS
-        rc._prepare_env()
-        rc._handle_command_wrap(rc.execution_mode, rc.cmdline_args)
+def test_containerization_settings(tmpdir, container_runtime, mocker):
+    mocker.patch.dict('os.environ', {'HOME': str(tmpdir)}, clear=True)
+    os.mkdir(os.path.join(tmpdir, '.ssh'))
+
+    mock_containerized = mocker.patch('ansible_runner.config._base.BaseConfig.containerized', new_callable=PropertyMock)
+    mock_containerized.return_value = True
+
+    rc = BaseConfig(private_data_dir=tmpdir)
+    rc.ident = 'foo'
+    rc.cmdline_args = ['main.yaml', '-i', '/tmp/inventory']
+    rc.command = ['ansible-playbook'] + rc.cmdline_args
+    rc.process_isolation = True
+    rc.runner_mode = 'pexpect'
+    rc.process_isolation_executable = container_runtime
+    rc.container_image = 'my_container'
+    rc.container_volume_mounts = ['/host1:/container1', 'host2:/container2']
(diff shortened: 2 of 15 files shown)


test/unit/config/test__base.py
```diff
@@ -275,39 +275,58 @@ def test_container_volume_mounting_with_Z(mock_isdir, mock_exists, mock_makedirs
...
-        rc.process_isolation = True
-        rc.runner_mode = 'pexpect'
-        rc.process_isolation_executable = container_runtime
-        rc.container_image = 'my_container'
-        rc.container_volume_mounts = ['/host1:/container1', 'host2:/container2']
-        mock_containerized.return_value = True
-        rc.execution_mode = BaseExecutionMode.ANSIBLE_COMMANDS
-        rc._prepare_env()
-        rc._handle_command_wrap(rc.execution_mode, rc.cmdline_args)
+def test_containerization_settings(tmpdir, container_runtime, mocker):
```
webknjaz: Nit: it is best to use `tmp_path` in place of the deprecated `tmpdir` because it has a pathlib interface (and is recommended generally).

samdoran: Done.

pytest.ini
```diff
@@ -2,3 +2,10 @@
 markers =
```
webknjaz, line 2: Since you're using markers, I suggest adding `--strict-markers` so it'd be forbidden to use random marker names.

samdoran: I actually don't know what that is there for or if we need it.

AlanCoding (approved):

test/unit/config/test_doc.py
```diff
@@ -81,25 +84,41 @@ def test_prepare_plugin_docs_command_with_containerization(tmpdir, container_run
     if container_runtime == 'podman':
         extra_container_args = ['--quiet']
     else:
-        extra_container_args = ['--user={os.getuid()}']
+        extra_container_args = [f'--user={os.getuid()}']
```
Shrews, line 87: Yay f-strings!

pytest.ini
Shrews: I'm not familiar with this either, but let's stay focused on the changes Sam is making.

Shrews (commented):
Haven't looked into why the tests are failing, but I'm on board with all of these changes. Love the `s/tmpdir/tmp_path/` part.

Shrews:
recheck

That SKIPPED zuul job is weird.

Shrews:
recheck

samdoran:
There are some test failures I need to work out. I can't run the integration tests locally, so I just have to kick them up and see what fails. 🤞

samdoran:
I think I've fixed the test failures related to switch to `tmp_path`. It's a `pathlib.Path` object so the `mkdir()` method has a slightly different behavior.

samdoran:
Latest failure is because `pathlib.Path.readlink()` is Python >= 3.9 only and I only tested in Python 3.9 locally. Hopefully that's the last failure.

samdoran:
These tests make me sad. :(

AlanCoding:
You can do it! ansible-runner needed a hero!

pytest.ini
```diff
@@ -2,3 +2,10 @@
...
     serial: tests that cannot be reliably ran with pytest multiprocessing
     timeout: used with pytest-timeout
+addopts =
+    -r a
+    --color yes
+    --showlocals
+    --verbose
+    --numprocesses auto
```
AlanCoding, line 10: Testing your branch, and I want to voice a preference that we not put this in the pytest config (where it will affect running locally), because running with multiprocessing makes debugging much harder. Options like `-s` are unusable.

AlanCoding: It also makes all the test commands fail for me for reasons I don't fully understand yet. Maybe how my virtual environment carries over into the subprocesses, but not sure

AlanCoding:
Checking out your branch and running locally, here's the failure I get (for a lot of tests):

https://gist.github.com/AlanCoding/15377a3099068506a6bfdb05b9b740d4

So there's still some brittleness somehow.

It confusingly gives the help text. Looking into the code path, I printed out `sys.argv` (right inside the `main` method) with and without multiprocessing turned on.

 - off: `['/home/alancoding/repos/awx/env/bin/py.test', 'test/integration/test_main.py', '-k', 'test_role_logfile_abs']`
 - on:  `['-c']`

And it turns out that `sys.argv` is used to make decisions, so my failures with multiprocessing on come down to this bit of code.

https://github.com/ansible/ansible-runner/blob/77f15a45b150dcd5b43d6a3fb0aec77c31f7f688/ansible_runner/__main__.py#L757-L760

Tests are ran with the pytest sys.argv (which are different for the parallel tests), and not the intended `sys_args` kwarg that the test is trying to pass.

The only coherent way I can see out of this is

```diff
diff --git a/ansible_runner/__main__.py b/ansible_runner/__main__.py
index d6222a1..fe2640c 100644
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -754,7 +754,7 @@ def main(sys_args=None):
     add_args_to_parser(isalive_container_group, DEFAULT_CLI_ARGS['container_group'])
     add_args_to_parser(transmit_container_group, DEFAULT_CLI_ARGS['container_group'])
 
-    if len(sys.argv) == 1:
+    if sys_args is None:
         parser.print_usage()
         print_common_usage()
         parser.exit(status=0)
```

I don't know why Zuul was not hitting the same failures as me, but maybe it's because all tests run on `gw0`, suggesting multiprocessing scaling automatically to 1 process... but that's wild speculation.

samdoran:
Yup, I ran into this reliance on `sys.argv` before. I [ended up mocking it](https://github.com/ansible/ansible-runner/pull/806/commits/183625ad020b1f8967e4bf750ca99106ef11e5f8) in the tests to get it passing. I wasn't sure if that was a problem with the tests or if the code should be changed.

samdoran:
> I don't know why Zuul was not hitting the same failures as me, but maybe it's because all tests run on gw0, suggesting multiprocessing scaling automatically to 1 process... but that's wild speculation.

It does seem to me that the tests are passing by accident currently. Not a great state of affairs.

AlanCoding:
> Yup, I ran into this reliance on sys.argv before. I ended up mocking it in the tests to get it passing. I wasn't sure if that was a problem with the tests or if the code should be changed.

I find the code's use of `sys.argv` there to be indefensible. It looks like a programming error. If it's easy for you to incorporate it, then go ahead, otherwise I would like to return to it and remove such mocks in a followup PR. You've already got a lot of good work here, and I'd like to do whatever I can to keep things moving forward.

pytest.ini
samdoran: My aim with adding these options was to have `pytest` run the same locally as it does in CI. But I have also felt the friction this adds. It's possible to disable multiprocessing with `tox -e py3 -- -n0`, but it's a bit of a pain.

I actually like the way `ansible-test` does this more dynamically with a cap, but `tox` doesn't have a way to do that easily (it can be done with a rather ungainly `shell` command, but I don't care for that approach).

I also want to move to using `--forked` for tests as well for tests that are not thread-safe. I'm hoping that will remove the need for the `-m serial` and `-m "not serial"` commands, but I'm not 100% sure why those are needed.

Anyway, what do you think about being able to do `-- -n0` locally in order to have the default behavior be the same locally and in CI?

AlanCoding: Yeah, I've come around to see the wisdom of `pytest.ini` keeping local runs synced with how CI works.

The main issue I had was that tests were failing due to the bug we've already discussed here. Solution there is to fix the bug! Running locally, `-n0` doesn't seem like too much hassle to expect of people.

pytest.ini
webknjaz: Pytest lets you create markers on the fly. So you can use `@pytest.mark.something-undocumented`. It is recommended that only custom markers listed here are allowed, for maintainability.
But I agree that this isn't necessarily related to this PR.
I only mentioned this because I saw Sam adding other pytest config options.

pytest.ini
webknjaz: Note that if you use `--pdb` it also tends to seamlessly turn the runs into `-s -n0`.

test/unit/config/test__base.py
```diff
@@ -275,45 +275,60 @@ def test_container_volume_mounting_with_Z(mock_isdir, mock_exists, mock_makedirs
...
-        rc.runner_mode = 'pexpect'
-        rc.process_isolation_executable = container_runtime
-        rc.container_image = 'my_container'
-        rc.container_volume_mounts = ['/host1:/container1', 'host2:/container2']
-        mock_containerized.return_value = True
-        rc.execution_mode = BaseExecutionMode.ANSIBLE_COMMANDS
-        rc._prepare_env()
-        rc._handle_command_wrap(rc.execution_mode, rc.cmdline_args)
+def test_containerization_settings(tmp_path, container_runtime, mocker):
+    mocker.patch.dict('os.environ', {'HOME': str(tmp_path)}, clear=True)
```
webknjaz, line 279: Have you considered using the built-in `monkeypatch.setenv()`?

https://docs.pytest.org/en/latest/how-to/monkeypatch.html

samdoran: I haven't yet figured out when to use `monkeypatch` vs `mocker.patch`. I still have things to learn. :)

AlanCoding: out of the 3 options outlined here:

https://github.com/pytest-dev/pytest/issues/4576#issuecomment-449864333

I have seen a lot converging on option (2), because python2 tends to be in the rear view mirror by now. The pattern `from unittest import mock` already exists in ansible-runner unit tests. However, the pytest-mock you're adding here is more naturally aware of the pytest test scope, which means you don't have to indent everything in context managers. Thus, I raise no objections to it.

pytest.ini
samdoran: Alright, I'll work on fixing the actual bug. I do love how `--pdb` and `-n auto` play nicely together.

test/unit/config/test__base.py
samdoran: > The pattern from unittest import mock already exists in ansible-runner unit tests.

I have a stashed commit locally that removes all of that in favor of the `pytest-mock` plugin. It ended up being pretty huge and took me several hours to get working which is why I didn't include those changes in this PR. I'll do that in (yet another) PR.

samdoran: I had a good talk with @webknjaz about `monkeypach` vs `mocker.patch` and I have a better idea of when to use each now. `monkeypatch` doesn't have the ability to clear the environment, but I'm not sure that is strictly necessary. I will try it out see if it works.

webknjaz: _Technically_, `monkeypatch.setattr(os, 'environ', {'HOME': str(tmp_path)})` would probably emulate this. But I agree, it's usually unnecessary to clear all the other unnecessary env vars. If you expect certain env vars not to be set, use `monkeypatch.delenv()` for that.

webknjaz: > out of the 3 options outlined here:
> 
> [pytest-dev/pytest#4576 (comment)](https://github.com/pytest-dev/pytest/issues/4576#issuecomment-449864333)
> 

I think that what Ronny points out in the very next comment is even more important to take into account: https://github.com/ansible/ansible-runner/pull/815#discussion_r707453680.

TL;DR If you end up using any sort of mocking technique, you probably have an architectural problem. If you were to just use dependency injection, there would be exactly zero need for this.

samdoran:
The [`sys.argv` docs](https://docs.python.org/3/library/sys.html#sys.argv) provide an explanation of where the `-c` comes from:
> If the command was executed using the -c command line option to the interpreter, argv[0] is set to the string '-c'.

`pytest-xdist` [obscures the original command line args](https://github.com/pytest-dev/pytest-xdist/issues/670#issuecomment-858873092) in `sys.argv`.

webknjaz:
Oh, if you only new what hacks CPython does with `-c` and `-m` :exploding_head:

webknjaz:
FWIW I believe it's wrong to rely on argv in tests. One should fall an entrypoint function where it's passed in as an arg.

AlanCoding:
Yes, I only highlighted the `sys.argv` to illustrate that they were pytest args, as opposed to ansible-runner args. The code does ogles of processing of args but only directly references `sys.argv` in one suspiciously trivial use. This was a bug, and writing tests to accommodate the bug is not ideal.

webknjaz:
> This was a bug, and writing tests to accommodate the bug is not ideal.

This is a bug in ansible-runner (and its tests), though.

AlanCoding:
I still believe [this diff](https://github.com/ansible/ansible-runner/pull/815#issuecomment-917795923) would fix the bug, sorry if I haven't communicated very clearly.

samdoran:
@AlanCoding Unfortunately, it doesn't. I spent all day yesterday and all morning on a fix. It's quite a bit more complicated than I had hoped, though I'm hoping we can go with a much simpler fix than what I came up with in order to keep the existing behavior.

AlanCoding:
The only failure I see in CI right now is `test_role_logfile_abs`, which pretty clearly cannot be replicated locally.

It's true that I can't run tests in parallel locally using this branch, but Zuul seems to use 1 process.

So what are you still trying to accomplish at the moment? If you turn off parallel tests you won't get any difference performance. You could skip the 1 flaky test, or maybe it won't even fail if you turn of parallel tests. These are small issues amid a large volume of generally agreed-upon changes.

test/unit/config/test__base.py
samdoran: > If you end up using any sort of mocking technique, you probably have an architectural problem. If you were to just use dependency injection, there would be exactly zero need for this.

I absolutely agree with this. I am huge fan of test driven development for this reason, among others. Currently, I'm just trying to improve the state of the tests. Eventually I hope to improve the code itself.

samdoran:
> The only failure I see in CI right now is test_role_logfile_abs, which pretty clearly cannot be replicated locally.

I am currently just trying to fix this last failure.

samdoran:
Glad to know I'm not the only one that can't duplicate this locally. It's a head scratcher.

AlanCoding:
@samdoran my best guess is the fact that `ensure_removed("new_logfile")` occurs twice in that file, indicating that the sister test may be removing the file before the `os.path.exists` assertion.

So, obvious fix may be to name it "new_logfile1"

samdoran:
I managed to duplicate it locally. It was the `len(sys.argv)` line again. 🙄 I figured out the failure.

samdoran:
Ho-lee-cow. It passed. 🙌

Shrews (approved):

PR #816, opened 2021-09-08
add fqcn to job events

chrismeyersfsu:
Have not tested.


samdoran:
Change looks good. Just needs to be verified and we need tests.

PR #817, opened 2021-09-08
Update Ansible Runner to Pull in Authentication Information for Protected Registries

beeankha:
Connecting https://github.com/ansible/awx/issues/10944

### DIFF
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -535,10 +535,12 @@ class BaseConfig(object):
         encrypted_container_auth_data = {'auths': {host: {'auth': b64encode(token.encode('UTF-8')).decode('UTF-8')}}}
         # Create a new temp file with container auth data
         path = tempfile.mkdtemp(prefix='ansible_runner_registry_%s_' % self.ident)
+
         @atexit.register
         def cleanup_ansible_runner_registry():
             shutil.rmtree(path)
-        with open(path + '/auth.json', 'w') as authfile:
+        registry_auth_path = os.path.join(path, 'auth.json')
+        with open(registry_auth_path, 'w') as authfile:
             os.chmod(authfile.name, stat.S_IRUSR | stat.S_IWUSR)
             authfile.write(json.dumps(encrypted_container_auth_data, indent=4))
         return authfile.name
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -451,7 +451,7 @@ class BaseConfig(object):
 
         # For run() and run_async() API value of base execution_mode is 'BaseExecutionMode.NONE'
         # and the container volume mounts are handled separately using 'container_volume_mounts'
-        # hence ignore additonal mount here
+        # hence ignore additional mount here
         if execution_mode != BaseExecutionMode.NONE:
             if execution_mode == BaseExecutionMode.ANSIBLE_COMMANDS:
                 self._handle_ansible_cmd_options_bind_mounts(new_args, cmdline_args)
@@ -532,7 +532,7 @@ class BaseConfig(object):
 
     def _generate_container_auth_file(self, host, username, password):
         token = "{}:{}".format(username, password)
-        encrypted_container_auth_data = {'auths': {host: {'auth': b64encode(token.encode('UTF-8')).decode('UTF-8')}}}
+        encoded_container_auth_data = {'auths': {host: {'auth': b64encode(token.encode('UTF-8')).decode('UTF-8')}}}
         # Create a new temp file with container auth data
         path = tempfile.mkdtemp(prefix='ansible_runner_registry_%s_' % self.ident)
 
@@ -542,7 +542,7 @@ class BaseConfig(object):
         registry_auth_path = os.path.join(path, 'auth.json')
         with open(registry_auth_path, 'w') as authfile:
             os.chmod(authfile.name, stat.S_IRUSR | stat.S_IWUSR)
-            authfile.write(json.dumps(encrypted_container_auth_data, indent=4))
+            authfile.write(json.dumps(encoded_container_auth_data, indent=4))
         return authfile.name
 
     def wrap_args_with_ssh_agent(self, args, ssh_key_path, ssh_auth_sock=None, silence_ssh_add=False):
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -471,17 +471,6 @@ class BaseConfig(object):
                 subdir_path = os.path.join(self.private_data_dir, subdir)
                 if not os.path.exists(subdir_path):
                     os.mkdir(subdir_path, 0o700)
-        if self.container_auth_data:
-            # Pull in the necessary registry auth info, if there is a container cred
-            host = self.container_auth_data.get('host')
-            username = self.container_auth_data.get('username')
-            password = self.container_auth_data.get('password')
-            self.registry_auth_path = self._generate_container_auth_file(host, username, password)
-            if 'podman' in self.process_isolation_executable:
-                new_args.extend(["--authfile={}".format(self.registry_auth_path)])
-            else:
-                docker_idx = new_args.index(self.process_isolation_executable)
-                new_args.insert(docker_idx + 1, "--config={}".format(self.registry_auth_path))
 
             # runtime commands need artifacts mounted to output data
             self._update_volume_mount_paths(new_args,
@@ -503,6 +492,18 @@ class BaseConfig(object):
             # Mount the entire private_data_dir
             # custom show paths inside private_data_dir do not make sense
             self._update_volume_mount_paths(new_args, "{}".format(self.private_data_dir), dst_mount_path="/runner", labels=":Z")
+
+        if self.container_auth_data:
+            # Pull in the necessary registry auth info, if there is a container cred
+            host = self.container_auth_data.get('host')
+            username = self.container_auth_data.get('username')
+            password = self.container_auth_data.get('password')
+            self.registry_auth_path = self._generate_container_auth_file(host, username, password)
+            if 'podman' in self.process_isolation_executable:
+                new_args.extend(["--authfile={}".format(self.registry_auth_path)])
+            else:
+                docker_idx = new_args.index(self.process_isolation_executable)
+                new_args.insert(docker_idx + 1, "--config={}".format(self.registry_auth_path))
 
         if self.container_volume_mounts:
             for mapping in self.container_volume_mounts:
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -77,6 +77,7 @@ class BaseConfig(object):
         self.container_volume_mounts = container_volume_mounts
         self.container_workdir = container_workdir
         self.container_auth_data = container_auth_data
+        self.registry_auth_path = None
         self.container_name = None  # like other properties, not accurate until prepare is called
         self.container_options = container_options
         self._volume_mount_paths = []
@@ -475,8 +476,8 @@ class BaseConfig(object):
             host = self.container_auth_data.get('host')
             username = self.container_auth_data.get('username')
             password = self.container_auth_data.get('password')
-            registry_auth_path = self._generate_container_auth_file(host, username, password)
-            new_args.extend(["--authfile={}".format(registry_auth_path)])
+            self.registry_auth_path = self._generate_container_auth_file(host, username, password)
+            new_args.extend(["--authfile={}".format(self.registry_auth_path)])
 
             # runtime commands need artifacts mounted to output data
             self._update_volume_mount_paths(new_args,
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -477,7 +477,11 @@ class BaseConfig(object):
             username = self.container_auth_data.get('username')
             password = self.container_auth_data.get('password')
             self.registry_auth_path = self._generate_container_auth_file(host, username, password)
-            new_args.extend(["--authfile={}".format(self.registry_auth_path)])
+            if 'podman' in self.process_isolation_executable:
+                new_args.extend(["--authfile={}".format(self.registry_auth_path)])
+            else:
+                docker_idx = new_args.index(self.process_isolation_executable)
+                new_args.insert(docker_idx + 1, "--config={}".format(self.registry_auth_path))
 
             # runtime commands need artifacts mounted to output data
             self._update_volume_mount_paths(new_args,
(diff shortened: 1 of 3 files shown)


ansible_runner/config/_base.py
```diff
@@ -516,6 +530,19 @@ def wrap_args_for_containerization(self, args, execution_mode, cmdline_args):
...
         return new_args
 
+    def _generate_container_auth_file(self, host, username, password):
+        token = "{}:{}".format(username, password)
+        encrypted_container_auth_data = {'auths': {host: {'auth': b64encode(token.encode('UTF-8')).decode('UTF-8')}}}
+        # Create a new temp file with container auth data
+        path = tempfile.mkdtemp(prefix='ansible_runner_registry_%s_' % self.ident)
+        @atexit.register
+        def cleanup_ansible_runner_registry():
+            shutil.rmtree(path)
```
beeankha, line 547: I've confirmed that the `auth.json` file along with the `ansible_runner_registry_*` directory is successfully deleted via this function that's decorated with`@atexit.register`.

beeankha:
This depends on changes being made to AWX's `awx/main/tasks.py` file to work properly, but I would like to get some feedback first before moving forward with the work on that codebase.

ansible_runner/config/_base.py
```diff
@@ -516,6 +530,19 @@ def wrap_args_for_containerization(self, args, execution_mode, cmdline_args):
...
+        with open(path + '/auth.json', 'w') as authfile:
```
AlanCoding: you should use `os.path.join` to add "auth.json" to path

AlanCoding (approved):

matburt (changes requested):
This looks good to me. I would like to see an update to docs to include this information though.

Shrews:
This needs a test to exercise the new code.

ansible_runner/config/_base.py
```diff
@@ -516,6 +530,21 @@ def wrap_args_for_containerization(self, args, execution_mode, cmdline_args):
         logger.debug(f"container engine invocation: {' '.join(new_args)}")
         return new_args
 
+    def _generate_container_auth_file(self, host, username, password):
+        token = "{}:{}".format(username, password)
+        encrypted_container_auth_data = {'auths': {host: {'auth': b64encode(token.encode('UTF-8')).decode('UTF-8')}}}
+        # Create a new temp file with container auth data
+        path = tempfile.mkdtemp(prefix='ansible_runner_registry_%s_' % self.ident)
```
chrismeyersfsu, line 543: Should this file be in `self.private_data_dir` ?

I think this would make it easier to clean up & debug.

AlanCoding: no, it can't be there, that's the point

these are credentials for pulling the container image. We specifically intend to _not_ make those secrets available to the playbook. The playbook runs in the volume mounted private_data_dir. That rules out that folder, which is why we started doing this in a wrapper folder, but that doesn't get transmitted.

chrismeyersfsu: Thanks @AlanCoding I see why this shouldn't be in `self.private_data_dir`.

I still have concerns about cleanup and debugging. I can't remember, is the parent directory of `self.private_data_dir` something we could use? 

I _do_ like that the `mkdtemp` call uses a prefix so that, given a job id, we can trace/find the directory that this `auth.json` lives.

AlanCoding: > I do like that the mkdtemp call uses a prefix so that, given a job id, we can trace/find the directory that this auth.json lives.

We plan on doing a last-ditch cleanup action in case something happens that prevents the `atexit` methods from running and deleting the directories. For that last-resort cleanup, we will do the same thing we did with isolated jobs. For this, we don't look for a specific job id, but instead list directories based on a pattern and exclude job ids which may still be running.

At the moment, however, we don't yet have cleanup for the `private_data_dir` on execution nodes _at all_. This was just never programmed into `ansible-runner worker` in the first place. There are no major issues in doing any of this, we just need to get it done. Our current plan still necessitates several additions to runner, and I just need to get a good cadence of merges happening.

ansible_runner/config/_base.py
```diff
@@ -187,6 +193,7 @@ def _prepare_env(self, runner_mode='pexpect'):
         self.container_image = self.settings.get('container_image', self.container_image)
         self.container_volume_mounts = self.settings.get('container_volume_mounts', self.container_volume_mounts)
         self.container_options = self.settings.get('container_options', self.container_options)
+        self.container_auth_data = self.settings.get('container_auth_data', self.container_auth_data)
```
Shrews, line 197: You have not added code for handling the option in the CLI (`ansible-runner/__main__.py`).

Shrews (changes requested):
This option is not handled via the CLI. I agree with @matburt about the need to update the docs. And all new feature code should include tests.

ansible_runner/config/_base.py
AlanCoding: This feature does not make any sense for `ansible-runner run`, because if you had the relevant host / username / password you should instead use a command like `podman login` to set up the config in the user-specific proper location.

However, it is completely unreasonable to login if you are running remotely.

If anything, this should be CLI options attached to `ansible-runner transmit`, because this is where AWX needs it, and it is the only use case where it makes sense to transmit the auth for a container registry. The need for this is created by the mutual requirements of (1) providing container registry credentials and (2) not exposing those credentials to the job runtime environment.

For now, I certainly think it is sufficient to leave this as a feature of the python interface.

ansible_runner/config/_base.py
```diff
@@ -443,8 +450,8 @@ def wrap_args_for_containerization(self, args, execution_mode, cmdline_args):
         new_args.extend(["--workdir", workdir])
 
         # For run() and run_async() API value of base execution_mode is 'BaseExecutionMode.NONE'
-        # and the container volume mounts are handled seperately using 'container_volume_mounts'
-        #  hence ignore additonal mount here
+        # and the container volume mounts are handled separately using 'container_volume_mounts'
+        # hence ignore additonal mount here
```
jbradberry: while we're fixing typos, `additonal` -> `additional`

AlanCoding (approved):

matburt (approved):
lgtm

ansible_runner/config/_base.py
jladdjr: @Shrews just wanted to check in again on this. Is this still a concern?

beeankha:
recheck

test/integration/containerized/test_container_management.py
```diff
@@ -70,3 +73,65 @@ def test_cancel_will_remove_container(test_data_dir, container_runtime_installed
...
+    pdd_path.mkdir()
+    private_data_dir = str(pdd_path)
+
+    image_name = 'quay.io/kdelee/does-not-exist'
+
+    res = run(
+        private_data_dir=private_data_dir,
+        playbook='ping.yml',
+        settings={
+            'process_isolation_executable': 'podman',
```
Shrews: We're currently using docker for the `integration` test suite, so this is unlikely to pass.

test/integration/containerized/test_container_management.py
```diff
@@ -70,3 +73,65 @@ def test_cancel_will_remove_container(test_data_dir, container_runtime_installed
...
+def test_registry_auth_file_cleanup(tmp_path, cli):
+    pdd_path = tmp_path / "private_data_dir"
+    pdd_path.mkdir()
+    private_data_dir = str(pdd_path)
+
+    auth_registry_glob = '/tmp/ansible_runner_registry_*'
+    registry_files_before = set(glob(auth_registry_glob))
+
+    settings_data = {
+        'process_isolation_executable': 'podman',
```
Shrews: Ditto on this. s/podman/docker/

Shrews (approved):
Fine with these changes in general, but I think you need to correct the new tests.

test/integration/containerized/test_container_management.py
```diff
@@ -70,3 +73,65 @@ def test_cancel_will_remove_container(test_data_dir, container_runtime_installed
     assert not is_running(
         cli, container_runtime_installed, 'ansible_runner_foo_bar'
     ), 'Found a running container, they should have all been stopped'
+
+
+def test_invalid_registry_host(tmp_path):
```
samdoran: Tests that require a specific command should be skipped.
```suggestion
@pytest.mark.skipif(shutil.which('podman') is None, reason="podman not available")
def test_invalid_registry_host(tmp_path):
```

samdoran: But as @Shrews mentioned, these should be changed to use Docker, or tested with both `podman` and `docker` skipping as appropriate.

samdoran: As an example:
```suggestion
@pytest.mark.parametrize('program', ('docker', 'podman'))
def test_invalid_registry_host(tmp_path, program):
    if shutil.which(program) is None:
        pytest.skip(f'{program} is not available')
```

Also needs to be changed on line 89:
```python
            'process_isolation_executable': program,
```

Shrews: @samdoran's suggestion to use `parametrize`  is the proper way to handle this.

samdoran: There are actually fixtures for this already, but they are one or the other currently.

ansible_runner/config/_base.py
```diff
@@ -463,6 +471,13 @@ def wrap_args_for_containerization(self, args, execution_mode, cmdline_args):
                 subdir_path = os.path.join(self.private_data_dir, subdir)
                 if not os.path.exists(subdir_path):
                     os.mkdir(subdir_path, 0o700)
+        if self.container_auth_data:
+            # Pull in the necessary registry auth info, if there is a container cred
+            host = self.container_auth_data.get('host')
+            username = self.container_auth_data.get('username')
+            password = self.container_auth_data.get('password')
+            self.registry_auth_path = self._generate_container_auth_file(host, username, password)
+            new_args.extend(["--authfile={}".format(self.registry_auth_path)])
```
samdoran: `--authfile` is not a valid flag for `docker run`. I'm not sure if it is a goal of this project to make it work seamlessly with both `docker` and `podman`. At a minimum, this means we have no way to test this functionality in our current CI environment.

ansible_runner/config/_base.py
samdoran: It seems like we should aim to keep parody between the Python interface and the CLI. I suppose if we have to deviate anywhere, first implementing a new feature in the Python interface before exposing it on the CLI is what I would prefer.

samdoran:
Most of the test failures are fixed in #815 (I'm still flogging that one). 

The new tests are failing as expected:

```
>       assert image_name in result_stdout
E       AssertionError: assert 'quay.io/kdelee/does-not-exist' in 'The command was not found or was not executable: podman.\n'
```

Unfortunately switching to using `docker` won't work since this change uses a flag, `--authfile`, that `docker` does not support.

This puts us in the uncomfortable situation of needing to 1) improve CI to allow running `podman` or 2) skip these tests and merge code we are not testing. I am not a fan of option two, especially if this code will be in a supported product.

I do plan to add `podman` support to our CI, but I don't have time to work on that currently.

samdoran:
I see what you did there. :wink:

You'll probably also need that on `test_registry_auth_file_cleanup`.

ansible_runner/config/_base.py
shanemcd: This is a good find. If we do choose to merge this, we need to make a note to address this in a follow-up PR ASAP.

beeankha:
> You'll probably also need that on `test_registry_auth_file_cleanup`.

@samdoran that test checks if the auth file got cleaned up, so even if the specified container runtime isn't available, it will pass because that file gets deleted even if the command was not successful

ansible_runner/config/_base.py
shanemcd: On second thought, we should just document that this feature only works with Podman.

shanemcd (approved):
Nice comprehensive PR.

test/integration/containerized/test_container_management.py
```diff
@@ -70,3 +73,73 @@ def test_cancel_will_remove_container(test_data_dir, container_runtime_installed
     assert not is_running(
         cli, container_runtime_installed, 'ansible_runner_foo_bar'
     ), 'Found a running container, they should have all been stopped'
+
+
+@pytest.mark.skipif(shutil.which('podman') is None and shutil.which('docker') is None, reason='podman and docker are not available')
+@pytest.mark.parametrize('runtime', ['podman', 'docker'])
+def test_invalid_registry_host(tmp_path, runtime):
```
samdoran: This will be skipped unless both `podman` and `docker` are available, which is not the case in our current CI. Ideally we would test 

Since the change now works with `podman` and `docker`, you could use the `container_runtime_installed` fixture to get one of the available runtimes. There would still be a gap in coverage, but it's better than skipping the test completely.
```suggestion
def test_invalid_registry_host(tmp_path, runtime, container_runtime_installed):
```

Or you could skip inside the test.
```suggestion
@pytest.mark.parametrize('runtime', ['podman', 'docker'])
def test_invalid_registry_host(tmp_path, runtime):
    if shutil.which(runtime) is None:
        pytest.skip(f'{runtime} is unavaialble')
```

That might be better since it shows the skipped tests.

Shrews:
FWIW, it seems that because you've added a new command line option, this is causing some rather fragile tests to fail based on your change. PR #815 would help with that situation, so we'd either need to merge that before this, or you'd need to adjust the failing tests yourself.

samdoran:
Now that #815 is merged (thanks @Shrews!), the remaining test failures are due to changes in this PR and I am able to reproduce them locally with `tox -e py38 test/unit/ -- -n2 -vv`.

In all cases, an expected mount is missing from the generated commands.
```
>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> PDB post_mortem (IO-capturing turned off) >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
> /Users/sdoran/Source/ansible-runner/test/unit/config/test_doc.py(121)test_prepare_plugin_docs_command_with_containerization()
-> assert expected_command_start == rc.command
(Pdb) set(expected_command_start).difference(set(rc.command))
{'/private/tmp/pytest-of-sdoran/pytest-75/test_prepare_plugin_docs_comma1/artifacts/:/runner/artifacts/:Z'}
```

AlanCoding:
Oh wow, @samdoran is right. @beeankha the new code block of `if self.container_auth_data:` was put into the wrong place, dividing up an existing `if` block. If you look at the diff, the indentations are off. The most appropriate home looks like it would be down before `if self.container_volume_mounts:`

Could have been a rebase gone wrong or something like that.

ansible_runner/config/_base.py
AlanCoding: It is possible via the CLI by defining `container_auth_data` in the `env/settings` file. We have been manually testing this use case.

ansible_runner/config/_base.py
```diff
@@ -516,6 +535,27 @@ def wrap_args_for_containerization(self, args, execution_mode, cmdline_args):
...
 
+    def _generate_container_auth_file(self, host, username, password):
+        token = "{}:{}".format(username, password)
+        encoded_container_auth_data = {'auths': {host: {'auth': b64encode(token.encode('UTF-8')).decode('UTF-8')}}}
+        # Create a new temp file with container auth data
+        path = tempfile.mkdtemp(prefix='ansible_runner_registry_%s_' % self.ident)
+
+        @atexit.register
+        def cleanup_ansible_runner_registry():
+            shutil.rmtree(path)
```
AlanCoding, line 547: Since my PR has merged and you rebased, you could replace this with `register_for_cleanup(path)`, which is a pattern in a few other places now. But you don't have to.

beeankha:
regate

samdoran:
recheck

PR #818, opened 2021-09-09
Cleanup test file

AlanCoding:


Shrews (changes requested):
If we're going to fix such things, we should do it right and use the `tmp_path` [pytest fixture](https://docs.pytest.org/en/latest/how-to/tmp_path.html#the-tmp-path-fixture). pytest should handle the cleanup for us.

AlanCoding:
@Shrews there is no precedent for `tmp_path` in the code base here, but there is `tmpdir.mkdir`. I'd be happy to switch to that.

https://github.com/ansible/ansible-runner/blob/cdc0961df51fa1e10b44371944aafe5ae140b98c/test/integration/test_transmit_worker_process.py#L58

Shrews:
> @Shrews there is no precedent for `tmp_path` in the code base here, but there is `tmpdir.mkdir`. I'd be happy to switch to that.
> 
> https://github.com/ansible/ansible-runner/blob/cdc0961df51fa1e10b44371944aafe5ae140b98c/test/integration/test_transmit_worker_process.py#L58

`tmpdir` is the legacy way to do things (it's recommended to switch to `tmp_path`), but as long as we're making use of pytest facilities for this type of thing, I'm fine with that choice, if you prefer.

samdoran:
This is something else I was cleaning up in #815, though I should have used `tmp_path` there.

AlanCoding:
Closing in favor of that other PR

PR #819, opened 2021-09-09
Try to make another test robust to run on other OSes

AlanCoding:
This test consistently fails on my Fedora machine, and I have figured out that it's due to these:

https://github.com/ansible/ansible-runner/blob/205c4940e44ae0cdbe418c2b4bcf3d2f312388eb/ansible_runner/config/_base.py#L547

It might be better to adjust this to mock the environment variables before getting into the meat of things. I might still try that.

### DIFF
--- a/test/unit/config/test_inventory.py
+++ b/test/unit/config/test_inventory.py
@@ -2,6 +2,7 @@
 
 import os
 import pytest
+from unittest import mock
 
 from tempfile import gettempdir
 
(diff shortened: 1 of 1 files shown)


samdoran:
I had a PR trying to address this. #815

AlanCoding:
I like some of the patterns you started there, but it doesn't have all the fixes this does. I'll try to copy your `os.environ` bit and see how that looks.

AlanCoding:
@samdoran have a look at the updated diff here. Feel free to close this and incorporate some of these things back into your work.

I certainly don't think we should abandon what you were trying to do. I get that we need to consistently apply this pattern in a few tests. If CI comes back good here, I could try to repeat for all the other tests your PR did.

samdoran:
I believe I incorporated the changes here into https://github.com/ansible/ansible-runner/pull/815/commits/a37d474820c04f7c5917290d0d7ac17d6650b8df.

AlanCoding:
confirmed. Approved @samdoran's PR

PR #820, opened 2021-09-10
Delete broken --keep-files, add new flag just for worker command

AlanCoding:
Connect https://github.com/ansible/awx/issues/10687

This would fulfill our requirement of deleting the private data dir after the worker command finishes. Background: the worker command streams and unzips it into the target dir (either specified or a new temporary file). So deleting the directory is just deleting what the command, itself, wrote.

I'm still debating a few specific implementation details, but this has developed enough that any questions would be helpful. I may be asking myself the same questions...

### DIFF
--- a/ansible_runner/streaming.py
+++ b/ansible_runner/streaming.py
@@ -15,6 +15,7 @@ import ansible_runner
 from ansible_runner.exceptions import ConfigurationError
 from ansible_runner.loader import ArtifactLoader
 import ansible_runner.plugins
+from ansible_runner.utils import register_for_cleanup
 from ansible_runner.utils.streaming import stream_dir, unstream_dir
 
 
@@ -74,6 +75,7 @@ class Worker(object):
         private_data_dir = kwargs.get('private_data_dir')
         if private_data_dir is None:
             private_data_dir = tempfile.mkdtemp()
+            register_for_cleanup(private_data_dir)
         self.private_data_dir = private_data_dir
 
         self.status = "unstarted"
--- a/ansible_runner/utils/__init__.py
+++ b/ansible_runner/utils/__init__.py
@@ -14,6 +14,7 @@ import threading
 import pipes
 import uuid
 import codecs
+import atexit
 
 from distutils.spawn import find_executable
 from ansible_runner.exceptions import ConfigurationError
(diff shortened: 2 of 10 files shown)


AlanCoding:
New proposal, after spending a lot of mental energy on the security footprint of this:

Instead of the `--keep-directory` flag, instead introduce an `--overwrite` flag for the worker command.

If `--overwrite` is given in conjunction with `--private-data-dir`, then it will delete any existing folder, so that a new folder will be re-created and populated by the streamed content.

If the worker command _ever_ creates a new private_data_dir _for any reason_ then it will delete it at the end of the run. The only time a private_data_dir will _not_ be cleaned up is if there is an existing folder and the `--overwrite` flag is not used.

AlanCoding:
Working on https://gist.github.com/AlanCoding/99de3b85031cae3b7f2c7c8f81560359 to explain this, but still haven't wrapped everything up.

ansible_runner/utils/__init__.py
```diff
@@ -26,6 +27,21 @@
 from six import string_types, PY2, PY3, text_type, binary_type
 
 
+def _cleanup_folder(folder):
```
jladdjr, line 30: Really like seeing this as a util, especially with the error handing 👍

AlanCoding:
I've come around to the reality that using the fixture `cli` is the only viable way to test this. If doing that, then the test become nearly exactly the cases outlined in the gist I linked above. Otherwise, you would do some awkward thing of swapping `sys.stdin` with some of the copypasta we have for in-memory buffers.

AlanCoding:
That test was not as bad as I had feared.

Now, we have a new / amended flag with help text :heavy_check_mark: mention contextually in docs :heavy_check_mark:, and add tests :heavy_check_mark: 

I'm going to mark this ready for review, but if you're really interested, please do read the gist I posted and let me know if any of that is something you'd like to see in actual docs. I have to table some of this, pending a fix for https://github.com/ansible/ansible-runner/issues/821, because no point in documenting something that's wrong.

ansible_runner/streaming.py
```diff
@@ -74,6 +75,7 @@ def __init__(self, _input=None, _output=None, **kwargs):
         private_data_dir = kwargs.get('private_data_dir')
         if private_data_dir is None:
             private_data_dir = tempfile.mkdtemp()
+            register_for_cleanup(private_data_dir)
```
jbradberry, line 78: :+1:

jbradberry (approved):
This seems reasonable to me.

jladdjr (approved):

Shrews (approved):

PR #333, opened 2019-08-19
return user error in some situations from cli

privateip:
This commit will check if the provided command line options and return
a more user friendly error if an invalid combination of options is
specified at run time.  This PR addresses the comments in #325

Signed-off-by: Peter Sprygada <psprygad@redhat.com>

PR #825, opened 2021-09-14
Do not rely on length of sys.argv

samdoran:
From the current behavior, it seems the goal is to print a common usage message and exit 0 if no sub command was given. Under other circumstances, such as an invalid or missing parameter, exit with 2 and print the normal usage message.

In order to preserve this behavior and make the tests run consistently whether or not they are run in multiple process, a custom error method was created that inspects the error message and exits 0 with the common usage message if no command was given. This is a bit fragile, but inside that method we do not have easy access to arguments passed to the main() function (they could be inferred by inspecting private attributes, but this is probably unwise).

This requires catching the SystemExit in the main() function and deciding there based on sys_args how to handle it.

This fix preservers the existing behavior whil making the tests pass consistently. However, it would be much simpler to rely on the default behavior of argparse and exit 2 in both cases. This would remove the need for a custom error method and exception handling. It would be a change in behavior, though.

Related to #815.

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -414,11 +414,9 @@ logger = logging.getLogger('ansible-runner')
 
 class AnsibleRunnerArgumentParser(argparse.ArgumentParser):
     def error(self, message):
-        # If no sub command was provided, print common usage and exit cleanly.
-        if 'required: command' in message:
-            self.print_usage()
+        # If no sub command was provided, print common usage then exit
+        if 'required: command' in message.lower():
             print_common_usage()
-            self.exit(0, message)
 
         super(AnsibleRunnerArgumentParser, self).error(message)
 
@@ -755,15 +753,7 @@ def main(sys_args=None):
     add_args_to_parser(isalive_container_group, DEFAULT_CLI_ARGS['container_group'])
     add_args_to_parser(transmit_container_group, DEFAULT_CLI_ARGS['container_group'])
 
-    try:
-        args = parser.parse_args(sys_args)
-    except SystemExit:
-        # If sys_args is None, assume this is interactive and use the behavior in the custom error() method.
-        if sys_args is None:
-            raise
-
-        # Otherwise, assume a call to main() passed in invalid arguments
-        parser.exit(2)
+    args = parser.parse_args(sys_args)
 
     vargs = vars(args)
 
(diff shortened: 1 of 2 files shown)


samdoran:
I added tests to ensure 100% coverage of the new code.

ansible_runner/__main__.py
```diff
@@ -754,12 +765,15 @@ def main(sys_args=None):
...
+    try:
+        args = parser.parse_args(sys_args)
+    except SystemExit:
+        # If sys_args is None, assume this is interactive and use the behavior in the custom error() method.
+        if sys_args is None:
+            raise
 
-    args = parser.parse_args(sys_args)
+        # Otherwise, assume a call to main() passed in invalid arguments
+        parser.exit(2)
```
AlanCoding: wouldn't `raise` work in this case too? Invalid arguments are already processed correctly as it stands, which works as a `SystemExit` exception uncaught here from `parser.parse_args(sys_args)`

AlanCoding (approved):

ansible_runner/__main__.py
samdoran: Yeah, this whole thing is a bit convoluted. I don't love it.

The reason `raise` won't work is that the custom `error()` method changes the exit code. And, at least my from understanding of the existing code and tests, it should raise 2 when no args are passed in the Python interface but 0 when no args are passed in the CLI.

I'm really hoping someone will tell me "Oops we didn't mean to do that. It's fine to just exit with 2 in all cases."

```
> ansible-runner > /dev/null 2>&1 ; echo $status
0
> python -c 'from ansible_runner.__main__ import main; main([])' > /dev/null 2>&1 ; echo $status
2
```

If I change it to raise, the Python interface behavior changes.

```
> python -c 'from ansible_runner.__main__ import main; main([])' > /dev/null 2>&1 ; echo $status
0
```

AlanCoding: The 0 exit code looks like it was introduced in #333

I offer no defense, whatsoever, to giving an exit code of 0 for `ansible-runner` by itself. A code of 2 sounds _more_ correct. I care about that less than I care about making this code clean and maintainable.

AlanCoding: @matburt do you have any reason to keep the exit codes as they are?

samdoran: I got same feedback from @shanemcd and it seems we're ok with exiting 2 in all cases. I will work on simplifying this PR in a bit.

samdoran: The latest changes are _much_ cleaner.

Shrews (approved):

Shrews:
recheck

PR #829, opened 2021-09-15
Support verify_ssl=false for container auth

AlanCoding:
This is needed for Automation Hub where we can't install trusted certs for testing.

This is absorbing AWX code from https://github.com/ansible/awx/pull/9951

### DIFF
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -502,9 +502,9 @@ class BaseConfig(object):
                 new_args.insert(docker_idx + 1, "--config={}".format(self.registry_auth_path))
             if registry_auth_conf_file is not None:
                 # Podman >= 3.1.0
-                self.env['CONTAINERS_REGISTRIES_CONF'] = registry_auth_conf_file
+                os.environ['CONTAINERS_REGISTRIES_CONF'] = registry_auth_conf_file
                 # Podman < 3.1.0
-                self.env['REGISTRIES_CONFIG_PATH'] = registry_auth_conf_file
+                os.environ['REGISTRIES_CONFIG_PATH'] = registry_auth_conf_file
 
         if self.container_volume_mounts:
             for mapping in self.container_volume_mounts:
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -502,9 +502,9 @@ class BaseConfig(object):
                 new_args.insert(docker_idx + 1, "--config={}".format(self.registry_auth_path))
             if registry_auth_conf_file is not None:
                 # Podman >= 3.1.0
-                os.environ['CONTAINERS_REGISTRIES_CONF'] = registry_auth_conf_file
+                self.env['CONTAINERS_REGISTRIES_CONF'] = registry_auth_conf_file
                 # Podman < 3.1.0
-                os.environ['REGISTRIES_CONFIG_PATH'] = registry_auth_conf_file
+                self.env['REGISTRIES_CONFIG_PATH'] = registry_auth_conf_file
 
         if self.container_volume_mounts:
             for mapping in self.container_volume_mounts:
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -42,7 +42,7 @@ from ansible_runner.utils import (
     args2cmdline,
     sanitize_container_name,
     cli_mounts,
-    register_for_cleanup
+    register_for_cleanup,
 )
 
 logger = logging.getLogger('ansible-runner')
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -16,13 +16,11 @@
 # specific language governing permissions and limitations
 # under the License.
 #
-import atexit
 import json
 import logging
 import os
 import pexpect
 import re
-import shutil
 import stat
 import tempfile
 
@@ -43,7 +41,8 @@ from ansible_runner.utils import (
     open_fifo_write,
     args2cmdline,
     sanitize_container_name,
-    cli_mounts
+    cli_mounts,
+    register_for_cleanup
 )
 
 logger = logging.getLogger('ansible-runner')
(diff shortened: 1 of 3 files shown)


ansible_runner/config/_base.py
```diff
@@ -495,15 +494,17 @@ def wrap_args_for_containerization(self, args, execution_mode, cmdline_args):
...
             if 'podman' in self.process_isolation_executable:
                 new_args.extend(["--authfile={}".format(self.registry_auth_path)])
             else:
                 docker_idx = new_args.index(self.process_isolation_executable)
                 new_args.insert(docker_idx + 1, "--config={}".format(self.registry_auth_path))
+            if registry_auth_conf_file is not None:
+                # Podman >= 3.1.0
+                self.env['CONTAINERS_REGISTRIES_CONF'] = registry_auth_conf_file
+                # Podman < 3.1.0
+                self.env['REGISTRIES_CONFIG_PATH'] = registry_auth_conf_file
```
AlanCoding, line 507: @shanemcd I am struggling with a decision here.

It would be more correct to set this in `os.environ` rather than `self.env`. It make no sense to expose this to the runtime in the container, but the file won't be viewable in the container.

I think `os.environ` would cause hesitations because it's expected that people use the python interface, and we have no clean way of reverting that. What we really need is to set it directly on the `pexepect_env` here:

https://github.com/ansible/ansible-runner/blob/cdc0961df51fa1e10b44371944aafe5ae140b98c/ansible_runner/runner.py#L178

But that would involve adding some other property on the RunnerConfig object, and there's already a lot. So because this isn't a regression, I'd prefer to be conservative and leave it like this (behaving the same it did before)

AlanCoding:
I have now tested this against a real Automation Hub instance.

 - I downloaded the awx-ee image
 - I made a new tag with the AH URL as the base from that image
 - pushed the image to AH with superuser account
 - added `demo/env/settings` file with all the basic containerization & auth parameters
 - `ansible-runner run demo -p test.yml`

#### without setting verify_ssl=false

This shows the certificate failure (from pulp registry), as it hasn't been set up in this case.

```
$ ansible-runner run demo -p test.yml
Error: Error initializing source docker://ec2-54-174-150-99.compute-1.amazonaws.com/alan/awx-ee:latest: error pinging docker registry ec2-54-174-150-99.compute-1.amazonaws.com: Get "https://ec2-54-174-150-99.compute-1.amazonaws.com/v2/": x509: certificate signed by unknown authority
```

#### with setting verify_ssl=false

it runs the test.yml playbook

ansible_runner/config/_base.py
shanemcd: I haven't had much time to let this thought marinade, so it might be dumb, but - since `container_auth_data` is a dictionary, it doesn't seem to crazy to shove another key under there and write the value to a file. Or something like that. WDYT?

AlanCoding: I don't really understand, I will try to catch up tomorrow.

shanemcd (approved):

ansible_runner/config/_base.py
shanemcd: We spoke in Slack, what I proposed here is what they are already doing.

beeankha (approved):

shanemcd:
Once this lands in RHEL maybe we can refactor to be a bit more elegant https://github.com/containers/podman/pull/11205

ansible_runner/config/_base.py
```diff
@@ -43,7 +41,8 @@
     open_fifo_write,
     args2cmdline,
     sanitize_container_name,
-    cli_mounts
+    cli_mounts,
+    register_for_cleanup
```
samdoran: To keep diffs cleaner in the future.
```suggestion
    register_for_cleanup,
```

ansible_runner/config/_base.py
```diff
@@ -536,15 +537,14 @@ def wrap_args_for_containerization(self, args, execution_mode, cmdline_args):
         logger.debug(f"container engine invocation: {' '.join(new_args)}")
         return new_args
 
-    def _generate_container_auth_file(self, host, username, password):
-        token = "{}:{}".format(username, password)
+    def _generate_container_auth_dir(self, auth_data):
+        host = auth_data.get('host')
+        token = "{}:{}".format(auth_data.get('username'), auth_data.get('password'))
```
samdoran, line 542: What happens if either of these values end up being `None`? Will that fail later is a reasonable manner?

test/integration/containerized/test_container_management.py
```diff
@@ -94,26 +94,37 @@ def test_invalid_registry_host(tmp_path, runtime):
...
     result_stdout = res.stdout.read()
     if runtime == 'podman':
         assert image_name in result_stdout
         assert 'unauthorized' in result_stdout
+        auth_file_path = res.config.registry_auth_path
+        registry_conf = os.path.join(os.path.dirname(res.config.registry_auth_path), 'registries.conf')
     else:
         assert 'access to the requested resource is not authorized' in result_stdout
+        auth_file_path = os.path.join(res.config.registry_auth_path, 'config.json')
+        registry_conf = os.path.join(res.config.registry_auth_path, 'registries.conf')
```
samdoran: The `else` isn't strictly necessary.
```suggestion
    auth_file_path = os.path.join(res.config.registry_auth_path, 'config.json')
    registry_conf = os.path.join(res.config.registry_auth_path, 'registries.conf')
    error_msg = 'access to the requested resource is not authorized'
    if runtime == 'podman':
        assert image_name in result_stdout
        error_msg = 'unauthorized'
        auth_file_path = res.config.registry_auth_path
        registry_conf = os.path.join(os.path.dirname(res.config.registry_auth_path), 'registries.conf')
        
    assert error_msg in result_stdout

```

samdoran (approved):

ansible_runner/config/_base.py
AlanCoding: No, it will not.

To state what the error handling would be - host, username, password are all required, provided that there is something in auth_data at all. The verify_ssl is optional, and should be `True` or `False` if it is. We are not validating it now, but if we were, these would be the rules.

test/integration/containerized/test_container_management.py
```diff
@@ -94,26 +94,39 @@ def test_invalid_registry_host(tmp_path, runtime):
...
+    error_msg = 'access to the requested resource is not authorized'
     if runtime == 'podman':
         assert image_name in result_stdout
-        assert 'unauthorized' in result_stdout
-    else:
-        assert 'access to the requested resource is not authorized' in result_stdout
+        error_msg = 'unauthorized'
+        auth_file_path = res.config.registry_auth_path
+        registry_conf = os.path.join(os.path.dirname(res.config.registry_auth_path), 'registries.conf')
+
```
samdoran:
```suggestion

```

test/integration/containerized/test_container_management.py
```diff
@@ -94,26 +94,39 @@ def test_invalid_registry_host(tmp_path, runtime):
...
+        error_msg = 'unauthorized'
+        auth_file_path = res.config.registry_auth_path
+        registry_conf = os.path.join(os.path.dirname(res.config.registry_auth_path), 'registries.conf')
+        
+    assert error_msg in result_stdout
 
-    assert os.path.exists(res.config.registry_auth_path)
-    auth_path = res.config.registry_auth_path if runtime == 'podman' else os.path.join(res.config.registry_auth_path, 'config.json')
-    with open(auth_path, 'r') as f:
+
```
samdoran:
```suggestion
```

samdoran:
recheck

PR #831, opened 2021-09-16
Use mocker fixture

samdoran:
Using the `mocker` fixture rather than `unittest.Mock` directly has a number of benefits. The tests read easier since context managers within tests are unnecessary and patches are automatically scoped to the test where the fixture is used.

Also overhaul the tests in `test___main__.py` to test the code as intended.

### DIFF
--- a/test/integration/test___main__.py
+++ b/test/integration/test___main__.py
@@ -4,12 +4,10 @@ import uuid
 import json
 import random
 import string
-import tempfile
-import shutil
 
 import pytest
 
-from ansible_runner.__main__ import main
+import ansible_runner.__main__
 
 
 def random_string():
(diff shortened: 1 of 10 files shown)


test/integration/test___main__.py
```diff
@@ -42,30 +41,26 @@ def test_main_bad_private_data_dir():
...
     finally:
         os.remove(tmpfile)
 
 
-def run_role(options, private_data_dir=None, expected_rc=0):
-    try:
-        private_data_dir = private_data_dir or tempfile.mkdtemp()
-        args = ['run', private_data_dir]
-        args.extend(options)
+def run_role(options, mocker, private_data_dir, expected_rc=0):
```
AlanCoding: You could turn `run_role` into a pytest fixture, to avoid passing `mocker` (a pytest fixture) as an argument to it.

test/integration/test___main__.py
```diff
@@ -42,30 +41,26 @@ def test_main_bad_private_data_dir():
...
+    args = ['run', private_data_dir]
+    args.extend(options)
+    mock_run = mocker.patch('ansible_runner.interface.run')
```
AlanCoding: It still feels like this should be `with mocker.patch`, because the feature of being test-scoped is now more like a bug, where anything running after `run_role` is called will (rather unexpectedly) have that patch active.

test/integration/test___main__.py
```diff
@@ -214,7 +209,7 @@ def test_cmdline_includes_one_option():
     """Test error is raised if not '-p', '-m' or '-r'
     """
     cmdline('run', 'private_data_dir')
-    with raises(SystemExit) as exc:
+    with pytest.raises(SystemExit) as exc:
```
AlanCoding, line 140: :+1: :100:

test/integration/test___main__.py
samdoran: I tried that and the issue is we need to pass args to it. I _could_ make it a fixture by putting a function within the fixture, but decided to just pass in `mocker`. I don't care for local functions, which is what you have to do to make fixtures accept args.

test/integration/test___main__.py
samdoran: Good point.

test/integration/test___main__.py
```diff
@@ -42,30 +41,26 @@ def test_main_bad_private_data_dir():
...
     try:
-        with raises(OSError):
+        with pytest.raises(OSError):
             main()
     finally:
         os.remove(tmpfile)
 
 
-def run_role(options, private_data_dir=None, expected_rc=0):
-    try:
```
samdoran, line 52: Removing this outer `try` revealed test failures that were being swallowed by the `finally` statement. It's related to #825. Can't get rid of that code fast enough.

test/integration/test___main__.py
samdoran: Wow. I also just realized the `args` here never gets passed to `main()`. 🤦‍♂️

AlanCoding: It is a very common pytest practice to define a function locally and return it (in a fixture). If you want to just define the method outside of the pytest fixture structure, then that's a perfectly fine preference to have. But in that case, using the mock from `from unittest import mock` makes more sense.

Directly defining the method in the test file but expecting a pytest `mocker` fixture to be passed is mixing the 2 worlds.

samdoran: > Directly defining the method in the test file but expecting a pytest mocker fixture to be passed is mixing the 2 worlds.

Yeah, I'm coming around to this. I'll rewrite it as a fixture.

samdoran: Getting this to accomplish the original intent of the test took quite a bit more work than changing this to a fixture. I ended up deleting this entirely.

AlanCoding (approved):

Shrews (approved):

PR #837, opened 2021-09-20
Fix Ansible version comparison in tests

Shrews:
Current Ansible version string comparison for the `test_display_callback.py::test_callback_plugin_records_notify_events` test is incorrect as `2.5` does not compare as less than `2.11`, etc., so the test is being skipped.

### DIFF
--- a/test/integration/test_display_callback.py
+++ b/test/integration/test_display_callback.py
@@ -5,8 +5,6 @@ import os
 import yaml
 import six
 
-from ansible import __version__ as ANSIBLE_VERSION
-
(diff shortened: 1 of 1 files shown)


samdoran (approved):

samdoran:
Can you remove the unused import of `ANSIBLE_VERSION`?

Shrews:
> Can you remove the unused import of `ANSIBLE_VERSION`?

Done in the previous commit.

samdoran (approved):

PR #838, opened 2021-09-20
Fix test consistency in integration/test_main.py

Shrews:
Fixes (hopefully) some random failures seen from `integration/test_main.py`.

- Stop sharing `test/integration` as private_data_dir across tests. I believe this to be the source of random failures.
- Specify `--artifacts-dir` for tests that _do_ use `test/integration` as private_data_dir so the data doesn't overlap.
- Consistently use `tmp_path` pytest fixture across all tests.
- Removes some redundant tests caused by the above changes (all-the-things are still appropriately tested).
- Removes a redundant test for logfile that was being problematic with another logfile test because of a (suspected) shared Python `logging` environment.

### DIFF
--- a/test/integration/test_main.py
+++ b/test/integration/test_main.py
@@ -102,17 +102,14 @@ def test_help(command, expected, capsys, monkeypatch):
     assert expected['err'] in stderr
 
 
-def test_module_run():
-    try:
-        rc = main(['run', '-m', 'ping',
-                   '--hosts', 'localhost',
-                   'ping'])
-        assert os.path.exists('./ping')
-        assert os.path.exists('./ping/artifacts')
-        assert rc == 0
-    finally:
-        if os.path.exists('./ping'):
-            shutil.rmtree('./ping')
+def test_module_run(tmp_path):
+    pvt_data_dir = tmp_path / 'ping'
+    rc = main(['run', '-m', 'ping',
+               '--hosts', 'localhost',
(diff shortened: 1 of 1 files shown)


test/integration/test_main.py
```diff
@@ -134,48 +131,41 @@ def test_module_run_clean(tmp_path):
     assert rc == 0
 
 
-def test_role_run(skipif_pre_ansible28, clear_integration_artifacts):
+def test_role_run(skipif_pre_ansible28, tmp_path):
+    artifact_dir = tmp_path / "otherartifacts"
     rc = main(['run', '-r', 'benthomasson.hello_role',
                '--hosts', 'localhost',
                '--roles-path', 'test/integration/roles',
```
AlanCoding, line 138: FYI, the `--roles-path` argument here is doing nothing. It's not the right path, because the role is in `test/integration/project/roles`. It's getting the role because it's in the `roles` directory relative to the project (where the playbook runs), so it doesn't need to be set by the Ansible config.

I think my number 1 preference would be do make a tmp dir and `shutil.copytree` the project folder to that.

Shrews: Yep, I discovered that as well. It's because the temporary playbook is created in the project directory, and the adjacent roles are found there. I'll probably fix that in a different PR.

AlanCoding (approved):

samdoran (approved):

PR #839, opened 2021-09-20
Update key names in setup.cfg

samdoran:
Use underscore separated names since dash separated names are deprecated.


Shrews (approved):

samdoran:
recheck

PR #846, opened 2021-09-21
Introduce `ansible-runner worker cleanup` as cleanup of last-resort for remote execution nodes

AlanCoding:
As of creating the draft PR, I might have all the test cases I brainstormed taken care of. I had some content for manually running demos:

https://gist.github.com/AlanCoding/99de3b85031cae3b7f2c7c8f81560359#file-cleanup_demo-md

I don't want to merge this until AWX changes have been finished (in basic terms) and tested locally. I do want to make sure runner CI is running and that I don't get any surprises.

----

EDIT: I am now pretty happy with the testing in conjunction with AWX.

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -39,6 +39,7 @@ from yaml import safe_dump, safe_load
 
 from ansible_runner import run
 from ansible_runner import output
+from ansible_runner import cleanup
 from ansible_runner.utils import dump_artifact, Bunch, register_for_cleanup
 from ansible_runner.utils.capacity import get_cpu_count, get_mem_in_bytes, get_uuid
 from ansible_runner.runner import Runner
@@ -621,41 +622,14 @@ def main(sys_args=None):
     )
     worker_subcommands = worker_subparser.add_subparsers(
         help="Sub-sub command to invoke",
-        dest='command',
+        dest='worker_subcommand',
         description="ansible-runner worker [sub-sub-command]"
     )
     cleanup_command = worker_subcommands.add_parser(
         'cleanup',
         help="Cleanup private_data_dir patterns from prior jobs and supporting temporary folders."
     )
-    cleanup_command.add_argument(
-        "--file-pattern",
-        help="A file glob to find private_data_dir folders to remove. "
-             "You may use {ident} in place of a ** to identify run IDs. "
-             "Example: --file-pattern=/tmp/foo_{ident}_**"
-    )
-    cleanup_command.add_argument(
-        "--exclude-idents",
-        help="A comma separated list of run IDs to preserve. "
-             "This will only work if the deletion pattern contains the {ident} syntax."
-    )
-    cleanup_command.add_argument(
-        "--untag-images",
-        help="A comma separated list of podman or docker tags to delete. "
-             "This will not remove the corresponding layers, use the image-prune option for that. "
-             "Example: --untag-images=quay.io/user/image:devel,quay.io/user/builder:latest"
-    )
-    cleanup_command.add_argument(
-        "--image-prune",
-        action="store_true",
-        help="If specified, will run docker / podman image prune --force. "
-             "This will only run after untagging."
-    )
-    cleanup_command.add_argument(
-        "--process-isolation-executable",
-        default="podman",
-        help="The container image to clean up images for (default=podman)"
-    )
+    cleanup.add_cleanup_args(cleanup_command)
 
     worker_subparser.add_argument(
         "--private-data-dir",
@@ -796,6 +770,9 @@ def main(sys_args=None):
     vargs = vars(args)
 
     if vargs.get('command') == 'worker':
+        if vargs.get('worker_subcommand') == 'cleanup':
+            cleanup.run_cleanup(vargs)
+            parser.exit(0)
         if vargs.get('worker_info'):
             cpu = get_cpu_count()
             mem = get_mem_in_bytes()
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -623,11 +623,11 @@ def main(sys_args=None):
     worker_subcommands = worker_subparser.add_subparsers(
         help="Sub-sub command to invoke",
         dest='worker_subcommand',
-        description="ansible-runner worker [sub-sub-command]"
+        description="ansible-runner worker [sub-sub-command]",
     )
     cleanup_command = worker_subcommands.add_parser(
         'cleanup',
-        help="Cleanup private_data_dir patterns from prior jobs and supporting temporary folders."
+        help="Cleanup private_data_dir patterns from prior jobs and supporting temporary folders.",
     )
     cleanup.add_cleanup_args(cleanup_command)
 
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -619,6 +619,44 @@ def main(sys_args=None):
         'worker',
         help="Execute work streamed from a controlling instance"
     )
+    worker_subcommands = worker_subparser.add_subparsers(
+        help="Sub-sub command to invoke",
+        dest='command',
+        description="ansible-runner worker [sub-sub-command]"
+    )
+    cleanup_command = worker_subcommands.add_parser(
+        'cleanup',
+        help="Cleanup private_data_dir patterns from prior jobs and supporting temporary folders."
+    )
+    cleanup_command.add_argument(
+        "--file-pattern",
+        help="A file glob to find private_data_dir folders to remove. "
+             "You may use {ident} in place of a ** to identify run IDs. "
+             "Example: --file-pattern=/tmp/foo_{ident}_**"
+    )
+    cleanup_command.add_argument(
+        "--exclude-idents",
+        help="A comma separated list of run IDs to preserve. "
+             "This will only work if the deletion pattern contains the {ident} syntax."
+    )
+    cleanup_command.add_argument(
+        "--untag-images",
+        help="A comma separated list of podman or docker tags to delete. "
+             "This will not remove the corresponding layers, use the image-prune option for that. "
+             "Example: --untag-images=quay.io/user/image:devel,quay.io/user/builder:latest"
+    )
+    cleanup_command.add_argument(
+        "--image-prune",
+        action="store_true",
+        help="If specified, will run docker / podman image prune --force. "
+             "This will only run after untagging."
+    )
+    cleanup_command.add_argument(
+        "--process-isolation-executable",
+        default="podman",
+        help="The container image to clean up images for (default=podman)"
+    )
+
     worker_subparser.add_argument(
         "--private-data-dir",
         help="base directory containing the ansible-runner metadata "
--- a/ansible_runner/cleanup.py
+++ b/ansible_runner/cleanup.py
@@ -2,12 +2,16 @@ import glob
 import subprocess
 import os
 import signal
+import datetime
 
 from ansible_runner.defaults import registry_auth_prefix
 from ansible_runner.utils import cleanup_folder
 
 
 __all__ = ['add_cleanup_args', 'run_cleanup']
+
+
+GRACE_PERIOD_DEFAULT = 60  # minutes
 
 
 def add_cleanup_args(command):
@@ -25,6 +29,14 @@ def add_cleanup_args(command):
         help="A comma separated list of podman or docker tags to delete. "
              "This may not remove the corresponding layers, use the image-prune option to assure full deletion. "
              "Example: --remove-images=quay.io/user/image:devel,quay.io/user/builder:latest"
+    )
+    command.add_argument(
+        "--grace-period",
+        default=GRACE_PERIOD_DEFAULT,
+        type=int,
+        help="Time (in minutes) after last modification to exclude a folder from deletion for. "
+             "This is to avoid deleting folders that were recently created, or folders not started via the start command. "
+             "Value of 0 indicates that no folders will be excluded based on modified time."
     )
     command.add_argument(
         "--image-prune",
@@ -86,11 +98,17 @@ def delete_associated_folders(dir):
                 print(f'Removed associated registry auth dir {dir}')
 
 
-def cleanup_dirs(pattern, exclude_strings=()):
+def cleanup_dirs(pattern, exclude_strings=(), grace_period=GRACE_PERIOD_DEFAULT):
     ct = 0
+    now_time = datetime.datetime.now()
     for dir in glob.glob(pattern):
-        if any(str(ident) in dir for ident in exclude_strings):
+        if any(str(exclude_string) in dir for exclude_string in exclude_strings):
             continue
+        if grace_period:
+            st = os.stat(dir)
+            modtime = datetime.datetime.fromtimestamp(st.st_mtime)
+            if modtime > now_time - datetime.timedelta(minutes=grace_period):
+                continue
         if is_alive(dir):
             print(f'Excluding running project {dir} from cleanup')
             continue
@@ -138,7 +156,7 @@ def run_cleanup(vargs):
     pruned = False
 
     if file_pattern:
-        dir_ct = cleanup_dirs(file_pattern, exclude_strings=exclude_strings)
+        dir_ct = cleanup_dirs(file_pattern, exclude_strings=exclude_strings, grace_period=vargs.get('grace_period'))
         if dir_ct:
             print(f'Removed {dir_ct} private data dir(s) in pattern {file_pattern}')
 
--- a/ansible_runner/cleanup.py
+++ b/ansible_runner/cleanup.py
@@ -50,9 +50,6 @@ def run_command(cmd):
         print('Stdout:')
         print(stdout)
         raise RuntimeError('Error running command')
-    print('Successfully ran command: {}'.format(' '.join(cmd)))
-    print(stdout)
-    # import pdb; pdb.set_trace()
     return stdout.strip()
 
 
@@ -73,23 +70,22 @@ def is_alive(dir):
 
 
 def cleanup_dirs(pattern, exclude_idents=()):
-    discovered_dirs = glob.glob(pattern)
     ct = 0
     running_idents = []
-    for dir in discovered_dirs:
+    for dir in glob.iglob(pattern):
         if any(ident in dir for ident in exclude_idents):
             continue
         if is_alive(dir):
             running_idents.extend(os.path.listdir(os.path.join(dir, 'artifacts')))
             continue
         shutil.rmtree(dir)
-        print(f'Removed directory {dir}')
         ct += 1
+    if ct:
+        print(f'Removed {ct} private data dir(s) in pattern {pattern}')
     if running_idents:
         print(f'Excluding from cleanup running jobs {running_idents}')
     registry_auth_pattern = f'{registry_auth_prefix}{{ident}}_**'
-    registry_auth_dirs = glob.glob(registry_auth_pattern)
-    for dir in registry_auth_dirs:
+    for dir in glob.iglob(registry_auth_pattern):
         if any(ident in dir for ident in exclude_idents) or any(ident in dir for ident in running_idents):
             continue
         shutil.rmtree(dir)
@@ -107,6 +103,8 @@ def cleanup_images(images, runtime='podman'):
         for discovered_tag in stdout.split('\n'):
             stdout = run_command([runtime, 'rmi', discovered_tag.strip().strip('"'), '-f'])
             rm_ct += stdout.count('Untagged:')
+    if rm_ct:
+        print(f'Removed {rm_ct} image(s)')
     return rm_ct
 
 
--- a/ansible_runner/cleanup.py
+++ b/ansible_runner/cleanup.py
@@ -22,10 +22,12 @@ def add_cleanup_args(command):
     )
     command.add_argument(
         "--exclude-strings",
-        help="A comma separated list of keywords in directory to avoid deleting."
+        nargs='*',
+        help="A comma separated list of keywords in directory name or path to avoid deleting."
     )
     command.add_argument(
         "--remove-images",
+        nargs='*',
         help="A comma separated list of podman or docker tags to delete. "
              "This may not remove the corresponding layers, use the image-prune option to assure full deletion. "
              "Example: --remove-images=quay.io/user/image:devel,quay.io/user/builder:latest"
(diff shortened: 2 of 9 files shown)


ansible_runner/cleanup.py
```diff
@@ -0,0 +1,172 @@
...
+            continue
+        for discovered_tag in stdout.split('\n'):
+            stdout = run_command([runtime, 'rmi', discovered_tag.strip().strip('"'), '-f'])
+            rm_ct += stdout.count('Untagged:')
+    return rm_ct
+
+
+def prune_images(runtime='podman'):
+    """Run the prune images command and return changed status"""
+    stdout = run_command([runtime, 'image', 'prune', '-f'])
```
AlanCoding, line 160: @fosterseth in AWX we did something a little different, and I'm not totally sure why. It kind of manually deleted dangling images instead of using the prune command.

https://github.com/ansible/awx/blob/7c9626b0e7c0b7835c8ff42970c6f1740f3b10fe/awx/main/tasks.py#L397

This might have been to fix some podman-specific bug, but I lack the knowledge to reproduce that bug. We could add it here using the same commands as you used. Wouldn't hurt.

docs/remote_jobs.rst
```diff
@@ -25,14 +25,31 @@ The `ansible-runner worker` command accepts this stream, runs the playbook, and
...
+------------------------------
+
+When running `ansible-runner worker`, if no `--private-data-dir` is given,
+it will extract the contents to a temporary directory which is deleted at the end of execution.
+You can use the `--delete` flag in conjunction with `--private-data-dir` to assure that
+the provided directory is deleted at the end of execution.
+
+The following command offers out-of-band cleanup.
+
+    $ ansible-runner worker cleanup --file-pattern=/tmp/foo_*
```
Shrews, line 46: Do we need to consider any security (or oopsie) implications of this? What if someone used `--file-pattern=/`, either accidentally or on purpose?

AlanCoding: I agree with some oopsie validation, but I don't know of any good pattern to follow for that. Yeah, throwing an error if given `/` would be a good idea, but I don't have a quality list of places that we should error for.

AlanCoding: I added a little bit of validation. It's not going to prevent everything that could go wrong, but at least it's something.

test/integration/containerized/test_cleanup_images.py
```diff
@@ -0,0 +1,58 @@
...
+    # Create new image just for this test with a unique layer
+    special_string = "Verify this in test - 1QT4r18a7E"
+    dockerfile_path = str(tmp_path / 'Dockerfile')
+    with open(dockerfile_path, 'w') as f:
+        f.write('\n'.join([
+            'FROM {}'.format(default_container_image),
+            'RUN echo {} > /tmp/for_test.txt'.format(special_string)
+        ]))
+    image_name = 'quay.io/fortest/hasfile:latest'
+    build_cmd = [runtime, 'build', '--rm=true', '-t', image_name, '-f', dockerfile_path, str(tmp_path)]
```
samdoran: No need to prematurely convert this to a string.
```suggestion
dockerfile_path = tmp_path / 'Dockerfile'
dockerfile_path.write_text('\n'.join([
    'FROM {}'.format(default_container_image),
    'RUN echo {} > /tmp/for_test.txt'.format(special_string)
]))
image_name = 'quay.io/fortest/hasfile:latest'
build_cmd = [runtime, 'build', '--rm=true', '-t', image_name, '-f', str(dockerfile_path), str(tmp_path)]
```

test/integration/containerized/test_cleanup_images.py
```diff
@@ -0,0 +1,58 @@
...
+    cli(build_cmd, bare=True)
+
+    # get an id for the unique layer
+    r = cli([runtime, 'images', image_name, '--format="{{.ID}}"'], bare=True)
+    layer_id = r.stdout.strip()
+    assert layer_id in cli([runtime, 'images'], bare=True).stdout
+
+    # workaround for https://github.com/ansible/ansible-runner/issues/758
+    os.mkdir(str(tmp_path / 'project'))
```
samdoran:
```suggestion
    tmp_path.joinpath('project').mkdir()
```

test/integration/containerized/test_cleanup_images.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # force no colors so that we can JSON load ad hoc output
+    os.mkdir(str(tmp_path / 'env'))
+    with open(str(tmp_path / 'env' / 'envvars'), 'w') as f:
+        f.write('{"ANSIBLE_NOCOLOR": "true"}')
```
samdoran:
```suggestion
    env_path = tmp_path.joinpath('env')
    env_path.mkdir()
    env_path.joinpath('envvars').write_text('{"ANSIBLE_NOCOLOR": "true"}')
```

test/integration/containerized/test_cleanup_images.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Create new image just for this test with a unique layer
+    special_string = "Verify this in test - 1QT4r18a7E"
+    dockerfile_path = str(tmp_path / 'Dockerfile')
+    with open(dockerfile_path, 'w') as f:
+        f.write('\n'.join([
+            'FROM {}'.format(default_container_image),
+            'RUN echo {} > /tmp/for_test.txt'.format(special_string)
+        ]))
+    image_name = 'quay.io/fortest/hasfile:latest'
```
samdoran: Is this unique enough so that multiple versions of this test running simultaneously on the same host will not encounter errors when the image is later removed?

test/unit/test_cleanup.py
```diff
@@ -0,0 +1,86 @@
+import random
+import os
+import time
```
samdoran: Sort imports alphabetically.
```suggestion
import os
import random
import time
```

test/unit/test_cleanup.py
```diff
@@ -0,0 +1,86 @@
...
+    paths = []
+    for i in range(0, 6, 2):
+        trailing = ''.join(random.choice("abcdefica3829") for i in range(8))
+        path = tmp_path / f'pattern_{i}_{trailing}'
+        path.mkdir()
+        paths.append(str(path))
+
+    a_file_path = os.path.join(tmp_path, 'pattern_32_donotcleanme')
+    with open(a_file_path, 'w') as f:
+        f.write('this is a file and should not be cleaned by the cleanup command')
```
samdoran:
```suggestion
    a_file_path = tmp_path / 'pattern_32_donotcleanme'
    a_file_path.write_text('this is a file and should not be cleaned by the cleanup command')
```

test/unit/test_cleanup.py
```diff
@@ -0,0 +1,86 @@
...
+from ansible_runner.config.runner import RunnerConfig
+
+
+def test_simple_dir_cleanup_with_exclusions(tmp_path):
+    paths = []
+    for i in range(0, 6, 2):
+        trailing = ''.join(random.choice("abcdefica3829") for i in range(8))
+        path = tmp_path / f'pattern_{i}_{trailing}'
+        path.mkdir()
+        paths.append(str(path))
```
samdoran:
```suggestion
        paths.append(path)

```

test/unit/test_cleanup.py
```diff
@@ -0,0 +1,86 @@
...
+        f.write('this is a file and should not be cleaned by the cleanup command')
+
+    keep_dir_path = tmp_path / 'pattern_42_alsokeepme'
+    keep_dir_path.mkdir()
+
+    ct = cleanup_dirs(pattern=str(tmp_path / 'pattern_*_*'), exclude_strings=[42], grace_period=0)
+    assert ct == 3  # cleaned up 3 private_data_dirs
+
+    for path in paths:
+        assert not os.path.exists(path)
```
samdoran:
```suggestion
        assert not path.exists()
```

test/unit/test_cleanup.py
```diff
@@ -0,0 +1,86 @@
...
+
+    assert os.path.exists(a_file_path)
+    assert os.path.exists(str(keep_dir_path))
```
samdoran:
```suggestion
    assert a_file_path.exists()
    assert keep_dir_path.exists()
```

ansible_runner/__main__.py
```diff
@@ -619,6 +620,17 @@ def main(sys_args=None):
         'worker',
         help="Execute work streamed from a controlling instance"
     )
+    worker_subcommands = worker_subparser.add_subparsers(
+        help="Sub-sub command to invoke",
+        dest='worker_subcommand',
+        description="ansible-runner worker [sub-sub-command]"
```
samdoran:
```suggestion
        description="ansible-runner worker [sub-sub-command]",
```

ansible_runner/__main__.py
```diff
@@ -619,6 +620,17 @@ def main(sys_args=None):
...
+    )
+    cleanup_command = worker_subcommands.add_parser(
+        'cleanup',
+        help="Cleanup private_data_dir patterns from prior jobs and supporting temporary folders."
```
samdoran:
```suggestion
        help="Cleanup private_data_dir patterns from prior jobs and supporting temporary folders.",
```

ansible_runner/cleanup.py
```diff
@@ -0,0 +1,192 @@
+import glob
+import subprocess
+import os
+import sys
+import signal
+import datetime
+from pathlib import Path
```
samdoran:
```suggestion
import datetime
import glob
import os
import signal
import subprocess
import sys

from pathlib import Path
```

ansible_runner/cleanup.py
```diff
@@ -0,0 +1,192 @@
...
+             "This will only run after untagging."
+    )
+    command.add_argument(
+        "--process-isolation-executable",
+        default="podman",
+        help="The container image to clean up images for (default=podman)"
+    )
+
+
+def run_command(cmd):
```
samdoran, line 58: We already have several other functions that run commands. Do we really want/need another one?

ansible_runner/cleanup.py
```diff
@@ -0,0 +1,192 @@
...
+
+    try:
+        os.kill(pid, signal.SIG_DFL)
+        return(0)
+    except OSError:
+        return(1)
+
+
+def project_idents(dir):
+    """Give dir, give list of idents that we have artifacts for"""
```
samdoran:
```suggestion
    """Given dir, give list of idents that we have artifacts for"""
```

ansible_runner/cleanup.py
```diff
@@ -0,0 +1,192 @@
...
+    try:
+        return os.listdir(os.path.join(dir, 'artifacts'))
+    except (FileNotFoundError, NotADirectoryError):
+        return []
+
+
+def delete_associated_folders(dir):
+    """Where dir is the private_data_dir for a completed job, this deletes related tmp folders it used"""
+    for ident in project_idents(dir):
+        registry_auth_pattern = f'/tmp/{registry_auth_prefix}{ident}_*'
```
samdoran: Is hardcoding `/tmp` here problematic? What if the system is using a different tmp directory?

ansible_runner/cleanup.py
```diff
@@ -0,0 +1,192 @@
...
+            'Refusing to continue for user system safety.'
+        )
+
+
+def cleanup_dirs(pattern, exclude_strings=(), grace_period=GRACE_PERIOD_DEFAULT):
+    try:
+        validate_pattern(pattern)
+    except RuntimeError as e:
+        print(str(e))
+        sys.exit(1)
```
samdoran: Passing a string to `sys.exit()` exits 1 printing the string to `stderr`. That's slightly different than the current code here but printing to `stderr` is probably more accurate.
```suggestion
        sys.exit(str(e))
```

ansible_runner/cleanup.py
```diff
@@ -0,0 +1,192 @@
...
+from pathlib import Path
+
+from ansible_runner.defaults import registry_auth_prefix
+from ansible_runner.utils import cleanup_folder
+
+
+__all__ = ['add_cleanup_args', 'run_cleanup']
+
+
+GRACE_PERIOD_DEFAULT = 60  # minutes
```
samdoran: Do we have a better place for storing configuration values rather than setting a global in just this file?

ansible_runner/utils/__init__.py
```diff
@@ -27,19 +27,21 @@
 from six import string_types, PY2, PY3, text_type, binary_type
 
 
-def _cleanup_folder(folder):
+def cleanup_folder(folder):
```
samdoran, line 30: This needs tests, especially if it's going to be public.

test/unit/test_cleanup.py
```diff
@@ -0,0 +1,86 @@
...
+    with pytest.raises(RuntimeError) as e:
+        validate_pattern(pattern)
+    assert pattern in str(e)
+    assert 'Provided pattern could result in deleting system folders' in str(e)
+
+
+def test_weird_pattern():
+    with pytest.raises(RuntimeError) as e:
+        validate_pattern('/hom*')
+    assert '/home' in str(e)
```
samdoran:
```suggestion
@pytest.mark.parametrize(
    ('pattern', 'match'), (
        ('/', '/'),
        ('/home', '/home'),
        ('/', 'Provided pattern could result in deleting system folders'),
        ('/home', 'Provided pattern could result in deleting system folders'),
        ('/hom*', '/home'),
    )
)
def test_validate_pattern(pattern, match):
    with pytest.raises(RuntimeError, match=match):
        validate_pattern(pattern)
```

samdoran (changes requested):

ansible_runner/cleanup.py
AlanCoding:
```python
>>> from tempfile import gettempdir
>>> gettempdir()
'/tmp'
```

Just drop that in instead?

ansible_runner/cleanup.py
AlanCoding: there's `ansible_runner.defaults`, which is fine with me

test/integration/containerized/test_cleanup_images.py
AlanCoding: If this were a fixture used by multiple tests I would have randomized the name. But why would the _same_ test run in parallel on the _same_ host? Yes, tests will run in parallel but I imagined it would never run the same test more than once. I still don't have a problem adding randomization because it won't hurt anything, I just wanted to explain myself.

ansible_runner/cleanup.py
AlanCoding: We do in the tests! As for the main code base, I came up a little empty on that search.

test/integration/containerized/test_cleanup_images.py
Shrews: Because of the use of `@pytest.mark.parametrize` on this test, it's possible this test could be running in parallel (once with podman and once with docker). We aren't currently using podman for these tests, only docker, but the plan is to eventually enable podman as well.

AlanCoding:
@samdoran thank you so much for the review comments. I hope that most of these will be resolved, but there's a question or two that still seems unresolved, like what pattern we want for invoking a subprocess.

ansible_runner/cleanup.py
fosterseth: prune here is probably better -- in awx it looks like we wanted to get proper stdout for the images being removed, so it first grabs a list of images, then removes them. Maybe I wasn't able to get it work properly with prune at the time

ansible_runner/cleanup.py
```diff
@@ -0,0 +1,191 @@
...
+    pidfile = os.path.join(dir, 'pid')
+
+    try:
+        with open(pidfile, 'r') as f:
+            pid = int(f.readline())
+    except IOError:
+        return False
+
+    try:
+        os.kill(pid, signal.SIG_DFL)
```
Shrews, line 81: Why the choice of `SIG_DFL` here? According to the [signal docs](https://docs.python.org/3/library/signal.html#signal.SIG_DFL) this would usually cause a core dump and exit. The usual pattern is to first use `SIGTERM` and then maybe `SIGKILL` if you need to be more forceful.

AlanCoding: Okay, what I _should_ have done is to refactor existing code here:

https://github.com/ansible/ansible-runner/blob/4f67a3a9e2f20e5d4da313958881eeef67397011/ansible_runner/__main__.py#L914

Into a `is_alive` method that lives somewhere else. However terrible the choice of this logic is, I do not have any intention of addressing it with this change set.

Shrews: Ah, you simply moved the code. Yes, I agree you shouldn't need to address that here. Thx for the explanation.

ansible_runner/utils/__init__.py
```diff
@@ -27,19 +27,21 @@
...
+    """Deletes folder, returns True or False based on whether a change happened."""
     try:
         shutil.rmtree(folder)
```
chrismeyersfsu, line 33: For systems that we care about, it seems that `rmtree()` (by default) will NOT follow symlinks. Which I like because security.

https://docs.python.org/3/library/shutil.html#shutil.rmtree.avoids_symlink_attacks

Shrews (approved):
I *think* all of Sam's comments here have been addressed. If not, those can be addressed in followup PRs.

beeankha (approved):

test/integration/containerized/test_cleanup_images.py
samdoran: The other thing is multiple instances of the same tests with different Python versions running at the same time.

ansible_runner/cleanup.py
samdoran: If there isn't a config option in runner for specifying the temp dir, then yes, `tempfile.gettempdir()` is a better option than hard coding `/tmp`.

ansible_runner/cleanup.py
samdoran: Let's go with that. Keep it all caps since it  is a global (the other values in that file should be all caps too, but that's for another PR).

ansible_runner/__main__.py
```diff
@@ -39,6 +39,7 @@
 
 from ansible_runner import run
 from ansible_runner import output
+from ansible_runner import cleanup
```
samdoran, line 42: Nit: combine multiple imports.
```suggestion
from ansible_runner import (
    cleanup,
    output,
    run,
)
```

ansible_runner/cleanup.py
samdoran: Ok. I was talking about `run_command()` and `run_command_async()` but those do a bit more than run commands since they also instantiate a `Runner` object.

We should add a generic and simple public function for running commands at a later date.

ansible_runner/defaults.py
```diff
@@ -1,2 +1,6 @@
 default_process_isolation_executable = 'podman'
 default_container_image = 'quay.io/ansible/ansible-runner:devel'
+registry_auth_prefix = 'ansible_runner_registry_'
```
samdoran, line 3: Nit: this should be all caps as well since it's a global.
```suggestion
REGISTRY_AUTH_PREFIX = 'ansible_runner_registry_'
```

docs/remote_jobs.rst
```diff
@@ -25,14 +25,36 @@ The `ansible-runner worker` command accepts this stream, runs the playbook, and
...
+Cleanup of Resources Used by Jobs
+---------------------------------
+
+The transmit and process commands do not offer any automatic deletion of the
+private data directory or artifacts, because these are how the user interacts with runner.
+
+When running `ansible-runner worker`, if no `--private-data-dir` is given,
+it will extract the contents to a temporary directory which is deleted at the end of execution.
+You can use the `--delete` flag in conjunction with `--private-data-dir` to assure that
+the provided directory is deleted at the end of execution.
```
samdoran, line 42: For RST, double backticks are needed for inline code formatting.
```suggestion
When running ``ansible-runner worker``, if no ``--private-data-dir`` is given,
it will extract the contents to a temporary directory which is deleted at the end of execution.
You can use the ``--delete`` flag in conjunction with ``--private-data-dir`` to assure that
the provided directory is deleted at the end of execution.
```

docs/remote_jobs.rst
```diff
@@ -25,14 +25,36 @@ The `ansible-runner worker` command accepts this stream, runs the playbook, and
...
+
+The following command offers out-of-band cleanup.
+
+    $ ansible-runner worker cleanup --file-pattern=/tmp/foo_*
```
samdoran, line 46: Probably need to quote this to avoid shell expansion.
```suggestion
    $ ansible-runner worker cleanup --file-pattern='/tmp/foo_*'
```

docs/remote_jobs.rst
```diff
@@ -25,14 +25,36 @@ The `ansible-runner worker` command accepts this stream, runs the playbook, and
...
+
+This would assure that old directories that fit the file glob "/tmp/foo_*" are deleted,
+which would could be used to assure cleanup of paths created by commands like
+`ansible-runner worker --private_data_dir=/tmp/foo_3`, for example.
+NOTE: see the `--grace-period` option, which sets the time window.
+
+This command also takes a `--remove-images` option to run the podman or docker `rmi` command.
+There is otherwise no automatic cleanup of images used by a run,
+even if `container_auth_data` is used to pull from a private container registry.
+To be sure that layers are deleted as well, the `--image-prune` flag is necessary.
```
samdoran, line 56:
```suggestion
``ansible-runner worker --private_data_dir=/tmp/foo_3``, for example.
NOTE: see the ``--grace-period`` option, which sets the time window.

This command also takes a ``--remove-images`` option to run the podman or docker ``rmi`` command.
There is otherwise no automatic cleanup of images used by a run,
even if ``container_auth_data`` is used to pull from a private container registry.
To be sure that layers are deleted as well, the ``--image-prune`` flag is necessary.
```

test/unit/test_cleanup.py
```diff
@@ -0,0 +1,85 @@
...
+def test_cleanup_command_grace_period(tmp_path):
+    old_dir = str(tmp_path / 'modtime_old_xyz')
+    new_dir = str(tmp_path / 'modtime_new_abc')
+    os.mkdir(old_dir)
+    time.sleep(1)
+    os.mkdir(new_dir)
+    ct = cleanup_dirs(pattern=str(tmp_path / 'modtime_*_*'), grace_period=1. / 60.)
+    assert ct == 1
+    assert not os.path.exists(old_dir)
+    assert os.path.exists(new_dir)
```
samdoran, line 46: You could keep these as `Path` objects.
```suggestion
    old_dir = tmp_path / 'modtime_old_xyz'
    new_dir = tmp_path / 'modtime_new_abc'
    
    old_dir.mkdir()
    time.sleep(1)
    new_dir.mkdir()
    
    ct = cleanup_dirs(pattern=str(tmp_path / 'modtime_*_*'), grace_period=1. / 60.)
    
    assert ct == 1
    assert not old_dir.exists()
    assert new_dir.exists()
```

Side note: using `time.sleep()` in tests is usually asking for instability. It's better to just set a/c/mtimes explicitly rather than slow down the tests unnecessary.

test/unit/test_utils.py
```diff
@@ -18,6 +19,28 @@
 from ansible_runner.utils.streaming import stream_dir, unstream_dir
 
 
+def test_cleanup_folder(tmp_path):
+    folder_path = tmp_path / 'a_folder'
+    folder_path.mkdir()
+    assert folder_path.exists()  # sanity
```
samdoran, line 25: I think this assertion and the same on line 32 are testing `pathlib` or the state of the test environment, not our function, and are unnecessary.

samdoran (approved):
No blockers, just some suggestions.

PR #847, opened 2021-09-23
Fix tests to allow them to run in parallel without conflict

samdoran:
Ensure a custom copy of the data directory is used for each test rather than having multiple tests share the same directory. This means the default tox environment no longer needs to have two separate `pytest` commands.

- Change the tests so they do not create artifacts in the source tree and remove corresponding ignore entries.
- Remove custom `pytest` marks and define `testpaths` in `pytest.ini`.
- Consolidate test fixtures and rename "data" to "projects"
- Various improvements to playbooks and inventory used it tests
- Disable Ansible devel warning in containerized tests
- Change `yamllint` settings to be a bit more sane
- Add a custom marker that will generate tests for each supported container runtime


AlanCoding:
This is great to see. I know it's a lot of work. I had tried to work at the edges of some of these issues before and got really overwhelmed.

I think I introduced the `test/data` folder. Everything you're doing here looks right, like, so obvious there is no question about correctness, but just difficulty in actually getting it working. Let me know if I can help to get it past the finish line.

(and really, it looks like the only CI failures are the linter failures which are already marked in the diff, plus the regular Zuul flake)

test/conftest.py
```diff
@@ -25,6 +28,15 @@ def skipif_pre_ansible28(is_pre_ansible28):
...
 
 
-@pytest.fixture(scope='session')
-def test_data_dir():
-    return os.path.join(os.path.dirname(__file__), 'data')
+@pytest.fixture
+def test_data_dir(tmp_path):
+    source = Path(__file__).parent / 'fixtures' / 'data'
+    dest = tmp_path / 'data'
+    shutil.copytree(source, dest)
```
AlanCoding: clever.

This is function scope, and copies the entire data directory? As long as the user clears old artifacts before running, that's 14 dirs / 14 files. All of these tend to be very small in the scheme of things, so this feels like the right kind of balance to me. :+1:

AlanCoding (approved):
looks like you might only need to fix the linter issues before this can be merged

test/conftest.py
samdoran: The more I look at this, the more it seems like it was a collection of project directories and only one is needed per test? I don't think the whole thing needs to be duplicated for each test, but that's what I did before I started to realize what these directories were. It's a bit blunt, but seems to adequately isolate the tests.

I'm open to better ways to do this rather than copying the whole directory structure around.

> This is function scope, and copies the entire data directory?

Yes.

> As long as the user clears old artifacts before running

Hopefully there are _no_ artifacts in these directories. That was the goal. I removed the `.gitignore` entries that would hide errant tests depositing artifacts in the source tree.

samdoran:
> looks like you might only need to fix the linter issues before this can be merged

I wish. 😞

It looks like the CI environment is different than my local environment in some way. `ansible-runner-tox-ansible-base` and `ansible-runner-tox-ansible29` are failing becasue they cannot find the playbooks inside the temporary directory. I'm unable to reproduce it locally (on Linux on macOS) so I'm not sure what's different in the CI environment.

test/integration/containerized/test_cli_containerized.py
```diff
@@ -12,20 +12,17 @@ def skip_if_no_podman(container_runtime_installed):
         pytest.skip('podman container runtime(s) not available')
 
 
-@pytest.mark.serial
 def test_module_run(cli, skip_if_no_podman):
     r = cli(['run', '-m', 'ping', '--hosts', 'localhost', os.path.join(HERE, 'priv_data')])
```
AlanCoding: don't you need to move the `priv_data` folder into the `test/data` folder? Otherwise, this isn't ready to be ran in parallel.

samdoran: Yup, I missed that one.

Shrews (approved):

PR #854, opened 2021-09-28
Ensure that ansible-runner worker --worker-info Returns Unique IDs

### LINKED ISSUE #850: ansible-runner worker --worker-info is returning same UUID for 3 different VM's
`ansible-runner worker --worker-info` is returning the same UUID for three different VM's on IBM Cloud. Looking at the code, it reads the machine-id and puts that as UUID, which is same for all the three vm's on IBM cloud. It seems the machine-id is not a full proof unique value. We should probably construct the unique UUID with something else, may be MAC address ?

The same UUID is causing an issue on Automation Controller installation. These vm's were supposed to be three different execution nodes, but because of the same UUID, that doesnt happen and Automation Controller only shows 1 execution node.

```
TASK [ansible.automation_platform_installer.receptor : Get UUID for execution nodes] ***
^[[0;36mskipping: [165.192.135.199] => {"changed": false, "skip_reason": "Conditional result was False"}^[[0m
^[[0;33mchanged: [158.175.191.143] => {"changed": true, "cmd": "ansible-runner worker --worker-info", "delta": "0:00:00.156578", "end": "2021-09-27 10:34:12.437573", "msg": "", "rc": 0, "start": "2021-09-27 10:34:12.280995", "stderr": "", "stderr_lines": [], "stdout": "{cpu_count: 4, errors: [], mem_in_bytes: 16600657920, runner_version: 2.1.0, uuid: 06dd765ca427f032a6f68d4152c8be32}", "stdout_lines": ["{cpu_count: 4, errors: [], mem_in_bytes: 16600657920, runner_version: 2.1.0, uuid: 06dd765ca427f032a6f68d4152c8be32}"]}^[[0m
^[[0;33mchanged: [161.156.198.69] => {"changed": true, "cmd": "ansible-runner worker --worker-info", "delta": "0:00:00.166301", "end": "2021-09-27 10:34:12.484405", "msg": "", "rc": 0, "start": "2021-09-27 10:34:12.318104", "stderr": "", "stderr_lines": [], "stdout": "{cpu_count: 4, errors: [], mem_in_bytes: 16600657920, runner_version: 2.1.0, uuid: 06dd765ca427f032a6f68d4152c8be32}", "stdout_lines": ["{cpu_count: 4, errors: [], mem_in_bytes: 16600657920, runner_version: 2.1.0, uuid: 06dd765ca427f032a6f68d4152c8be32}"]}^[[0m
^[[0;33mchanged: [158.175.191.141] => {"changed": true, "cmd": "ansible-runner worker --worker-info", "delta": "0:00:00.160766", "end": "2021-09-27 10:34:12.539002", "msg": "", "rc": 0, "start": "2021-09-27 10:34:12.378236", "stderr": "", "stderr_lines": [], "stdout": "{cpu_count: 4, errors: [], mem_in_bytes: 16600657920, runner_version: 2.1.0, uuid: 06dd765ca427f032a6f68d4152c8be32}", "stdout_lines": ["{cpu_count: 4, errors: [], mem_in_bytes: 16600657920, runner_version: 2.1.0, uuid: 06dd765ca427f032a6f68d4152c8be32}"]}^[[0m
```

beeankha:
Fixes https://github.com/ansible/ansible-runner/issues/850

In certain cases (e.g. when utilizing VMs spun up with IBM Cloud), the provided machine ID for each node will sometimes be identical; this code change would make it so that we purposefully generate a brand new UUID for each node and store it in a file for retrieval (the details of this file and its location are still TBD); this will ensure an actually unique UUID for each worker node.

### DIFF
--- a/ansible_runner/utils/capacity.py
+++ b/ansible_runner/utils/capacity.py
@@ -24,7 +24,10 @@ def get_mem_in_bytes():
         return error
 
 
-def ensure_uuid(uuid_file_path=Path.home().joinpath('.ansible_runner_uuid')):
+def ensure_uuid(uuid_file_path=None, mode=0o600):
+    if uuid_file_path is None:
+        uuid_file_path = Path.home().joinpath('.ansible_runner_uuid')
+
     if uuid_file_path.exists():
         # Read the contents of file if it already exists
         saved_uuid = uuid_file_path.read_text()
@@ -35,7 +38,10 @@ def ensure_uuid(uuid_file_path=Path.home().joinpath('.ansible_runner_uuid')):
         return newly_generated_uuid
 
 
-def _set_uuid(uuid_file_path=Path.home().joinpath('.ansible_runner_uuid')):
+def _set_uuid(uuid_file_path=None, mode=0o600):
+    if uuid_file_path is None:
+        uuid_file_path = Path.home().joinpath('.ansible_runner_uuid')
+
     generated_uuid = str(uuid.uuid4())
 
     # Store the newly-generated UUID in a new file in home dir
--- a/ansible_runner/utils/capacity.py
+++ b/ansible_runner/utils/capacity.py
@@ -34,7 +34,7 @@ def ensure_uuid(uuid_file_path=None, mode=0o0600):
         return saved_uuid.strip()
     else:
         # Generate a new UUID if file is not found
-        newly_generated_uuid = _set_uuid(mode)
+        newly_generated_uuid = _set_uuid(uuid_file_path, mode)
         return newly_generated_uuid
 
 
--- a/ansible_runner/utils/capacity.py
+++ b/ansible_runner/utils/capacity.py
@@ -1,7 +1,5 @@
 import multiprocessing
-import os
 import re
-import stat
 import uuid
 
 from pathlib import Path
@@ -27,14 +25,14 @@ def get_mem_in_bytes():
 
 
 def get_uuid():
-    uuid_file_path = Path('/etc/ansible/facts.d/uuid.txt')
+    uuid_file_path = Path.home() / '.ansible_runner_uuid'
     if uuid_file_path.exists():
-        # Read the contents of the uuid.txt file if it already exists
+        # Read the contents of .ansible_runner_uuid if it already exists
         with open(uuid_file_path) as f:
             saved_uuid = f.read()
-        return saved_uuid
+        return saved_uuid.strip()
     else:
-        # Generate a new UUID if no uuid.txt file is found
+        # Generate a new UUID if no .ansible_runner_uuid file is found
         newly_generated_uuid = _generate_uuid()
         return newly_generated_uuid
 
@@ -42,13 +40,8 @@ def get_uuid():
 def _generate_uuid():
     generated_uuid = str(uuid.uuid4())
 
-    # Store the newly-generated UUID in a new dir/file
-    uuid_dir = Path('/etc/ansible/facts.d')
-    uuid_dir.mkdir(parents=True, exist_ok=True)
-    uuid_file = 'uuid.txt'
-    uuid_file_path = uuid_dir / uuid_file
-
+    # Store the newly-generated UUID in a new file in home dir
+    uuid_file_path = Path.home() / '.ansible_runner_uuid'
     with uuid_file_path.open('w', encoding='utf-8') as uuid_file:
-        os.chmod(uuid_file_path, stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR)
         uuid_file.write(generated_uuid)
     return generated_uuid
--- a/ansible_runner/utils/capacity.py
+++ b/ansible_runner/utils/capacity.py
@@ -24,24 +24,21 @@ def get_mem_in_bytes():
         return error
 
 
-def get_uuid():
-    uuid_file_path = Path.home() / '.ansible_runner_uuid'
+def ensure_uuid(uuid_file_path=Path.home().joinpath('.ansible_runner_uuid')):
     if uuid_file_path.exists():
-        # Read the contents of .ansible_runner_uuid if it already exists
-        with open(uuid_file_path) as f:
-            saved_uuid = f.read()
+        # Read the contents of file if it already exists
+        saved_uuid = uuid_file_path.read_text()
         return saved_uuid.strip()
     else:
-        # Generate a new UUID if no .ansible_runner_uuid file is found
-        newly_generated_uuid = _generate_uuid()
+        # Generate a new UUID if file is not found
+        newly_generated_uuid = _set_uuid()
         return newly_generated_uuid
 
 
-def _generate_uuid():
+def _set_uuid(uuid_file_path=Path.home().joinpath('.ansible_runner_uuid')):
     generated_uuid = str(uuid.uuid4())
 
     # Store the newly-generated UUID in a new file in home dir
-    uuid_file_path = Path.home() / '.ansible_runner_uuid'
-    with uuid_file_path.open('w', encoding='utf-8') as uuid_file:
-        uuid_file.write(generated_uuid)
+    uuid_file_path.write_text(generated_uuid)
+
     return generated_uuid
--- a/ansible_runner/utils/capacity.py
+++ b/ansible_runner/utils/capacity.py
@@ -1,5 +1,10 @@
+import glob
 import multiprocessing
+import os
 import re
+import stat
+import tempfile
+import uuid
 
 
 def get_cpu_count():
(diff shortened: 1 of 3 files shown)


beeankha:
recheck

ansible_runner/utils/capacity.py
```diff
@@ -21,17 +26,24 @@ def get_mem_in_bytes():
...
-    except FileNotFoundError:
-        try:
-            with open('/etc/machine-id') as f:
-                uuid = f.read()
-                return uuid.strip()
-        except FileNotFoundError:
-            error = ("Could not find /var/lib/dbus/machine-id or "
-                     "/etc/machine-id files, UUID undiscoverable.")
-            return error
+def generate_or_get_uuid():
```
samdoran: Functions that combine multiple capabilities like this are generally harder to test and maintain. It's better to have simple, single purpose functions. That allows the caller to control the desired behavior.

This function could be changed to call `get_uuid()`, then save it to disk. That way `get_uuid()` would be concerned with uuid generation and this function would be concerned with making sure the file exists on disk. Maybe call rename it to `save_uuid()`, have it accept a path to check or save to as a parameter, and return the value that it either gets from the path or save a new uuid to if the file doesn't exist.

ansible_runner/utils/capacity.py
```diff
@@ -1,5 +1,10 @@
+import glob
 import multiprocessing
+import os
```
samdoran: You could use `pathlib` since this a Python 3 only project. That would remove the need for some of the other imports since `Path` objects have methods that provide the same functionality.

ansible_runner/utils/capacity.py
```diff
@@ -21,17 +26,24 @@ def get_mem_in_bytes():
         return error
 
 
-def get_uuid():
```
samdoran, line 24: Rather than deleting a public function, why not fall back to `uuid.getnode()` or `uuid.uuid4()`?

samdoran: Something like this may work:
```python
def get_uuid(path=''):
files = [
    Path('/var/lib/dbus/machine-id'),
    Path('/etc/machine-id'),
]

if path:
    files.append(Path(path))

for file in files:
    try:
        uuid = file.read_text()
        return uuid.strip()
     except FileNotFoundError:
        continue
        
return uuid.getnode()
```

ansible_runner/utils/capacity.py
```diff
@@ -21,17 +26,24 @@ def get_mem_in_bytes():
...
-        try:
-            with open('/etc/machine-id') as f:
-                uuid = f.read()
-                return uuid.strip()
-        except FileNotFoundError:
-            error = ("Could not find /var/lib/dbus/machine-id or "
-                     "/etc/machine-id files, UUID undiscoverable.")
-            return error
+def generate_or_get_uuid():
+    uuid_directory = glob.glob('/tmp/node_uuid_*')  # Location TBD
```
samdoran: I'm guessing this is a temporary placeholder. Using `glob` here could pick up multiple files, which could introduce a whole host of unexpected behavior. This should be looking at one and only one very specific location.

samdoran:
I just looked at the existing tests and there are _no_ tests at all for `get_uuid()`. That means that changing this function is risky since we 1) have nothing that validates the current behavior, 2) have no way of knowing if the changes made here break that behavior, and 3) have no tests for the new code.

At a minimum, we need tests for the net new code. It would be really great to create tests for the existing function before changing it as well.

ansible_runner/utils/capacity.py
```diff
@@ -22,16 +27,27 @@ def get_mem_in_bytes():
...
-    except FileNotFoundError:
-        try:
-            with open('/etc/machine-id') as f:
-                uuid = f.read()
-                return uuid.strip()
-        except FileNotFoundError:
-            error = ("Could not find /var/lib/dbus/machine-id or "
-                     "/etc/machine-id files, UUID undiscoverable.")
-            return error
+    uuid_directory = Path('/etc/ansible/facts.d')
```
shanemcd: I'm curious why we picked this location. Is this a canonical place for this kind of file?

ansible_runner/utils/capacity.py
```diff
@@ -22,16 +27,27 @@ def get_mem_in_bytes():
...
+        return newly_generated_uuid
+
+
+def generate_uuid():
+    generated_uuid = str(uuid.uuid4())
+
+    # Store the newly-generated UUID in a new dir/file
+    path = '/etc/ansible/facts.d'
+    uuid_file = 'uuid.txt'
+    uuid_file_path = pathlib.Path(path) / uuid_file
```
shanemcd: Is there any reason why this can't just be `os.path.join(path, uuid_file)`?

ansible_runner/utils/capacity.py
shanemcd: Oh, I think I see why now. So that file will be read and sent back when the heartbeat runs? Neat.

ansible_runner/utils/capacity.py
```diff
@@ -22,16 +27,27 @@ def get_mem_in_bytes():
...
+        with open(uuid_file_path) as f:
+            saved_uuid = f.read()
+        return saved_uuid
+    else:
+        # Generate a new UUID if no uuid.txt file is found
+        newly_generated_uuid = generate_uuid()
+        return newly_generated_uuid
+
+
+def generate_uuid():
```
shanemcd: Can this function get marked private? I only see it being called internally in this module.

ansible_runner/utils/capacity.py
chrismeyersfsu: We needed a place. I picked a place that was ansibley. VERY open to a better place.

shanemcd: Can someone point me to the place that we read this on the AWX side?

ansible_runner/utils/capacity.py
samdoran: That would change the result from a `Path` object to a `str`.

ansible_runner/utils/capacity.py
```diff
@@ -22,16 +27,28 @@ def get_mem_in_bytes():
...
+def _generate_uuid():
+    generated_uuid = str(uuid.uuid4())
+
+    # Store the newly-generated UUID in a new dir/file
+    uuid_dir = Path('/etc/ansible/facts.d')
+    uuid_dir.mkdir(parents=True, exist_ok=True)
+    uuid_file = 'uuid.txt'
+    uuid_file_path = uuid_dir / uuid_file
+
+    with uuid_file_path.open('w', encoding='utf-8') as uuid_file:
```
chrismeyersfsu: We want to make sure that the files starts its life with the correct permissions. Otherwise, an attack could write a program that just sits in a loop and keeps trying to "hijack" the new file that we are creating. They would be successful if their code ran in between the file open line and the chmod line.

Ansible core has to deal with this problem often. I think @samdoran can recommend the best way to deal with this.

ansible_runner/utils/capacity.py
```diff
@@ -24,24 +24,21 @@ def get_mem_in_bytes():
         return error
 
 
-def get_uuid():
```
samdoran, line 24: Do we need to keep `get_uuid()` around for backwards compatibility?

samdoran: It probably wouldn't hurt. We can just deprecate it. Deleting it entirely is a bit abrupt since it's a public function.

shanemcd: I don't think so, it was never shipped, just landed a month ago https://github.com/ansible/ansible-runner/commit/49353d2d1d6a4fbb9ed47c3ca9a2b6e82f392e41

shanemcd (approved):

ansible_runner/utils/capacity.py
samdoran: What should the correct permissions be? How sensitive is this file?

The way Core (partially) defends against this type of race condition/attack is by the randomness of the parent temporary directory name, not the module payload file name. I'm not sure the payload that contains a module to be executed (in the case of Core) has the same risk profile as a file that contains a UUID. But I don't fully understand what a malicious actor could gain by substituting their own UUID for ours.

In the case of this file, we want it to be in a deterministic location, so randomizing the path and/or filename aren't really options. I suppose just setting the permissions to 400 would be enough?

That the default path is inside the user's home directory, and this code does not create any directories in the `uuid_file_path` also mitigates quite a bit of risk. The real trouble comes with `os.makedirs()` or `mkdir -p`, which isn't happening here.

ansible_runner/utils/capacity.py
samdoran: Sweet!

ansible_runner/utils/capacity.py
samdoran: In any case, it's probably a good idea to be explicit about the permissions on the file rather than relying on the `umask`. Should this be 600, 640, or mainly just not world readable?

samdoran: I set the file to `600` by default, but the caller can change it.

samdoran (approved):

samdoran:
recheck

PR #855, opened 2021-09-28
Accept a dictionary for playbook

### LINKED ISSUE #642: Inventory specified as a dictionary isn't actually supported
The documentation says the following about the `inventory` parameter:
>  Overridees the inventory directory/file (supplied at private_data_dir/inventory) with
>  a specific host or list of hosts. This can take the form of
> 
>  - Path to the inventory file in the private_data_dir
>  - **Native python dict supporting the YAML/json inventory structure**
>  - A text INI formatted string
>  - A list of inventory sources, or an empty list to disable passing inventory

And yet if one passes a dictionary, it is not stringified properly which results e.g. in the following cmdline:
```
['ansible', '-i', {'all': {'dev': {'hosts': ['localhost']}, 'ungrouped': {'hosts': ['my_machine1', 'my_machine2']}}}, '-m', 'setup', '*']
```
...which leads to the following exception:
```
  File "/usr/lib/python3.9/site-packages/ansible_runner/runner.py", line 171, in run
    child = pexpect.spawn(
  File "/usr/lib/python3.9/site-packages/pexpect/pty_spawn.py", line 205, in __init__
    self._spawn(command, args, preexec_fn, dimensions)
  File "/usr/lib/python3.9/site-packages/pexpect/pty_spawn.py", line 281, in _spawn
    self.name = '<' + ' '.join(self.args) + '>'
TypeError: sequence item 2: expected str instance, dict found
```

### LINKED ISSUE #852: sending the playbook contents as a dictionary to the runner python module not working.
Hello As requested in the libera chat i open this issue here. 

https://github.com/ansible/ansible-runner/blob/devel/ansible_runner/interface.py#L138-L139
according to this this parameter to the enty poing will take a list as a data structure or even just a string of content. but it seems it only accepts the 'string' of a filename. 

I try this: 

```
import ansible_runner

inv = {'all': {'hosts': 'vms' }}

play_src={
{
  "gather_facts": false,
  "tasks": [
    {
      "debug": {
        "msg": "hello"
      },
      "name": "debug output"
    }
  ],
  "hosts": "all",
  "become": true
}

r = ansible_runner.run(private_data_dir='.', inventory=inv, playbook=play_src)

```
this gives an error: 

  File "./test_runner.py", line 45, in <module>
    r = ansible_runner.run(private_data_dir='.', inventory=inv, playbook=play_src)
  File "/usr/home/ansible/venv/lib/python2.7/site-packages/ansible_runner/interface.py", line 178, in run
    r.run()
  File "/usr/home/ansible/venv/lib/python2.7/site-packages/ansible_runner/runner.py", line 112, in run
    command = [a.decode('utf-8') for a in self.config.command]
AttributeError: 'dict' object has no attribute 'decode'

it appears that runner is expecting a string (decode attribute is only available for strings. so I  hardcoded the yaml file as a string. 

```
play= """
---
- hosts: all
  gather_facts: no
  become: yes

  tasks:
    - name: debug output
      debug:
        msg: 'hello'
"""

print play.decode('utf-8')


r = ansible_runner.run(private_data_dir='.', inventory=inv, playbook=play)
```
this gives the error: 

```
ERROR! the playbook:
---
- hosts: all
  gather_facts: no
  become: yes

  tasks:
    - name: debug output
      debug:
        msg: 'hello'
 could not be found
```

the expected behavior is that the python dictionary is used as a playbook source.

samdoran:
Fixes #852
Fixes #642

The validation function `isplaybook()` expects a playbook to be a list of dictionaries, not a bare dictionary or a string. Update the docs to reflect this requirement and cast a bare dictionary to a list before passing to `isplaybook()`.

- [x] Add tests


samdoran:
It may also be a good idea to add some error handling here for invalid values to `playbook` in order to fail early rather than later in unexpected ways.

samdoran:
recheck

samdoran:
recheck

ansible_runner/interface.py
```diff
@@ -135,7 +135,7 @@ def run(**kwargs):
     :param ident: The run identifier for this invocation of Runner. Will be used to create and name
                   the artifact directory holding the results of the invocation.
     :param json_mode: Store event data in place of stdout on the console and in the stdout file
-    :param playbook: The playbook (either supplied here as a list or string... or as a path relative to
+    :param playbook: The playbook (either a list or dictionary or plays, or as a path relative to
```
Shrews:
```suggestion
    :param playbook: The playbook (either a list or dictionary of plays, or as a path relative to
```

ansible_runner/utils/__init__.py
```diff
@@ -201,10 +202,15 @@ def dump_artifacts(kwargs):
...
         kwargs['envvars']['ANSIBLE_ROLES_PATH'] = roles_path
 
-    obj = kwargs.get('playbook')
-    if obj and isplaybook(obj):
-        path = os.path.join(private_data_dir, 'project')
-        kwargs['playbook'] = dump_artifact(json.dumps(obj), path, 'main.json')
+    playbook = kwargs.get('playbook')
+    if playbook:
+        # Ensure the play is a list of dictionaries
+        if isinstance(playbook, MutableMapping):
```
Shrews, line 209: Wouldn't `Mapping` work just as well here? Not sure why we need the extra import.

samdoran: It would. I would probably prefer to change to using `MutableMapping` instead of `Mapping` since it's more accurate.

Shrews (approved):

shanemcd (approved):
Nice work

PR #862, opened 2021-10-05
add default cancel_callback to handle common signals

### LINKED ISSUE #861: Add default cancel_callback for ansible-runner CLI invocation
If you invoke runner as a python library, you have the option of specifying a cancel_callback ([example](https://github.com/redhat-cip/dci-pipeline/commit/4fa280b139436e420720b928b384a9a0172c91a0#diff-73210e559f23e668fa9e27db783bd2f46342a0b4aae68d7cdcc7a141821eb90aR769)). When calling runner via the CLI, however, there isn't an option to [pass in a cancel_callback](https://github.com/ansible/ansible-runner/blob/b9a4a79fa489d823b991c453c23877bc500f5a1d/ansible_runner/__main__.py#L852), so it defaults to being left [undefined](https://github.com/ansible/ansible-runner/blob/b9a4a79fa489d823b991c453c23877bc500f5a1d/ansible_runner/interface.py#L94).

For the CLI, it seems like it would be reasonable to default to using a cancel_callback with a signal handler that would catch SIGINT. This would allow runner to recognize an incoming signal as a request to gracefully shut down the running job, including ensuring that a container spun up as part of process isolation was stopped.

jladdjr:
Fixes https://github.com/ansible/ansible-runner/issues/861

--

If you invoke runner as a python library, you have the option of specifying a cancel_callback ([example](https://github.com/redhat-cip/dci-pipeline/commit/4fa280b139436e420720b928b384a9a0172c91a0#diff-73210e559f23e668fa9e27db783bd2f46342a0b4aae68d7cdcc7a141821eb90aR769)). When calling runner via the CLI, however, there isn't an option to [pass in a cancel_callback](https://github.com/ansible/ansible-runner/blob/b9a4a79fa489d823b991c453c23877bc500f5a1d/ansible_runner/__main__.py#L852), so it defaults to being left [undefined](https://github.com/ansible/ansible-runner/blob/b9a4a79fa489d823b991c453c23877bc500f5a1d/ansible_runner/interface.py#L94).

For the CLI, it seems like it would be reasonable to default to using a cancel_callback with a signal handler that would catch SIGINT and SIGTERM. This would allow runner to recognize an incoming signal as a request to gracefully shut down the running job, including ensuring that a container spun up as part of process isolation was stopped.

This PR updates runner to use a default cancel_callback when none is provided. The default cancel_callback catches both SIGINT and SIGTERM and gives the runner process (and other spawned processes, such as those started by a container runtime) a chance to exit gracefully.

## steps to reproduce

1. Modify demo/project/test.yml so that it runs a pause task w/ minutes: 90
2. `ansible-runner run --process-isolation --process-isolation-executable podman -p test.yml demo`
3. pkill ansible-runner
4. Verify container continues running even after ansible-runner process exits.


AlanCoding:
:+1: This appears to me to be doing the right thing.

There's a lot of testing work still to be done, of course, but it's important to show people the WIP to set expectations. I could see this being the final patch to the code itself.

AlanCoding:
I wrote a test here expecting to reproduce the original bug

https://github.com/ansible/ansible-runner/compare/devel...AlanCoding:container_kill_test?expand=1

However, this test was not able to replicate the issue. When the ansible-runner process is killed, the container dies with it. That is without this patch.

jladdjr:
@AlanCoding made a few tweaks to the test that were needed for it to run for me:
* added `--process-isolation` flags so that when we run `ansible-runner start` we ensure that podman is used
* instead of podman pdd, used sleep pdd (only place I saw the sleep.yml playbook in the repo)

With those changes I found that the test passed on devel runner (w/out my patch). Suspecting that this might be a matter of timing, I added a five second sleep after we call `ansible-runner start`. Sure enough, this resulted in the test failing 80% of the time (the same results we saw when running a test at the controller level).

I think that even with the five second sleep, the test still represents a very real situation that users could find themselves in. If anything, it seems likely that a user would cancel a job once it's well underway, moreso than cancelling an instant after it launched.

@AlanCoding I went ahead and added your changeset into this PR. We can continue polishing it as needed, but it seems like a reliable reproducer of the issue we're trying to address here.

ansible_runner/interface.py
```diff
@@ -91,7 +93,15 @@ def init_runner(**kwargs):
     event_callback_handler = kwargs.pop('event_handler', None)
     status_callback_handler = kwargs.pop('status_handler', None)
     artifacts_handler = kwargs.pop('artifacts_handler', None)
-    cancel_callback = kwargs.pop('cancel_callback', None)
+    if kwargs.get('cancel_callback'):
+        cancel_callback = kwargs.pop('cancel_callback')
+    # Only the main thread is allowed to set a new signal handler
+    elif threading.current_thread() is threading.main_thread():
```
jladdjr: There are some tests that will call runner inside of a thread. This exposed the issue of trying to set a SignalHandler when runner isn't running inside the main thread.

This seems like it would be an issue when runner is called asynchronously (see `ansible_runner.interface.run_async`'s use of threading). This has not been an issue for the awx use case where receptor calls runner via `ansible-runner worker`.

jladdjr:
@samdoran @Shrews - Wanted to check in with you to see if you have any feedback / concerns about getting this in. This PR helps address a scenario we're seeing with awx where receptor sends an ansible-runner process a SIGTERM which results in the container started by runner to be orphaned. The net effect is that the awx job is listed as cancelled despite a container continuing to execute the associated job. We would make use of a cancel_callback, except that runner is being invoked via CLI by receptor. Since that's not an option, thought we could introduce signal-handling behavior as a sane default implementation of the cancel_callback. Let us know what you think.

ansible_runner/interface.py
```diff
@@ -34,10 +34,12 @@
 from ansible_runner.utils import (
     dump_artifacts,
     check_isolation_executable_installed,
-    santize_json_response
+    santize_json_response,
+    SignalHandler
```
samdoran: Just use `threading.Event` since it essentially does the name thing.
```suggestion
```

test/integration/containerized/test_cli_containerized.py
```diff
@@ -29,3 +33,39 @@ def test_playbook_run(cli, skip_if_no_podman):
 def test_provide_env_var(cli, skip_if_no_podman, test_data_dir):
     r = cli(['run', os.path.join(test_data_dir, 'job_env'), '-p', 'printenv.yml'])
     assert 'gifmyvqok2' in r.stdout, r.stdout
+
+
+@pytest.mark.serial
```
samdoran, line 41: I am working on removing the need for this custom mark. It usually indicates there is a problem with the test not properly isolating itself.
```suggestion
```

ansible_runner/interface.py
```diff
@@ -91,7 +93,15 @@ def init_runner(**kwargs):
...
-    cancel_callback = kwargs.pop('cancel_callback', None)
+    if kwargs.get('cancel_callback'):
+        cancel_callback = kwargs.pop('cancel_callback')
+    # Only the main thread is allowed to set a new signal handler
+    elif threading.current_thread() is threading.main_thread():
+        signal_handler = SignalHandler()
+        cancel_callback = signal_handler.called
+    else:
+        logger.warning("Unable to set cancel_callback for signal handling; not in main thread")
+        cancel_callback = None
```
samdoran: This could potentially be somewhat noisy. Anytime this is run in a thread using the default cancel callback, this warning will be emitted. Maybe use a Sentinel or some other check to make sure this is emitted only once.

ansible_runner/interface.py
```diff
@@ -91,7 +93,15 @@ def init_runner(**kwargs):
     event_callback_handler = kwargs.pop('event_handler', None)
     status_callback_handler = kwargs.pop('status_handler', None)
     artifacts_handler = kwargs.pop('artifacts_handler', None)
-    cancel_callback = kwargs.pop('cancel_callback', None)
+    if kwargs.get('cancel_callback'):
+        cancel_callback = kwargs.pop('cancel_callback')
```
samdoran: For easier reading, just `pop` the value and store it.

```suggestion
    cancel_callback = kwargs.pop('cancel_callback')
    if cancel_callback:
```

ansible_runner/interface.py
```diff
@@ -91,7 +93,15 @@ def init_runner(**kwargs):
...
+    # Only the main thread is allowed to set a new signal handler
+    elif threading.current_thread() is threading.main_thread():
+        signal_handler = SignalHandler()
+        cancel_callback = signal_handler.called
```
samdoran: An example of using `threading.Event` to accomplish the same thing.
```suggestion
        signal_event = threading.Event()
        signal.signal(signal.SIGTERM, _signal_shutdown(signal_event))
        signal.signal(signal.SIGINT, _signal_shutdown(signal_event))
        cancel_callback = signal_event.is_set
```

```python
def _signal_shutdown(signal_event):
    def handler(number, frame):
        signal_event.set()
        
    return handler
```

test/integration/containerized/test_cli_containerized.py
```diff
@@ -29,3 +33,40 @@ def test_playbook_run(cli, skip_if_no_podman):
...
+    tries = 5
+    for i in range(tries):
+        if container_is_running():
+            break
+        time.sleep(1)
+    else:
+        assert container_is_running()
+
+    # give playbook execution time to start
+    time.sleep(5)
```
samdoran: Is there a better way to detect this other than artificial sleeps? Intentionally slowing down tests is always something to avoid if possible.

ansible_runner/utils/__init__.py
```diff
@@ -464,3 +465,17 @@ def get_executable_path(name):
     if exec_path is None:
         raise ConfigurationError(f"{name} command not found")
     return exec_path
+
+
+# signal handler for default cancel_callback
+class SignalHandler:
```
samdoran: This basically reimplements `threading.Event`. That class cloud be used instead.

samdoran (changes requested):
We should also add unit/functional tests to ensure the correct the cancel callback is set properly and that the handler is called as expected.

test/integration/containerized/test_cli_containerized.py
```diff
@@ -29,3 +33,40 @@ def test_playbook_run(cli, skip_if_no_podman):
 def test_provide_env_var(cli, skip_if_no_podman, test_data_dir):
     r = cli(['run', os.path.join(test_data_dir, 'job_env'), '-p', 'printenv.yml'])
     assert 'gifmyvqok2' in r.stdout, r.stdout
+
+
+@pytest.mark.serial
+def test_cli_kill_cleanup(cli, test_data_dir, container_runtime_installed):
+    unique_string = str(uuid4()).replace('-', '')
+    ident = f'kill_test_{unique_string}'
+    pdd = os.path.join(test_data_dir, 'sleep')
```
samdoran, line 50: Currently this fixture is pointing to directory in the repo, which will cause problems. I have a PR to create isolated copies of the data dir, but it isn't merged yet.

If the test is writing to `pdd`, then that is problematic.

AlanCoding: I tested running

```
ansible-runner run demo --playbook=test.yml --process-isolation --process-isolation-executable=podman
```

And this doesn't appear to leave any gunk around in `demo/env/settings` that would make the containerization settings "sticky" with other tests. So maybe this is fine. Otherwise, my original test draft puts the test files in its own directory in `test/data`.

AlanCoding:
I think I've about finalized my independent testing script:

```bash
rm -rf /tmp/ansible-runner
cd /tmp
git clone https://github.com/jladdjr/ansible-runner.git
# git clone https://github.com/alancoding/ansible-runner.git
cd ansible-runner
git checkout add_default_cancel_callback_for_cli
# git checkout container_kill_test

python3 -m venv env
source env/bin/activate

pip install -e .
pip install ansible-core
pip install mock

for i in `seq 10`; do py.test test/integration -k test_cli_kill_cleanup -n0 > /dev/null; echo $?; done
deactivate
```

I run 2 variations of this, where it runs the test in the branch I linked above, or it runs the tests as a part of this PR.

Running my branch (with the test but without the fix) produces about 1 or 2 failures over the 10 runs. These failures are failures to teardown the container. Running with this branch consistently gives 0 failures. That's the confirmation I was looking for to say that this does what was intended.

Of course this is all with podman (what we care about for AWX).

ansible_runner/interface.py
jladdjr: Makes sense. I didn't want to silently disable the default, but I can see how this could get to be a bit much. I'll see if there's something I can do here.

test/integration/containerized/test_cli_containerized.py
jladdjr: 💯  totally agree. @AlanCoding mentioned that you were working on a more general solution for the test isolation problem.

I'll drop the marker and see if it results in this test being a noisy neighbor. With your larger fix on the way, though, how much do you care about us making this particular test isolated / avoiding the use of the serial marker?

test/integration/containerized/test_cli_containerized.py
jladdjr: Good question. I'm not exactly sure why it is that this increases reproducibility, but you're right we should poll for a known event if possible. I'll see if there's anything we can key off of.

AlanCoding: I want to add that reproducing the failure is flaky and hardware-dependent. The sleep has been found to increase the rate of failures encountered, but (depending on hardware) you can still get failures without the sleep. With or without the sleep, this has never been observed to fail when the fix is in place.

test/integration/containerized/test_cli_containerized.py
samdoran: Try the tests without it and see what happens. We can leave it in if needed.

The more important point is to just think about making isolated tests when writing them, and that this marker will go away in the future.

test/integration/containerized/test_cli_containerized.py
samdoran: If being able to reproduce this with in integration test is unreliable, then unit and functional tests would be more beneficial until we have a better understanding of the exact failure scenario.

ansible_runner/interface.py
jladdjr: I'm wrestling with this one a bit. I tried writing this up so that we had a flag indicating whether or not the warning has been issued already. But then I wondered, where would that flag need to be set so that separate runs were retrieving the same flag? In the scenario you outlined the warning is being generated anytime we're in a new thread. That seems to suggest that when we're in a thread, we need to set a flag so that future threads know not to send a warning. It seems like this would require sending a message via a queue, I'm not sure. Again, I'm just not exactly sure how we would do this. Do you have thoughts on how to make this work? How strongly do you feel about having this warning? It seems like it might be fine to leave it out. Just adding the signal handling behavior by itself is a big improvement, even if we can't always set one (and even if we don't notify the user that they're missing out).

test/integration/containerized/test_cli_containerized.py
jladdjr: For the purposes of this PR, it sounds like we can drop the sleep altogether, would you agree @AlanCoding based on your testing?

jladdjr: I dropped the sleep and still get a 20% failure rate. Leaving the sleep out.

@samdoran how do you feel about a 20% failure rate? The test will not provide a super strong signal, but at the same time it does seem like we can expect this test to start talking to us when things are off (and we will avoid the 5s penalty altogether).

AlanCoding: I would certainly like to avoid a 5 second sleep in a serial test, so I would delete it for the merge.

We do have an explanation for why the container is left running. We do not have an explanation for why the container often dies with the parent process (before this patch). I was never able to replicate it on my personal machine, and it could be because the podman `--tty` option is expected to make it die with its parent. I'm on the very recent 3.3.1. It's entirely possible that podman always intended this and just fixed some bugs in recent versions that make it work as expected.

ansible_runner/interface.py
jladdjr: Reimplemented signal handler using Threading library.

ansible_runner/interface.py
jladdjr: Reimplemented signal handler using Threading library.

ansible_runner/utils/__init__.py
jladdjr: Reimplemented signal handler using Threading library.

test/integration/containerized/test_cli_containerized.py
jladdjr: Deleted the sleep (and thanks for the speculation about what might account for the variation in reproducibility between environments).

test/integration/containerized/test_cli_containerized.py
jladdjr: Dropped serial, testing to see if there are any noisy neighbors

ansible_runner/interface.py
jladdjr: Updated

jladdjr:
@samdoran I think I've gone through most of the feedback at this point. Still waiting on some unit test results locally and from zuul atm.

It doesn't look like I have perms needed to mark discussions as resolved so I'll go through and drop a ✅  in places where I believe things have been addressed.

ansible_runner/interface.py
jladdjr: ✅

ansible_runner/interface.py
jladdjr: ✅

test/integration/containerized/test_cli_containerized.py
jladdjr: (In Progress - waiting for unit test results locally and from zuul)

ansible_runner/interface.py
jladdjr: (Pending feedback)

ansible_runner/interface.py
jladdjr: ✅

ansible_runner/interface.py
jladdjr: ✅

test/integration/containerized/test_cli_containerized.py
jladdjr: ✅

ansible_runner/utils/__init__.py
jladdjr: ✅

jladdjr:
Only see two items of feedback left here:

- [x] Dropping serial marker (waiting on test results) - UPDATE: tests passed:
https://github.com/ansible/ansible-runner/pull/862#discussion_r724352896

- [x] Looking for feedback re: issuing a warning to user re: not being able to set signal handler:
https://github.com/ansible/ansible-runner/pull/862#discussion_r724365069

@samdoran any thoughts on the warning flag? ^

ansible_runner/utils/__init__.py
```diff
@@ -464,3 +465,19 @@ def get_executable_path(name):
...
+
+
+def signal_handler():
+    # Only the main thread is allowed to set a new signal handler
+    if threading.current_thread() is not threading.main_thread():
+        return None
+
+    signal_event = threading.Event()
+    # closure to set signal event
+    def _handler(number, frame):
```
jladdjr, line 478: we want to create the signal event
.. then pass it in to a closure
.. that will trigger the signal
.. and finally register the function handle / callable to `signal.signal`

When the associated signal happens `_handler` gets called, the signal event is registered, `signal_event.is_set` becomes true, meaning that `cancel_callback` becomes true and 🥁  ansible-runner knows it's time to clean up

ansible_runner/interface.py
jladdjr: needed to convert this to a closure

jladdjr:
@samdoran - @thenets and I started testing the newest implementation using the threading / signal library and are seeing flakiness. The unit test that we added here has gone from passing 100% of the time to passing 55% of the time. I'm not exactly sure where the flakiness is coming from, but I was wondering how strongly you felt about taking this approach compared to the original one using SignalHandler?

For the purposes of testing, going to revert back to the Signal Handler approach. The threading / signal approach that I tried out today can be seen at 303794c1b8ecdb5ca211ab3cfe2accf0c6d316e7.

jladdjr:
@samdoran - took a break and came back to make fresh pass at the tests again. Very carefully re-ran the tests for the new version of cancel_callback using the threading / signal libraries and can confirm that I got a clean set of results this time, both from the runner unit tests and from awx testing. Here are ten back-to-back test runs of the new unit test added in this PR:

image

So, I think this _is_ looking good. I reviewed everything and think the PR should be good-to-go again. So, the only remaining item to revisit (apart from a general review of how I implemented your suggested signal handler) is:

- [x] Looking for feedback re: issuing a warning to user re: not being able to set signal handler:
https://github.com/ansible/ansible-runner/pull/862#discussion_r724365069

ansible_runner/interface.py
```diff
@@ -34,7 +34,8 @@
 from ansible_runner.utils import (
     dump_artifacts,
     check_isolation_executable_installed,
-    santize_json_response
+    santize_json_response,
+    signal_handler
```
samdoran: Trailing commas make future diffs easier to read.
```suggestion
    signal_handler,
```

samdoran:
> how do you feel about a 20% failure rate? 

I think you may have resolved this, but merging flaky tests is only hurting us. The tests should be fixed so they are reliable.

> I started testing the newest implementation using the threading / signal library and are seeing flakiness. 

Looks like you figured out the bug in my original suggestion. I updated it but seems like that was after you had already started testing. The current implementation looks good.

> Looking for feedback re: issuing a warning to user re: not being able to set signal handler

I don't think we need a warning.

ansible_runner/utils/__init__.py
samdoran: This looks great!

samdoran:
The changes look good, but the test needs work. I'm not able to get them to pass at all on macOS with Docker. This may not be an issue with the test but an issue with `ansible-runner start` on macOS since running the same command run by the test also does nothing.

On Linux, the test does seem to pass but only if other test instances are not running. I confirmed the changes in #847 fix that issue.

test/integration/containerized/test_cli_containerized.py
```diff
@@ -29,3 +33,36 @@ def test_playbook_run(cli, skip_if_no_podman):
 def test_provide_env_var(cli, skip_if_no_podman, test_data_dir):
     r = cli(['run', os.path.join(test_data_dir, 'job_env'), '-p', 'printenv.yml'])
     assert 'gifmyvqok2' in r.stdout, r.stdout
+
+
+def test_cli_kill_cleanup(cli, test_data_dir, container_runtime_installed):
```
samdoran: The `container_runtime_installed` fixture uses the first found run time. This test should be run against all supported runtimes. We have bigger issues in that our CI is missing `podman`, but tests should be written to run against all available runtimes, not just the first found. I have a lot of existing tests to fix, but new ones should do the right thing (unless we are testing a `docker` or `podman` only feature).

```suggestion
@pytest.mark.parametrize('runtime', ('podman', 'docker'))
def test_cli_kill_cleanup(cli, test_data_dir, runtime):
    if not shutil.which(runtime):
        pytest.skip(f'{runtime}' is not installed')
```

test/integration/containerized/test_cli_containerized.py
```diff
@@ -29,3 +33,36 @@ def test_playbook_run(cli, skip_if_no_podman):
...
+    cli_args = ['start', pdd, '-p', 'sleep.yml', '--ident', ident,
+                '--process-isolation', '--process-isolation-executable', container_runtime_installed]
+    cli(cli_args)
+
+    def container_is_running():
+        r = cli([container_runtime_installed, 'ps', '-f', f'name=ansible_runner_{ident}', '--format={{.Names}}'], bare=True)
+        return ident in r.stdout
+
+    tries = 5
+    for i in range(tries):
```
Shrews: FYI, we have a method in `test/integration/test_main.py` called `iterate_timeout()` specifically for this type of thing. I think we should move that to a common area and use that here, rather than re-invent the wheel.

test/integration/containerized/test_cli_containerized.py
```diff
@@ -29,3 +33,36 @@ def test_playbook_run(cli, skip_if_no_podman):
...
+        time.sleep(1)
+    else:
+        assert container_is_running()
+
+    # Here, we will do sigterm to kill the parent process, it should handle this gracefully
+    with open(os.path.join(pdd, 'pid'), 'r') as f:
+        pid = int(f.read().strip())
+    os.kill(pid, signal.SIGTERM)
+
+    for i in range(tries):
```
Shrews: Ditto on `iterate_timeout()` comment.

jladdjr:
With _latest_ docker and podman, confirmed that unit test gives perfect signal:
```
| Runner Version | Latest Podman (v3.4.0) | Latest Docker (v20.10.9) |
|----------------|------------------------|--------------------------|
| No Fix         | 10 / 10 fails          | 10 / 10 fails            |
| Has Fix        | 10 / 10 passes         | 10 / 10 passes           |
```

jladdjr:
> I'm not able to get them to pass at all on macOS with Docker. This may not be an issue with the test but an issue with ansible-runner start on macOS since running the same command run by the test also does nothing.

@samdoran - can you provide more details on what went wrong here?

ansible_runner/interface.py
jladdjr: ✅

ansible_runner/interface.py
jladdjr: ✅

test/integration/containerized/test_cli_containerized.py
jladdjr: (done) ✅

jladdjr:
`ansible-runner-build-container-image-stable-2.9` zuul check [failing](https://dashboard.zuul.ansible.com/t/ansible/build/bab519b6f7a7477b926680e2f96c189a):

```
podman push zuul-jobs.buildset-registry:5000/quay.io/ansible/ansible-runner:stable-2.9-devel
...
Copying blob sha256:25d101049a9c1ac3507bfd2306116336482ca8964a9f29d99ba19ee9e3f1fa42
time="2021-10-12T05:07:54Z" level=warning msg="failed, retrying in 1s ... (1/3). Error: writing blob: Patch \"https://zuul-jobs.buildset-registry:5000/v2/quay.io/ansible/ansible-runner/blobs/uploads/d425b3662e234b168166ecea81645aeb\": write tcp 162.253.43.20:59688->162.253.42.209:5000: write: connection timed out"
Getting image source signatures
Error: trying to reuse blob sha256:74ddd0ec08fa43d09f32636ba91a0a3053b02cb4627c35051aff89f853606b59 at destination: pinging container registry zuul-jobs.buildset-registry:5000: Get "https://zuul-jobs.buildset-registry:5000/v2/": dial tcp 162.253.42.209:5000: i/o timeout
```

looks like maybe it was unable to reach the container registry while trying to push an image

test/integration/containerized/test_cli_containerized.py
jladdjr: ✅

test/integration/containerized/test_cli_containerized.py
jladdjr: ✅

jladdjr:
@samdoran @Shrews - all feedback addressed, let me know if there's anything else you think should be tweaked.

shanemcd (approved):

shanemcd:
recheck

test/integration/test_main.py
```diff
@@ -8,26 +8,16 @@
 import shutil
 import yaml
 import tempfile
-import time
 from contextlib import contextmanager
 import pytest
 
 
 from ansible_runner.exceptions import AnsibleRunnerException
+from test.utils.common import iterate_timeout
```
samdoran, line 16: This directory should be reserved for files that are testing `ansible_runner/utils`, not common code shared between tests. Common test code should go in the  appropriate `conftest.py` or maybe a `test/helpers/` or `test/support` directory.

test/integration/containerized/test_container_management.py
```diff
@@ -74,6 +74,10 @@ def test_cancel_will_remove_container(test_data_dir, container_runtime_installed
         cli, container_runtime_installed, 'ansible_runner_foo_bar'
     ), 'Found a running container, they should have all been stopped'
 
+    # clean up settings artifacts generated by test
+    if os.path.exists(env_dir):
+        shutil.rmtree(env_dir)
```
samdoran, line 79: If this is necessary, then the test is doing something wrong. At least add a `FIXME` note here that this can be removed once #847 is merged.

test/integration/test_main.py
samdoran: Oops, I just realized this is in the `tests/` dir, and not `test/units`. That presents some [another potential issue](https://docs.pytest.org/en/6.2.x/goodpractices.html#tests-outside-application-code) related to `pytest` and imports, but I suppose this is fine for now.

test/integration/containerized/test_cli_containerized.py
samdoran: Did this marker end up being necessary to make the test stable? It wouldn't surprise me, but that's unfortunate.

samdoran: It looks like this is needed, unfortunately. The tests wouldn't pass if run simultaneously since they are using the same data directory.

samdoran (changes requested):
The current integration test fails on macOS due to an unrelated issue with the `daemon` library (just a guess). I would rather not merge tests that do not pass locally.

I think we also need unit tests of `signal_handler()` and functional tests to make sure the signals are handled as expected.

If we decide to keep this integration test, it needs a `skipif` so it will be skipped on platforms where the tests do not work.

jladdjr:
> The current integration test fails on macOS due to an unrelated issue with the daemon library (just a guess). I would rather not merge tests that do not pass locally.

Can you provide some more details on the failure? How are you invoking the test? I needed to run with `pytest -n 0` to ensure serial execution. I believe this is how zuul invokes serial tests. What specific pytest failure are you seeing? You mention a daemon library, so I'm assuming this fails for docker? Are you seeing failures for both container runtimes? Are you running the latest version of docker / podman (this made a significant difference for me). When the failure happens, what is the current state of any running containers? (I typically run a `watch <container runtime> ps` while the test is running so I can also observe what is happening).

The two main points where I've seen an assertion fail in the test fail are:
1. Test fails because it is not seeing a container spun up when runner is first invoked.
2. Container is successfully started but fails to exit when we send a signal to the ansible-runner process.

@AlanCoding I believe that you are running with a docker for mac setup. Can you confirm if the tests fail for you when the fix is in place?

jladdjr:
@samdoran has mentioned that the macos failures are likely unrelated to this PR. Given that this fix addresses a significant error we're seeing in awx (cancel a job, job continues running in EE container), recommend we get this in as soon as we can and not block on existing / unrelated failures.

samdoran:
Running `tox -e py test/integration/containerized/test_cli_containerized.py -- -n0` results in the following failure:

```
FAILED test/integration/containerized/test_cli_containerized.py::test_cli_kill_cleanup[docker] - Exception: Timeout waiting for confirm ansible-runner started container
```

<details>

```
> tox -e py test/integration/containerized/test_cli_containerized.py -- -n0
/Users/sdoran/Source/ansible-runner/.tox/py/lib/python3.10/site-packages/setuptools/dist.py:487: UserWarning: Normalizing '2.0.0.0alpha4.dev249' to '2.0.0.0a4.dev249'
  warnings.warn(tmpl.format(**locals()))
py develop-inst-noop: /Users/sdoran/Source/ansible-runner
py installed: ansible-core==2.11.5,-e git+ssh://git@github.com/ansible/ansible-runner.git@34d0e7639cec4e77a6645058fa67a67f84b8e9a8#egg=ansible_runner,attrs==21.2.0,cffi==1.14.6,cryptography==35.0.0,docutils==0.17.1,execnet==1.9.0,flake8==3.9.2,iniconfig==1.1.1,Jinja2==3.0.2,lockfile==0.12.2,MarkupSafe==2.0.1,mccabe==0.6.1,packaging==21.0,pathspec==0.9.0,pexpect==4.8.0,pluggy==1.0.0,ptyprocess==0.7.0,py==1.10.0,pycodestyle==2.7.0,pycparser==2.20,pyflakes==2.3.1,pyparsing==2.4.7,pytest==6.2.5,pytest-forked==1.3.0,pytest-mock==3.6.1,pytest-repeat==0.9.1,pytest-timeout==1.4.2,pytest-xdist==2.4.0,python-daemon==2.3.0,PyYAML==5.4.1,resolvelib==0.5.4,six==1.16.0,toml==0.10.2,yamllint==1.26.3
py run-test-pre: PYTHONHASHSEED='1417195590'
py run-test: commands[0] | pytest -m 'not serial' test/integration/containerized/test_cli_containerized.py -n0
====================================================================================== test session starts =======================================================================================
platform darwin -- Python 3.10.0, pytest-6.2.5, py-1.10.0, pluggy-1.0.0 -- /Users/sdoran/Source/ansible-runner/.tox/py/bin/python
cachedir: .tox/py/.pytest_cache
rootdir: /Users/sdoran/Source/ansible-runner, configfile: pytest.ini
plugins: mock-3.6.1, xdist-2.4.0, repeat-0.9.1, timeout-1.4.2, forked-1.3.0
collected 5 items / 5 deselected

===================================================================================== 5 deselected in 0.02s ======================================================================================
ERROR: InvocationError for command /Users/sdoran/Source/ansible-runner/.tox/py/bin/pytest -m 'not serial' test/integration/containerized/test_cli_containerized.py -n0 (exited with code 5)
____________________________________________________________________________________________ summary _____________________________________________________________________________________________
ERROR:   py: commands failed
sdoran@mbp-2018 ~/S/ansible-runner (PR/862) [1]> vim tox.ini
sdoran@mbp-2018 ~/S/ansible-runner (PR/862)> tox -e py test/integration/containerized/test_cli_containerized.py -- -n0
/Users/sdoran/Source/ansible-runner/.tox/py/lib/python3.10/site-packages/setuptools/dist.py:487: UserWarning: Normalizing '2.0.0.0alpha4.dev249' to '2.0.0.0a4.dev249'
  warnings.warn(tmpl.format(**locals()))
py develop-inst-noop: /Users/sdoran/Source/ansible-runner
py installed: ansible-core==2.11.5,-e git+ssh://git@github.com/ansible/ansible-runner.git@34d0e7639cec4e77a6645058fa67a67f84b8e9a8#egg=ansible_runner,attrs==21.2.0,cffi==1.14.6,cryptography==35.0.0,docutils==0.17.1,execnet==1.9.0,flake8==3.9.2,iniconfig==1.1.1,Jinja2==3.0.2,lockfile==0.12.2,MarkupSafe==2.0.1,mccabe==0.6.1,packaging==21.0,pathspec==0.9.0,pexpect==4.8.0,pluggy==1.0.0,ptyprocess==0.7.0,py==1.10.0,pycodestyle==2.7.0,pycparser==2.20,pyflakes==2.3.1,pyparsing==2.4.7,pytest==6.2.5,pytest-forked==1.3.0,pytest-mock==3.6.1,pytest-repeat==0.9.1,pytest-timeout==1.4.2,pytest-xdist==2.4.0,python-daemon==2.3.0,PyYAML==5.4.1,resolvelib==0.5.4,six==1.16.0,toml==0.10.2,yamllint==1.26.3
py run-test-pre: PYTHONHASHSEED='999930756'
py run-test: commands[0] | pytest -n 0 -m serial test/integration/containerized/test_cli_containerized.py -n0
====================================================================================== test session starts =======================================================================================
platform darwin -- Python 3.10.0, pytest-6.2.5, py-1.10.0, pluggy-1.0.0 -- /Users/sdoran/Source/ansible-runner/.tox/py/bin/python
cachedir: .tox/py/.pytest_cache
rootdir: /Users/sdoran/Source/ansible-runner, configfile: pytest.ini
plugins: mock-3.6.1, xdist-2.4.0, repeat-0.9.1, timeout-1.4.2, forked-1.3.0
collected 5 items

test/integration/containerized/test_cli_containerized.py::test_module_run SKIPPED (podman container runtime(s) not available)                                                              [ 20%]
test/integration/containerized/test_cli_containerized.py::test_playbook_run SKIPPED (podman container runtime(s) not available)                                                            [ 40%]
test/integration/containerized/test_cli_containerized.py::test_provide_env_var SKIPPED (podman container runtime(s) not available)                                                         [ 60%]
test/integration/containerized/test_cli_containerized.py::test_cli_kill_cleanup[podman] SKIPPED (podman is unavailable)                                                                    [ 80%]
test/integration/containerized/test_cli_containerized.py::test_cli_kill_cleanup[docker] FAILED                                                                                             [100%]

============================================================================================ FAILURES ============================================================================================
_________________________________________________________________________________ test_cli_kill_cleanup[docker] __________________________________________________________________________________

cli = <function cli.<locals>.run at 0x106f85480>, runtime = 'docker', test_data_dir = '/Users/sdoran/Source/ansible-runner/test/data'

    @pytest.mark.serial
    @pytest.mark.parametrize('runtime', ['podman', 'docker'])
    def test_cli_kill_cleanup(cli, runtime, test_data_dir):
        if shutil.which(runtime) is None:
            pytest.skip(f'{runtime} is unavailable')

        unique_string = str(uuid4()).replace('-', '')
        ident = f'kill_test_{unique_string}'
        pdd = os.path.join(test_data_dir, 'sleep')
        cli_args = ['start', pdd, '-p', 'sleep.yml', '--ident', ident,
                    '--process-isolation', '--process-isolation-executable', runtime]
        cli(cli_args)

        def container_is_running():
            r = cli([runtime, 'ps', '-f', f'name=ansible_runner_{ident}', '--format={{.Names}}'], bare=True)
            return ident in r.stdout

        timeout = 10
>       for _ in iterate_timeout(timeout, 'confirm ansible-runner started container', interval=1):

_          = 9
cli        = <function cli.<locals>.run at 0x106f85480>
cli_args   = ['start', '/Users/sdoran/Source/ansible-runner/test/data/sleep', '-p', 'sleep.yml', '--ident', 'kill_test_a1dbe8595a5a485c8a0c77d665eddb78', ...]
container_is_running = <function test_cli_kill_cleanup.<locals>.container_is_running at 0x106f85510>
ident      = 'kill_test_a1dbe8595a5a485c8a0c77d665eddb78'
pdd        = '/Users/sdoran/Source/ansible-runner/test/data/sleep'
runtime    = 'docker'
test_data_dir = '/Users/sdoran/Source/ansible-runner/test/data'
timeout    = 10
unique_string = 'a1dbe8595a5a485c8a0c77d665eddb78'

test/integration/containerized/test_cli_containerized.py:57:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

max_seconds = 10, purpose = 'confirm ansible-runner started container', interval = 1

    def iterate_timeout(max_seconds, purpose, interval=2):
        start = time.time()
        count = 0
        while (time.time() < start + max_seconds):
            count += 1
            yield count
            time.sleep(interval)
>       raise Exception("Timeout waiting for %s" % purpose)
E       Exception: Timeout waiting for confirm ansible-runner started container

count      = 9
interval   = 1
max_seconds = 10
purpose    = 'confirm ansible-runner started container'
start      = 1634055424.229028

test/utils/common.py:11: Exception
======================================================================================== warnings summary ========================================================================================
test/integration/containerized/test_cli_containerized.py::test_module_run
  /Users/sdoran/Source/ansible-runner/test/integration/conftest.py:41: UserWarning: podman not available
    warnings.warn(UserWarning(f"{runtime} not available"))

-- Docs: https://docs.pytest.org/en/stable/warnings.html
====================================================================================== slowest 10 durations ======================================================================================
11.11s call     test/integration/containerized/test_cli_containerized.py::test_cli_kill_cleanup[docker]

(9 durations < 1s hidden.  Use -vv to show these durations.)
==================================================================================== short test summary info =====================================================================================
SKIPPED [3] test/integration/containerized/test_cli_containerized.py:17: podman container runtime(s) not available
SKIPPED [1] test/integration/containerized/test_cli_containerized.py:43: podman is unavailable
FAILED test/integration/containerized/test_cli_containerized.py::test_cli_kill_cleanup[docker] - Exception: Timeout waiting for confirm ansible-runner started container
============================================================================ 1 failed, 4 skipped, 1 warning in 11.88s ============================================================================
ERROR: InvocationError for command /Users/sdoran/Source/ansible-runner/.tox/py/bin/pytest -n 0 -m serial test/integration/containerized/test_cli_containerized.py -n0 (exited with code 1)
____________________________________________________________________________________________ summary _____________________________________________________________________________________________
ERROR:   py: commands failed
```

AlanCoding:
I'm on Fedora 34 with both podman and docker installed, checking out current branch I'm able to run the docker test.

```
$ py.test test/ -n0 test/integration/containerized/test_cli_containerized.py -k test_cli_kill_cleanup[docker]
==================================================== test session starts ====================================================
platform linux -- Python 3.9.6, pytest-6.2.5, py-1.9.0, pluggy-0.13.1 -- /home/alancoding/repos/awx/env/bin/python3
cachedir: .pytest_cache
benchmark: 3.2.3 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
ansible: 2.13.0.dev0
rootdir: /home/alancoding/repos/ansible-runner, configfile: pytest.ini
plugins: forked-1.3.0, github-0.3.0, ordering-0.6, xdist-2.2.0, mock-3.6.1, cov-2.12.1, ansible-2.2.4, benchmark-3.2.3
collecting ... collected 0 github issues
collected 2041 items / 2039 deselected / 2 selected                                                                         

test/integration/containerized/test_cli_containerized.py::test_cli_kill_cleanup[docker] PASSED                        [ 50%]
test/integration/containerized/test_cli_containerized.py::test_cli_kill_cleanup[docker] PASSED                        [ 50%]

=================================================== slowest 10 durations ====================================================
6.90s call     test/integration/containerized/test_cli_containerized.py::test_cli_kill_cleanup[docker]
5.76s call     test/integration/containerized/test_cli_containerized.py::test_cli_kill_cleanup[docker]

(4 durations < 1s hidden.  Use -vv to show these durations.)
============================================ 2 passed, 2039 deselected in 13.23s ============================================
```

Failing waiting for the container to start sounds like it might be either a timing issue, or something more basic isn't working.

samdoran:
Yup, it works fine on Linux run serially.

Shrews:
FYI, any test that uses `ansible-runner start`, which forks the main process using the `python-daemon` library, and tries to use a subprocess (e.g., `subprocess.Popen()`) from within the new forked process to do anything is just not going to work on a Mac. It seems to start the container ok because it defaults to using `pexpect.spawn()`, but anything else, like killing the running container (which uses `subprocess.Popen()`) is likely to not work. For that reason, we should skip such tests on OSX.

jladdjr:
> is just not going to work on a Mac.

@Shrews do you know what it is about OS X that isn't supporting the use of the libraries / calls you mentioned?

samdoran (approved):

test/unit/test_utils.py
```diff
@@ -353,3 +355,59 @@ def test_transmit_permissions(tmp_path, fperm):
...
+    assert mock_signal.call_args_list[1][0][0] == signal.SIGINT
+
+
+def test_signal_handler_outside_main_thread(mocker):
+    """Test that the default handler will not try to set signal handlers if not in the main thread"""
+
+    mocker.patch('ansible_runner.utils.threading.main_thread', return_value='thread0')
+    mocker.patch('ansible_runner.utils.threading.current_thread', return_value='thread1')
+
+    assert signal_handler() is None
```
jladdjr, line 389: 👍

test/unit/test_utils.py
```diff
@@ -353,3 +355,59 @@ def test_transmit_permissions(tmp_path, fperm):
...
+            return self._is_set
+
+    mocker.patch('ansible_runner.utils.threading.main_thread', return_value='thread0')
+    mocker.patch('ansible_runner.utils.threading.current_thread', return_value='thread0')
+    mocker.patch('ansible_runner.utils.threading.Event', MockEvent)
+    mock_signal = mocker.patch('ansible_runner.utils.signal.signal')
+
+    assert signal_handler()() is False
+    assert mock_signal.call_args_list[0][0][0] == signal.SIGTERM
+    assert mock_signal.call_args_list[1][0][0] == signal.SIGINT
```
jladdjr, line 380: 👍

test/unit/test_utils.py
```diff
@@ -353,3 +355,59 @@ def test_transmit_permissions(tmp_path, fperm):
...
+            return self._is_set
+
+    mocker.patch('ansible_runner.utils.threading.main_thread', return_value='thread0')
+    mocker.patch('ansible_runner.utils.threading.current_thread', return_value='thread0')
+    mocker.patch('ansible_runner.utils.threading.Event', MockEvent)
+    mock_signal = mocker.patch('ansible_runner.utils.signal.signal')
+
+    signal_handler()
+
+    with pytest.raises(AttributeError, match='Raised intentionally'):
```
jladdjr, line 412: 👍

PR #865, opened 2021-10-12
Prevent deletion of both prohibited paths and their resolved targets

samdoran:
If the path is a symlink, only the symlinked path was protected from being deleted, not the target the link pointed to.

Add tests to cover this scenario.


AlanCoding (approved):
subtle, good call.

Shrews (approved):

PR #887, opened 2021-10-25
Keep worker --private-data-dir when --delete not given

AlanCoding:
I messed up our requirements a little bit with https://github.com/ansible/ansible-runner/pull/820, and now I'm getting reports that people _want_ the private_data_dir to be kept on the node running the `ansible-runner worker` command, and it is (unwantedly) getting deleted at the end of the run.

If you look at how I wrote the docs, I was dodgy about this case. I wrote it so that it would get cleaned up if it was a _new_ directory. That didn't even work for the AWX use case, for which there is an intent to respect the setting `AWX_CLEANUP_PATHS` on both control and worker nodes.

I will take out of draft when I have confirmed this works with a production install.

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -792,10 +792,12 @@ def main(sys_args=None):
             print(safe_dump(info, default_flow_style=True))
             parser.exit(0)
 
-        if vargs.get('private_data_dir') and vargs.get('delete_directory', False):
-            cleanup_folder(vargs['private_data_dir'])
-            register_for_cleanup(vargs['private_data_dir'])
-        elif not vargs.get('private_data_dir'):
+        private_data_dir = vargs.get('private_data_dir')
+        delete_directory = vargs.get('delete_directory', False)
+        if private_data_dir and delete_directory:
+            shutil.rmtree(private_data_dir, True)
+            register_for_cleanup(private_data_dir)removed?
+        elif private_data_dir is None:
             temp_private_dir = tempfile.mkdtemp()
             vargs['private_data_dir'] = temp_private_dir
             register_for_cleanup(temp_private_dir)
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -28,7 +28,6 @@ import errno
 import json
 import stat
 import os
-import shutil
 import textwrap
 import tempfile
 
@@ -40,7 +39,7 @@ from yaml import safe_dump, safe_load
 from ansible_runner import run
 from ansible_runner import output
 from ansible_runner import cleanup
-from ansible_runner.utils import dump_artifact, Bunch, register_for_cleanup
+from ansible_runner.utils import dump_artifact, Bunch, register_for_cleanup, cleanup_folder
 from ansible_runner.utils.capacity import get_cpu_count, get_mem_in_bytes, ensure_uuid
 from ansible_runner.runner import Runner
 from ansible_runner.exceptions import AnsibleRunnerException
@@ -504,7 +503,7 @@ def role_manager(vargs):
     if vargs.get('role'):
         if not project_exists and os.path.exists(project_path):
             logger.debug('removing dynamically generated project folder')
-            shutil.rmtree(project_path)
+            cleanup_folder(project_path)
         elif playbook and os.path.isfile(playbook):
             logger.debug('removing dynamically generated playbook')
             os.remove(playbook)
@@ -521,7 +520,7 @@ def role_manager(vargs):
         # since ansible-runner created the env folder, remove it
         if not env_exists and os.path.exists(env_path):
             logger.debug('removing dynamically generated env folder')
-            shutil.rmtree(env_path)
+            cleanup_folder(env_path)
 
 
 def print_common_usage():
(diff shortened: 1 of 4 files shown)


AlanCoding:
I think I have verified this sufficiently for myself now.

docs/remote_jobs.rst
```diff
@@ -38,8 +38,8 @@ private data directory or artifacts, because these are how the user interacts wi
 
 When running `ansible-runner worker`, if no `--private-data-dir` is given,
 it will extract the contents to a temporary directory which is deleted at the end of execution.
-You can use the `--delete` flag in conjunction with `--private-data-dir` to assure that
-the provided directory is deleted at the end of execution.
+If the `--private-data-dir` option is given, then the directory will persist after the run finishes,
+unless you give the `--delete` flag, in which case the directory will be deleted both before and after the run.
```
samdoran: RST uses double backticks for inline code formatting. I clarified the wording a bit, if I'm understanding the behavior correctly.
```suggestion
If the ``--private-data-dir`` option is given, then the directory will persist after the run finishes
unless the ``--delete`` flag is also set. In that case, the private data directory will be deleted before execution if it exists and also removed after execution.
```

ansible_runner/__main__.py
```diff
@@ -504,7 +503,7 @@ def role_manager(vargs):
     if vargs.get('role'):
         if not project_exists and os.path.exists(project_path):
             logger.debug('removing dynamically generated project folder')
-            shutil.rmtree(project_path)
+            cleanup_folder(project_path)
```
samdoran: Is `cleanup_folder` providing more functionality here than `shutil.rmtree(project_path, ignore_errors=True`? There are a couple of places where we use the return value, but not here.

ansible_runner/__main__.py
```diff
@@ -521,7 +520,7 @@ def role_manager(vargs):
         # since ansible-runner created the env folder, remove it
         if not env_exists and os.path.exists(env_path):
             logger.debug('removing dynamically generated env folder')
-            shutil.rmtree(env_path)
+            cleanup_folder(env_path)
```
samdoran: Same comment as above.

Shrews (approved):
LGTM. Make sure this is backported to the appropriate version(s).

ansible_runner/__main__.py
```diff
@@ -793,18 +792,13 @@ def main(sys_args=None):
...
-            if os.path.exists(vargs.get('private_data_dir')):
-                if vargs.get('delete_directory', False):
-                    shutil.rmtree(vargs['private_data_dir'])
-                else:
-                    cleanup_data_dir = False
-        else:
+        if vargs.get('private_data_dir') and vargs.get('delete_directory', False):
+            cleanup_folder(vargs['private_data_dir'])
+            register_for_cleanup(vargs['private_data_dir'])
+        elif not vargs.get('private_data_dir'):
```
samdoran: For clarity, store the values. This would be a great place to use an assignment expression (`:=`) but I don't think we're ready to make Python 3.8 a hard requirement yet.
```suggestion
        private_data_dir = vargs.get('private_data_dir')
        delete_directory = vargs.get('delete_directory', False)
        if private_data_dir and delete_directory:
            shutil.rmtree(private_data_dir, True)
            register_for_cleanup(private_data_dir)removed?
        elif private_data_dir is None:
```

samdoran: I'm not sure about the call to `register_for_cleanup()` after removing the directory. Does it get recreated and this is to ensure the directory is removed before it is later created? It's probably worth putting a comment in here explaining the reason behind the code.

test/integration/test_transmit_worker_process.py
```diff
@@ -207,6 +207,28 @@ def test_worker_preserve_or_delete_dir(tmp_path, cli, transmit_stream, delete):
             assert test_path.exists()
 
 
+@pytest.mark.parametrize('delete', [False, True])
```
samdoran: This should be broken into separate tests to avoid the use of conditional asserts, making the tests easier to read:
- no `--private-data-dir` or `--delete`
- only `--private-data-dir`
- both `--private-data-dir` and `--delete`

We should also have unit tests to make sure `shutil`, `register_for_cleanup`, and `cleanup_folder` are called as expected.

samdoran (changes requested):

docs/remote_jobs.rst
AlanCoding: looks great, the only thing I might add is putting () around (if it exists) to help readability

ansible_runner/__main__.py
AlanCoding: that's true, I didn't know to use `ignore_errors=True`

ansible_runner/__main__.py
AlanCoding: These two lines track with what was said in the docs. If the directory exists before the run, it deletes it, and it also deletes it after the run. So, deletion before and after. The run will _always_ create the directory, because ansible-runner works via a private data dir.

test/integration/test_transmit_worker_process.py
AlanCoding: I could see the argument for re-organizing the parametrize. But there's a lot of different directions I could take your feedback in, and I don't want to create a bunch of redundant test content. So I'll try what makes sense to me and go from there.

samdoran: Would you mind if I added tests to this PR?

samdoran:
Docs build failure is unrelated to this PR (see #888).

test/integration/test_transmit_worker_process.py
AlanCoding: no problem at all.

samdoran:
Sorry I haven't made any progress here. I've been trying to fix the docs build all last week and am fighting CI issues this morning. This PR will need to be rebased once the docs build fixes land since RTD does not do speculative merges when running tests.

AlanCoding:
Are the aforementioned fixes for the docs build https://github.com/ansible/ansible-runner/pull/894?

samdoran:
Yes.

samdoran:
Looks like we are having CI infrastructure issues, so I'm not sure when we be able to get changes merged.

samdoran:
I added the tests we needed. If all goes well, we can merge this.

samdoran (approved):

AlanCoding:
Looks good to me :+1: thanks!

AlanCoding:
I'm fine to squash all commits, but I don't know what's really going on with the bot.

PR #894, opened 2021-10-29
Set version in docs configuration

samdoran:
Also update copyright information.

The EPUB build raises a warning when no version is provided. Since I changed the RTD configuration to treat warnings as errors it fails the actual build. Evidently the EPUB build is not done for PRs, so the failure only showed up after merging.

It may be a good idea to add an EPUB build to the `docs` environment.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,8 +12,11 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
+import datetime
+import importlib.metadata
 import os
 import sys
+
 sys.path.insert(0, os.path.abspath('../'))
 sys.path.insert(0, os.path.abspath('.'))
 sys.path.insert(0, os.path.abspath('./'))
@@ -21,9 +24,16 @@ sys.path.insert(0, os.path.abspath('./'))
 
 # -- Project information -----------------------------------------------------
 
+def _get_version():
+    version_parts = importlib.metadata.version('ansible-runner').split('.', 3)[:3]
+
+    return '.'.join(version_parts)
+
+
 project = 'ansible-runner'
-copyright = '2018, Red Hat Ansible'
-author = 'Red Hat Ansible'
+copyright = f'2018-{datetime.datetime.today().year}, Red Hat, Inc'
+author = 'Red Hat, Inc.'
+version = _get_version()
 
 # -- General configuration ---------------------------------------------------
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,18 +13,14 @@
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
 import datetime
-import importlib.metadata
-import sys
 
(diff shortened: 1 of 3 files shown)


shanemcd (approved):

docs/conf.py
```diff
@@ -12,18 +12,23 @@
...
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-import os
-import sys
-sys.path.insert(0, os.path.abspath('../'))
-sys.path.insert(0, os.path.abspath('.'))
-sys.path.insert(0, os.path.abspath('./'))
+import datetime
+
+from ansible_runner.__main__ import VERSION
```
webknjaz, line 17: If it's expected to be installed, just use `importlib.metadata.version('ansible-runner')` instead of importing it.

docs/conf.py
```diff
@@ -12,18 +12,23 @@
...
+def _get_version():
+    version_parts = VERSION.split('.', 3)[:3]
+
+    return '.'.join(version_parts)
+
+
 project = 'ansible-runner'
-copyright = '2018, Red Hat Ansible'
-author = 'Red Hat Ansible'
+copyright = f'2018-{datetime.datetime.today().year}, Red Hat, Inc'
```
webknjaz, line 29: Usually, the copyright year shows the start of the content creation and the follow-up year updates demonstrate when the updates have been made. The current year is not always the year of an update, just rebuild. So I'm not sure if it makes sense to use it.

Besides, according to the Linux Foundation, it's reasonable to drop this completely since the authors preserve the copyright regardless of the presence of the copyright notice which is hard to track/update over time: https://linuxfoundation.org/blog/copyright-notices-in-open-source-software-projects/.

P.S. If you reorder the `author` and `copyright` vars, you could use the former in the f-string too.

AlanCoding (approved):

docs/conf.py
samdoran: I was doing that originally, but ran into an issue because we use Python 3.6 by default in CI. I think I could get around it by setting `basepython` for the `tox` environment, but this seemed easier (and it's a bit of dogfooding our API for getting version information).

docs/conf.py
samdoran: Copyright stuff is really fuzzy. I mainly was trying to make the site not look stale since old copyright years can (incorrectly) be interpreted as "this site hasn't been updated in a while".

I read through Red Hat's guidance on copyright notices and it sounded like the year should be the year any content was created. I figured so long as we are adding _something_ new in a given calendar year, the dates will be accurate. But that is a bit fuzzy, for sure.

samdoran: > P.S. If you reorder the author and copyright vars, you could use the former in the f-string too.

The minor annoyance I ran into with that is the trailing `.` is added at the end of the copyright notice. I don't want the `.` to exists in the copyright notice since it would render as `2018-2021 Red Hat, Inc..`, but I _do_ want it in the `author` string.

PR #901, opened 2021-11-01
Stop using `/tmp` as a private data dir in tests

AlanCoding:
When I run the tests locally, I wind up with a lot of junk in `/tmp`

This isn't _so_ bad when the directories created are `/tmp/.ansible-runner-j6kn22dk/`. At least I know what it is, and anyway, we could develop a catch-all solution in the future to clean these up if we wanted.

On the other hand, `/tmp/artifacts` is not really forgivable. Inherently, that means the tests are not isolated (not everything is done in that `artifacts/<ident>` dir).

This PR seeks to stop directory creation inside of `/tmp` where if `/tmp` is the private data directory for the job being tested. To clarify - I don't have a problem if something is _mocked_ with a `/tmp` location, or some function output gives such a location. However, the initialization of `BaseConfig` is not read-only, and it went around creating completely non-isolated directories.

### DIFF
--- a/test/unit/config/test__base.py
+++ b/test/unit/config/test__base.py
@@ -223,35 +223,32 @@ def test_prepare_with_ssh_key(mocker, tmp_path):
 
 
 def test_wrap_args_with_ssh_agent_defaults(tmp_path):
-    tp = tmp_path.as_posix()
-    rc = BaseConfig(private_data_dir=tp)
-    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], f'{tp}/sshkey')
+    rc = BaseConfig(private_data_dir=str(tmp_path))
+    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], f'{tmp_path}/sshkey')
     assert res == [
         'ssh-agent',
         'sh', '-c',
-        f"trap 'rm -f {tp}/sshkey' EXIT && ssh-add {tp}/sshkey && rm -f {tp}/sshkey && ansible-playbook main.yaml"
+        f"trap 'rm -f {tmp_path}/sshkey' EXIT && ssh-add {tmp_path}/sshkey && rm -f {tmp_path}/sshkey && ansible-playbook main.yaml"
     ]
 
 
 def test_wrap_args_with_ssh_agent_with_auth(tmp_path):
-    tp = tmp_path.as_posix()
-    rc = BaseConfig(private_data_dir=tp)
-    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], f'{tp}/sshkey', f'{tp}/sshauth')
+    rc = BaseConfig(private_data_dir=str(tmp_path))
+    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], f'{tmp_path}/sshkey', f'{tmp_path}/sshauth')
     assert res == [
-        'ssh-agent', '-a', f'{tp}/sshauth',
+        'ssh-agent', '-a', f'{tmp_path}/sshauth',
         'sh', '-c',
-        f"trap 'rm -f {tp}/sshkey' EXIT && ssh-add {tp}/sshkey && rm -f {tp}/sshkey && ansible-playbook main.yaml"
+        f"trap 'rm -f {tmp_path}/sshkey' EXIT && ssh-add {tmp_path}/sshkey && rm -f {tmp_path}/sshkey && ansible-playbook main.yaml"
     ]
 
 
 def test_wrap_args_with_ssh_agent_silent(tmp_path):
-    tp = tmp_path.as_posix()
-    rc = BaseConfig(private_data_dir=tp)
-    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], f'{tp}/sshkey', silence_ssh_add=True)
+    rc = BaseConfig(private_data_dir=str(tmp_path))
+    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], f'{tmp_path}/sshkey', silence_ssh_add=True)
     assert res == [
         'ssh-agent',
         'sh', '-c',
-        f"trap 'rm -f {tp}/sshkey' EXIT && ssh-add {tp}/sshkey 2>/dev/null && rm -f {tp}/sshkey && ansible-playbook main.yaml"
+        f"trap 'rm -f {tmp_path}/sshkey' EXIT && ssh-add {tmp_path}/sshkey 2>/dev/null && rm -f {tmp_path}/sshkey && ansible-playbook main.yaml"
     ]
 
 
--- a/test/unit/config/test__base.py
+++ b/test/unit/config/test__base.py
@@ -52,6 +52,10 @@ def test_base_config_with_artifact_dir(tmp_path, patch_private_data_dir):
     assert rc.private_data_dir.startswith(base_private_data_dir)
     assert len(rc.private_data_dir) > len(base_private_data_dir)
 
+    rc._prepare_env()
+    assert not tmp_path.joinpath('artifacts').exists()
+    assert tmp_path.joinpath('this-is-some-dir').exists()
+
 
 def test_base_config_init_with_ident(tmp_path):
     rc = BaseConfig(private_data_dir=tmp_path.as_posix(), ident='test')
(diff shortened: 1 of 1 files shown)


samdoran:
> This isn't so bad when the directories created are /tmp/.ansible-runner-j6kn22dk/. At least I know what it is, and anyway, we could develop a catch-all solution in the future to clean these up if we wanted.

I figured this was a default behavior of runner and didn't take the time to figure out how to mock it out. Ideally it would be using `tmp_path` instead of `/tmp`.

> On the other hand, /tmp/artifacts is not really forgivable. Inherently, that means the tests are not isolated (not everything is done in that artifacts/<ident> dir).

I thought I got all of those. Evidently not...

test/unit/config/test__base.py
```diff
@@ -61,15 +65,15 @@ def test_base_config_init_with_ident(tmp_path):
...
     assert rc.project_dir == '/another/path'
-    rc = BaseConfig(private_data_dir='/tmp')
-    assert rc.project_dir == '/tmp/project'
+    rc = BaseConfig(private_data_dir=tmp_path.as_posix())
+    assert rc.project_dir == tmp_path.joinpath('project').as_posix()
 
 
 def test_prepare_environment_vars_only_strings(mocker):
-    rc = BaseConfig(private_data_dir="/tmp", envvars=dict(D='D'))
+    rc = BaseConfig(envvars=dict(D='D'))
```
AlanCoding, line 76: In these cases, the patch will let it fall back to using `/tmp/.ansible-runner-*`, which is quite a lot better.

test/unit/config/test__base.py
```diff
@@ -88,7 +92,7 @@ def test_prepare_environment_vars_only_strings(mocker):
 
 
 def test_prepare_environment_pexpect_defaults():
-    rc = BaseConfig(private_data_dir="/tmp")
+    rc = BaseConfig()
```
samdoran, line 95: Won't this still result in files being written directly to `/tmp`? I think it'd be better to use `tmp_path` since `pytest` cleans those up automatically and it keeps all the test artifacts in one directory.
```python
def test_prepare_environment_pexpect_defaults(tmp_path):
     rc = BaseConfig(private_data_dir=str(tmp_path))
```

test/unit/config/test__base.py
```diff
@@ -204,50 +205,53 @@ def test_prepare_env_ansible_vars(mocker):
...
-    assert rc.ssh_key_path == '/tmp/artifact/ssh_key_data'
+    assert rc.ssh_key_path == custom_artifacts.joinpath('ssh_key_data').as_posix()
     assert open_fifo_write_mock.called
 
 
-def test_wrap_args_with_ssh_agent_defaults():
-    rc = BaseConfig(private_data_dir='/tmp')
-    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], '/tmp/sshkey')
+def test_wrap_args_with_ssh_agent_defaults(tmp_path):
+    tp = tmp_path.as_posix()
```
samdoran: This doesn't need to be converted to a string prematurely. The `__str__` method is called when a `Path` object used in an `f` string. 
```suggestion
    tp = tmp_path
```

test/unit/config/test__base.py
```diff
@@ -204,50 +205,53 @@ def test_prepare_env_ansible_vars(mocker):
...
+    rc = BaseConfig(private_data_dir=tp)
+    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], f'{tp}/sshkey')
```
samdoran:
```suggestion
    rc = BaseConfig(private_data_dir=str(tmp_path))
    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], f'{tmp_path}/sshkey')
```

test/unit/config/test__base.py
```diff
@@ -204,50 +205,53 @@ def test_prepare_env_ansible_vars(mocker):
...
     assert res == [
         'ssh-agent',
         'sh', '-c',
-        "trap 'rm -f /tmp/sshkey' EXIT && ssh-add /tmp/sshkey && rm -f /tmp/sshkey && ansible-playbook main.yaml"
+        f"trap 'rm -f {tp}/sshkey' EXIT && ssh-add {tp}/sshkey && rm -f {tp}/sshkey && ansible-playbook main.yaml"
```
samdoran:
```suggestion
        f"trap 'rm -f {tmp_path}/sshkey' EXIT && ssh-add {tmp_path}/sshkey && rm -f {tmp_path}/sshkey && ansible-playbook main.yaml"
```

test/unit/config/test__base.py
```diff
@@ -204,50 +205,53 @@ def test_prepare_env_ansible_vars(mocker):
...
     ]
 
 
-def test_wrap_args_with_ssh_agent_with_auth():
-    rc = BaseConfig(private_data_dir='/tmp')
-    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], '/tmp/sshkey', '/tmp/sshauth')
+def test_wrap_args_with_ssh_agent_with_auth(tmp_path):
+    tp = tmp_path.as_posix()
+    rc = BaseConfig(private_data_dir=tp)
+    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], f'{tp}/sshkey', f'{tp}/sshauth')
```
samdoran:
```suggestion
    rc = BaseConfig(private_data_dir=str(tmp_path))
    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], f'{tmp_path}/sshkey', f'{tmp_path}/sshauth')
```

test/unit/config/test__base.py
```diff
@@ -204,50 +205,53 @@ def test_prepare_env_ansible_vars(mocker):
...
     assert res == [
-        'ssh-agent', '-a', '/tmp/sshauth',
+        'ssh-agent', '-a', f'{tp}/sshauth',
```
samdoran:
```suggestion
        'ssh-agent', '-a', f'{tmp_path}/sshauth',
```

test/unit/config/test__base.py
```diff
@@ -204,50 +205,53 @@ def test_prepare_env_ansible_vars(mocker):
...
         'sh', '-c',
-        "trap 'rm -f /tmp/sshkey' EXIT && ssh-add /tmp/sshkey && rm -f /tmp/sshkey && ansible-playbook main.yaml"
+        f"trap 'rm -f {tp}/sshkey' EXIT && ssh-add {tp}/sshkey && rm -f {tp}/sshkey && ansible-playbook main.yaml"
```
samdoran:
```suggestion
        f"trap 'rm -f {tmp_path}/sshkey' EXIT && ssh-add {tmp_path}/sshkey && rm -f {tmp_path}/sshkey && ansible-playbook main.yaml"
```

test/unit/config/test__base.py
```diff
@@ -204,50 +205,53 @@ def test_prepare_env_ansible_vars(mocker):
...
     ]
 
 
-def test_wrap_args_with_ssh_agent_silent():
-    rc = BaseConfig(private_data_dir='/tmp')
-    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], '/tmp/sshkey', silence_ssh_add=True)
+def test_wrap_args_with_ssh_agent_silent(tmp_path):
+    tp = tmp_path.as_posix()
+    rc = BaseConfig(private_data_dir=tp)
+    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], f'{tp}/sshkey', silence_ssh_add=True)
```
samdoran:
```suggestion
    rc = BaseConfig(private_data_dir=str(tmp_path))
    res = rc.wrap_args_with_ssh_agent(['ansible-playbook', 'main.yaml'], f'{tmp_path}/sshkey', silence_ssh_add=True)
```

test/unit/config/test__base.py
```diff
@@ -204,50 +205,53 @@ def test_prepare_env_ansible_vars(mocker):
...
     assert res == [
         'ssh-agent',
         'sh', '-c',
-        "trap 'rm -f /tmp/sshkey' EXIT && ssh-add /tmp/sshkey 2>/dev/null && rm -f /tmp/sshkey && ansible-playbook main.yaml"
+        f"trap 'rm -f {tp}/sshkey' EXIT && ssh-add {tp}/sshkey 2>/dev/null && rm -f {tp}/sshkey && ansible-playbook main.yaml"
```
samdoran:
```suggestion
        f"trap 'rm -f {tmp_path}/sshkey' EXIT && ssh-add {tmp_path}/sshkey 2>/dev/null && rm -f {tmp_path}/sshkey && ansible-playbook main.yaml"
```

test/unit/config/test__base.py
samdoran: I think it would be better to use `tmp_path` rather than the default temp dir of `/tmp/.ansible-runner-*`.

AlanCoding: What's going to happen when you start down that path is that you find a _lot_ more tests doing the initialization without specifying private_data_dir. There were _relatively_ few tests hard-coding the `'/tmp'` folder into this. The practice was also relatively specific to this test module.

So I looked back in my history from when I ran the full test suite, and did an `ls /tmp`. I paste that into a text editor. I find 1755 occurrences of such directories. These cases are absolutely _not_ specific to this test module. If I went and passed in the `tmp_path` here, I believe that would be a bad idea, because I would only cover a few cases, and would further make the test patterns _less_ consistent.

If you want to get rid of the `.ansible-runner` directories, I would suggest changing the line that does `self.private_data_dir = tempfile.mkdtemp(prefix=".ansible-runner-")` to instead get that string from a constant. Then add an `autouse` fixture that mocks the constant to somewhere in the pytest tmp paths.

samdoran: That's what I figured was happening. The correct solution is to create a fixture to mock the default path or _always_ specify `private_data_dir` for tests, as you suggested.

I would rather avoid mocking, especially since the code already accepts a parameter. That does leave the responsibility on the test author/reviewer to make sure it is always set correctly, which is a downside.

samdoran: Been thinking this over more. I think that just using the default behavior for testing is ok. Setting it for all the tests is too onerous and making a fixture plus mocking would also be a bit much.

Sorry it took me a while to come around.

samdoran:
recheck

samdoran (approved):

PR #917, opened 2021-11-05
Add coverage reporting to tests

samdoran:
Enable and configure test coverage.
Add a `tox` environment for cleaning up coverage reports and docs builds.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,16 @@
+[run]
+source = ansible_runner
+data_file = test/coverage/data/coverage
+
+[report]
+skip_covered = True
+skip_empty = True
+
+[html]
+directory = test/coverage/reports/html
+
+[xml]
+output = test/coverage/reports/coverage.xml
+
+[json]
+output = test/coverage/reports/coverage.json
--- a/pytest.ini
+++ b/pytest.ini
@@ -11,3 +11,7 @@ addopts =
     --durations 10
     --durations-min 1
     --strict-markers
+    --cov
+    --cov-report html
+    --cov-report term
+    --cov-report xml
--- a/tox.ini
+++ b/tox.ini
@@ -29,3 +29,12 @@ description = Build documentation
 deps = -r{toxinidir}/docs/requirements.txt
 commands =
     sphinx-build -T -E -W --keep-going {tty:--color} -j auto -d docs/build/doctrees -b html docs docs/build/html
+
+[testenv:clean]
+description = Erase docs and coverage artifacts
+deps =
+skip_install = True
+allow_external = /bin/sh
+commands =
+    /bin/sh -c "rm -rf {toxinidir}/test/coverage/*"
+    /bin/sh -c "rm -rf {toxinidir}/docs/{_,}build"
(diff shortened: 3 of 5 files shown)


Shrews (approved):

pytest.ini
```diff
@@ -11,3 +11,7 @@ addopts =
     --durations 10
     --durations-min 1
     --strict-markers
+    --cov
```
webknjaz, line 14: @samdoran I also recommend adding an explicit `-p pytest_cov` to make sure that this plugin is pre-loaded earlier (this is sometimes important) rather than later (on the plugin discovery stage).

pytest.ini
```diff
@@ -11,3 +11,7 @@ addopts =
...
+    --cov-report html
+    --cov-report term
```
webknjaz, line 16: @samdoran I believe you may want to have `term:skip-missing` instead.

tox.ini
```diff
@@ -29,3 +29,12 @@ description = Build documentation
...
 commands =
     sphinx-build -T -E -W --keep-going {tty:--color} -j auto -d docs/build/doctrees -b html docs docs/build/html
+
+[testenv:clean]
+description = Erase docs and coverage artifacts
+deps =
+skip_install = True
+allow_external = /bin/sh
+commands =
+    /bin/sh -c "rm -rf {toxinidir}/test/coverage/*"
```
webknjaz, line 39: @samdoran I have a few comments here:
1) You could've used a combination of `git clean` commands which is easier to maintain, here's an example: https://github.com/ansible/pylibssh/blob/924771a/tox.ini#L369-L380.
2) Instead of calling sh/rm, you could do `{envpython} -c 'import shutil; shutil.rmdir("{toxinidir}/test/coverage/")'` and avoid having to use `allow_externals`.

.coveragerc
```diff
@@ -0,0 +1,16 @@
+[run]
```
webknjaz, line 1: I think you may need `parallel = True` and `branch = True` as well.

pytest.ini
webknjaz: Another comment is that you should measure the coverage for both the project and the tests. For this, you'd need to use a few subsequent args: `--cov=ansible_runner --cov=test/`. It may be really helpful to find out if some of the tests never run in CI, for example.

Here's what the coveragepy author posted last year: https://nedbatchelder.com/blog/202008/you_should_include_your_tests_in_coverage.html.

webknjaz (commented):
Plus here's my example of the coveragepy config https://github.com/ansible/pylibssh/blob/devel/.coveragerc. Hope you'll find it useful (although, some parts are related to having C-extensions so you may not need those).

.coveragerc
samdoran: `branch = True`, yes I should add that. `parallel = True` is being [set by `pytest-cov`](https://pytest-cov.readthedocs.io/en/latest/config.html#configuration).

pytest.ini
samdoran: I think it would be `term:skip-covered` or `term-missing:skip-covered`. I have `skip_covered = True` in the `coveragerc` file and I don't think the output from missing lines is useful.

PR #918, opened 2021-11-06
Runner tests require a valid PATH

urbas:
It turns out that tests `test/integration/test_runner.py` and `test/unit/test_runner.py` try to execute tools like `bash`, but the `rc` fixture completely nulls out the `env`. This means that even the `PATH` is not available on which to find `bash`. This makes the tests fail if they run in a sandboxed environment like this build: https://hydra.nixos.org/build/157248684

### DIFF
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -21,7 +21,7 @@ def rc(tmp_path):
         pexpect.EOF: None
     }
     rc.cwd = str(tmp_path)
-    rc.env = {}
+    rc.env = {"PATH": os.environ["PATH"]}
     rc.job_timeout = 10
     rc.idle_timeout = 0
     rc.pexpect_timeout = 2.


samdoran:
From looking at the test log, it doesn't look like this being being run with `tox`. Are these tests being run directly?

urbas:
> From looking at the test log, it doesn't look like this being being run with `tox`. Are these tests being run directly?

It runs directly. It seems unlikely that this would have have an effect on the `PATH` though :thinking: 

Perhaps the reason is that we run the tests in a sandbox that doesn't have `/etc/environment` (or similar) that would set the `PATH` of each process? I'm not sure about this.

Shrews:
> > From looking at the test log, it doesn't look like this being being run with `tox`. Are these tests being run directly?
> 
> It runs directly. It seems unlikely that this would have have an effect on the `PATH` though 🤔
> 

Our tests are designed to run within a tox-created environment.  For example:

   `tox -e py3`

Is there some reason you cannot run the tests as they were designed?  Supporting a different environment is a bit out of scope.

urbas:
> Is there some reason you cannot run the tests as they were designed? Supporting a different environment is a bit out of scope.

That's fair, I understand this not a problem in your CI. So feel free to ignore this PR.

I was hoping this change would fix the build in the Nix package manager (which doesn't use tox) and make tests more robust at the same time.

The thing is that tests in this project are using executables from the `PATH`. However, the `PATH` is unset in the fixture. My reasoning is that the fixture should preserve the `PATH` (since tests rely on executables from the `PATH`). Actually, I must admit that I find it rather surprising the tests are passing :thinking:

PR #930, opened 2021-11-30
Fix incorrect formatting in docs

samdoran:
RST uses double backticks to indicate verbatim formatting but there were many places that used only a single backtick, which is correct for Markdown.

Also change the documentation config to catch these errors in CI.


docs/conf.py
```diff
@@ -25,6 +25,8 @@ def _get_version():
     return '.'.join(version_parts)
 
 
+nitpicky = True
```
webknjaz, line 28: Why don't you keep it close to it's "ignore"? I normally even make a comment-delimited subsection dedicated to strictness settings that also includes "default_role".

samdoran: I had actually forgotten about the `nitpick_ignore` setting at the bottom of the file (and also noticed this line is a duplicate). I'm working through fixing those now.

Shrews (approved):

PR #936, opened 2021-12-07
Fix labeler workflow

samdoran:
Use separate actions and configurations to label PRs and issues.

Change the event trigger to `pull_request_target` to allow PRs from forks to edit the PR.


Shrews (approved):

samdoran:
recheck

Issue #926, opened 2021-11-16
Option to not write artifacts/id/stdout


AlanCoding:
Writing a `stdout` file into the artifacts directory can significantly increase the disc write volume for jobs with a lot of output. I would like to have a way to disable this.

The key line seems to be:

https://github.com/ansible/ansible-runner/blob/7fe85e791ab3194355c95d783c01d0333fca71bd/ansible_runner/runner.py#L310

I haven't looked into alternatives the pexpect provides yet.

We have a setting `suppress_ansible_output`, and AWX sets this to False. However, that only disables writing to the actual standard out. It still writes to the stdout file.

Anything in the stdout file is also present in the "stdout" key in the job event data, so this is never something which cannot be obtained through another means.

AlanCoding:
closing as done

PR #937, opened 2021-12-07
Add option to disable streaming to stdout file

AlanCoding:
This addresses https://github.com/ansible/ansible-runner/issues/926

My current intention is to make use of the setting (set it to True) in AWX, unconditionally. We don't want the additional disk writes this incurs. If the executor team is okay with the looks of this, I may ask @amolgautam25 to make that PR.

This use case is only concerned with the `pexpect` runner mode. I only messed with `subprocess` here for completeness. The subprocess mode also appears to somehow be connected with the introduction of this behavior. Maybe @ganeshrn has more history to share about this.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -139,7 +139,8 @@ The **settings** file is a little different than the other files provided in thi
 * ``pexpect_timeout``: ``10`` Number of seconds for the internal pexpect command to wait to block on input before continuing
 * ``pexpect_use_poll``: ``True`` Use ``poll()`` function for communication with child processes instead of ``select()``. ``select()`` is used when the value is set to ``False``. ``select()`` has a known limitation of using only up to 1024 file descriptors.
 
-* ``suppress_ansible_output``: ``False`` Allow output from ansible to not be printed to the screen
+* ``suppress_output_file``: ``False`` Allow output from ansible to not be streamed to the ``stdout`` or ``stderr`` files inside of the artifacts directory.
+* ``suppress_ansible_output``: ``False`` Allow output from ansible to not be printed to the screen.
 * ``fact_cache``: ``'fact_cache'`` The directory relative to ``artifacts`` where ``jsonfile`` fact caching will be stored.  Defaults to ``fact_cache``.  This is ignored if ``fact_cache_type`` is different than ``jsonfile``.
 * ``fact_cache_type``: ``'jsonfile'`` The type of fact cache to use.  Defaults to ``jsonfile``.
 
--- a/test/unit/test_runner.py
+++ b/test/unit/test_runner.py
@@ -10,6 +10,8 @@ import pytest
 import six
 import sys
 
+from test.utils.common import iterate_timeout
+
 from ansible_runner import Runner
 from ansible_runner.exceptions import CallbackError, AnsibleRunnerException
 from ansible_runner.config.runner import RunnerConfig
@@ -30,16 +32,6 @@ def rc(request, tmp_path):
     rc.pexpect_timeout = .1
     rc.pexpect_use_poll = True
     return rc
-
-
-@pytest.fixture(autouse=True)
-def mock_sleep(request, mocker):
-    # the handle_termination process teardown mechanism uses `time.sleep` to
-    # wait on processes to respond to SIGTERM; these are tests and don't care
-    # about being nice
-    m = mocker.patch('time.sleep')
-    m.start()
-    request.addfinalizer(m.stop)
 
 
 def test_simple_spawn(rc):
@@ -157,12 +149,17 @@ def test_status_callback_interface(rc, mocker):
 
 @pytest.mark.parametrize('runner_mode', ['pexpect', 'subprocess'])
 def test_stdout_file_write(rc, runner_mode):
+    if runner_mode == 'pexpect':
+        pytest.skip('Writing to stdout can be flaky, probably due to some pexpect bug')
     rc.command = ['echo', 'hello_world_marker']
     rc.runner_mode = runner_mode
     status, exitcode = Runner(config=rc).run()
     assert status == 'successful'
     stdout_path = Path(rc.artifact_dir) / 'stdout'
-    assert stdout_path.read_text().strip() == 'hello_world_marker'
+    # this can be subject to a race condition so we will be patient with the check
+    for _ in iterate_timeout(30.0, 'stdout file to be written', interval=0.2):
+        if 'hello_world_marker' in stdout_path.read_text():
+            break
 
 
 @pytest.mark.parametrize('runner_mode', ['pexpect', 'subprocess'])
--- a/test/unit/test_runner.py
+++ b/test/unit/test_runner.py
@@ -4,6 +4,7 @@ import codecs
 import os
 
 import json
+from pathlib import Path
 import pexpect
 import pytest
 import six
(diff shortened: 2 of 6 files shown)


docs/intro.rst
```diff
@@ -139,7 +139,8 @@ The **settings** file is a little different than the other files provided in thi
 * ``pexpect_timeout``: ``10`` Number of seconds for the internal pexpect command to wait to block on input before continuing
 * ``pexpect_use_poll``: ``True`` Use ``poll()`` function for communication with child processes instead of ``select()``. ``select()`` is used when the value is set to ``False``. ``select()`` has a known limitation of using only up to 1024 file descriptors.
 
-* ``suppress_ansible_output``: ``False`` Allow output from ansible to not be printed to the screen
+* ``suppress_output_file``: ``False`` Allow output from ansible to not be streamed to the ``stdout`` or ``stderr`` files inside of the artifacts directory.
+* ``suppress_ansible_output``: ``False`` Allow output from ansible to not be printed to the screen.
```
john-westcott-iv, line 143: Should there be a check to make these mutually exclusive? If they are both set ansible would "just run" with no output, maybe we don't care?

AlanCoding: AWX will set them both to `False` by design. The output is processed by the callback receiver. If someone really wanted the standard out file (which is normally deleted at the end of the run anyway) then we can introduce a setting for them to get it back.

There's a feature in AWX to download the standard out file from a job run. It is illustrative that this works by a completely unrelated method. In the request-response cycle, it is written to disk from the database data and then delivered to the client. Then deleted.

test/unit/test_runner.py
```diff
@@ -31,16 +33,6 @@ def rc(request, tmp_path):
...
 
 
-@pytest.fixture(autouse=True)
-def mock_sleep(request, mocker):
-    # the handle_termination process teardown mechanism uses `time.sleep` to
-    # wait on processes to respond to SIGTERM; these are tests and don't care
-    # about being nice
-    m = mocker.patch('time.sleep')
-    m.start()
-    request.addfinalizer(m.stop)
```
AlanCoding, line 41: I checked the timing, and the whole module takes about 3.6 seconds. Maybe this was reducing it to 3.5 or 3.4, but the difference also could have been indistinguishable. I'm using `time.sleep` to make the assertion reliable, so that's why I did this.

test/unit/test_runner.py
```diff
@@ -152,3 +144,32 @@ def test_status_callback_interface(rc, mocker):
...
+@pytest.mark.parametrize('runner_mode', ['pexpect', 'subprocess'])
+def test_stdout_file_write(rc, runner_mode):
+    rc.command = ['echo', 'hello_world_marker']
+    rc.runner_mode = runner_mode
+    status, exitcode = Runner(config=rc).run()
+    assert status == 'successful'
+    stdout_path = Path(rc.artifact_dir) / 'stdout'
+    # this can be subject to a race condition so we will be patient with the check
+    start_time = time.time()
+    while time.time() - start_time < 2.0:
```
Shrews: Use `iterate_timeout()` here from `test/utils/common.py`. You can find examples in the test code.

Shrews (changes requested):
Looks pretty good to me, other than the comment on the tests to re-use the `iterate_timeout()` function.

test/unit/test_runner.py
```diff
@@ -154,13 +155,9 @@ def test_stdout_file_write(rc, runner_mode):
     assert status == 'successful'
     stdout_path = Path(rc.artifact_dir) / 'stdout'
     # this can be subject to a race condition so we will be patient with the check
-    start_time = time.time()
-    while time.time() - start_time < 2.0:
+    for _ in iterate_timeout(2.0, 'stdout file to be written', interval=0.2):
```
Shrews: We can get some pretty slow test VMs, so 2 seconds can be cutting it pretty darn close, which is likely why you see the test failure. I see no reason not to just bump this to, say, 30 seconds.

AlanCoding: Thanks. The first time I checked results I saw 2 configurations failing and after adding the 2 second poll 1 was failing. So this would fit the picture that it's slow, dependent on the hardware.

Shrews (approved):

AlanCoding:
Interesting... I'm wondering if this was a pre-existing failure of the same type

```
=================================== FAILURES ===================================
[31m[1m_____________________________ test_password_prompt _____________________________[0m
[gw3] linux -- Python 3.8.12 /home/zuul/src/github.com/ansible/ansible-runner/.tox/py38/bin/python

rc = <ansible_runner.config.runner.RunnerConfig object at 0x7ff864f2e520>

    def test_password_prompt(rc):
        rc.command = [sys.executable, '-c' 'import time; print(input("Password: "))']
        rc.expect_passwords[re.compile(r'Password:\s*?$', re.M)] = '1234'
        status, exitcode = Runner(config=rc).run()
        assert status == 'successful'
        assert exitcode == 0
        with open(os.path.join(rc.artifact_dir, 'stdout')) as f:
>           assert '1234' in f.read()
[1m[31mE           AssertionError: assert '1234' in ''[0m
[1m[31mE            +  where '' = <built-in method read of _io.TextIOWrapper object at 0x7ff864d44ee0>()[0m
[1m[31mE            +    where <built-in method read of _io.TextIOWrapper object at 0x7ff864d44ee0> = <_io.TextIOWrapper name='/tmp/pytest-of-zuul/pytest-0/popen-gw3/test_password_prompt0/artifacts/21213e93-55b2-4899-8dc7-edeebb7c59ec/stdout' mode='r' encoding='UTF-8'>.read[0m

exitcode   = 0
f          = <_io.TextIOWrapper name='/tmp/pytest-of-zuul/pytest-0/popen-gw3/test_password_prompt0/artifacts/21213e93-55b2-4899-8dc7-edeebb7c59ec/stdout' mode='r' encoding='UTF-8'>
rc         = <ansible_runner.config.runner.RunnerConfig object at 0x7ff864f2e520>
status     = 'successful'

[1m[31mtest/integration/test_runner.py[0m:22: AssertionError
```

I can apply the same methods to that assertion. That looks like it may be the proper thing to do.

Shrews:
The disk space issue aside (no idea why we're seeing that on occasion), each time the new `test_stdout_file_write` has failed has been with the `pexpect` path (not the `subprocess` path). That may be worth looking into.

Shrews:
> The disk space issue aside (no idea why we're seeing that on occasion), each time the new `test_stdout_file_write` has failed has been with the `pexpect` path (not the `subprocess` path). That may be worth looking into.

FYI, the failure using `pexpect` is reproducible on Linux. It can take several runs, but definitely reproducible.

Shrews (changes requested):
Removing approval until the test failure under `pexpect` is figured out.

AlanCoding:
Would it be a problem for me to do a `pytest.skip`  for the pexpect case? It looks like it's just buggy from some problem with the library maybe.

AlanCoding:
@Shrews I added the skip, which looks like it's working to stabilize tests. I can make an issue to cover the stdout log file not working with pexpect

Shrews (approved):
It makes me a sad panda that we have to skip the test case we _explicitly_ want this feature for, but this change obviously didn't create the error, only exposed it. So I guess skipping it is ok for now.

PR #943, opened 2021-12-17
Fix suppress_output_file bug where event_callback was not called

AlanCoding:
Sorry, I had a major flaw with https://github.com/ansible/ansible-runner/pull/937 and I should have tested that better before it was merged.

The handle was not passed to pexpect, and that was wrong. The handle is the way that it dispatches events to the `event_callback`. Even if you do not intend to write to the stdout file, this is necessary.

Design-wise, if `suppress_output_file` is given, we _do_ want to use the stdout handle wrapper, but we do not want to use the stdout handle itself. I think this code structure articulates that about as best as possible. Alternatively, we could create a null handler and pass that to `OutputEventFilter`, instead of handling the `None` case.

----

Separately, I may have discovered a bug related to the test failures on the prior PR.

https://github.com/ansible/ansible-runner/blob/d01304455c3c25669fba70d9983487ffdb85f1c5/ansible_runner/utils/__init__.py#L373

This is assuming the `.splitlines(True)` leaves `\r\n` at the end of the line. This is... not universal. Because of that, the final character in the `echo` command gets truncated, because it has a `\n` without the `\r`. This looks completely fixable, but I wish to keep that separate.

### DIFF
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -10,7 +10,6 @@ import codecs
 import collections
 import datetime
 import logging
-from io import StringIO
 
 import six
 import pexpect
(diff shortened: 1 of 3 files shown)


AlanCoding:
Here's what the null file-like object looks like:

```python
        class NullWriter:
            def write(self, s):
                pass

            def flush(self):
                pass

            def close(self):
                pass
```

This is not the implementation here, but it would replace `StringIO` in the alternative that I discussed. I'm torn as to which is the best.

Shrews (approved):

eqrx (approved):

PR #944, opened 2021-12-17
Create GitHub Actions workflow for running tests

samdoran:
Run sanity, unit, and integration tests in GitHub actions. We should be able to disable some Zuul jobs after merging this.

Also upload code coverage reports.

We may want to look at a matrix of Python version X Ansible version. Currently it only tests one Ansible version per Python version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,9 +37,29 @@ jobs:
         run: tox
 
 
+  images:
+    name: Build Image - ${{ matrix.runtime }}
+    runs-on: ubuntu-20.04
+
+    strategy:
+      fail-fast: false
+      matrix:
+        runtime:
+          - docker
+          - podman
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Build Container Image
+        run: ${{ matrix.runtime }} build --rm=true -t quay.io/ansible/ansible-runner:devel -t quay.io/ansible/ansible-runner:latest .
+
+
   integration:
     runs-on: ubuntu-20.04
     name: Integration - ${{ matrix.py_version.name }}
+    needs: images
 
     env:
       TOXENV: ${{ matrix.py_version.tox_env }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,71 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+
+
+jobs:
+  sanity:
+    name: ${{ matrix.test.name }}
+    runs-on: ubuntu-20.04
+    container:
+      image: quay.io/ansible/ansible-runner-test-container:2.0.0
+      env:
+        PIP_CACHE_DIR: ${{ runner.temp }}/.cache/pip
+        PY_COLORS: 1
+        TOXENV: ${{ matrix.test.tox_env }}
+
+    strategy:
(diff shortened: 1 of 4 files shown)


AlanCoding:
It looks like this is running the docker and podman tests? This is fantastic. The whole thing ran in under 5 minutes too.

samdoran:
Yup, just running the tests in a VM hosted by GitHub Actions. Ideally we would run those tests in our test container, but I haven't worked how to run `podman` inside of a Docker container where we don't control the command used to invoke the container.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,116 @@
...
+        uses: actions/checkout@v2
+
+      - name: Create tox environment
+        run: tox --notest
+
+      - name: Run tests
+        run: tox
+
+
+  integration:
```
Shrews, line 40: These would need to be run using the runner image built using the code within the PR being submitted.

AlanCoding: yeah, I was wondering how that might be working.

samdoran: Oh, I didn't think we'd need that for these tests. I'll add a job to build the images first.

samdoran: Done.

samdoran:
Adding the image build steps brings overall execution up to a little over nine minutes. Not too bad, but building that image every time is costly.

Shrews:
Note for @shanemcd and myself to review this after the holidays to make sure we would be covered on both GHA side and Zuul side (global job references?).

Shrews:
> Note for @shanemcd and myself to review this after the holidays to make sure we would be covered on both GHA side and Zuul side (global job references?).

FYI, there are references to some of the runner jobs in the `ansible/ansible` project at https://github.com/ansible/project-config/blob/2bd212f0b14ae32bf687865f95cb7c695808ed4f/zuul.d/projects.yaml#L853-L877.

Shrews:
I tried to validate that we were actually using the image created from the new GHA job that creates the image for the integration tests. I expected PR #960 to fail the integration tests, but they did not (and it does seem the exception is being written to the proper location for the callback in the container image). It might be explained by the [comment left by @AlanCoding](https://github.com/ansible/ansible-runner/pull/960/files#r786246748).

I know @samdoran put *a lot* of work into this (thanks Sam!), but given that this doesn't seem to be testing the image created from the PR in the integration tests, I'm not on board with merging this.

Shrews:
> I tried to validate that we were actually using the image created from the new GHA job that creates the image for the integration tests. I expected PR #960 to fail the integration tests, but they did not (and it does seem the exception is being written to the proper location for the callback in the container image). It might be explained by the [comment left by @AlanCoding](https://github.com/ansible/ansible-runner/pull/960/files#r786246748).
> 
> I know @samdoran put _a lot_ of work into this (thanks Sam!), but given that this doesn't seem to be testing the image created from the PR in the integration tests, I'm not on board with merging this.

I believe that I've now found a way to make this work properly for integration tests. I'll be updating this PR.

Shrews:
Commit https://github.com/ansible/ansible-runner/pull/960/commits/08e3284d46a643313a1c4155cefef921515db6ea on PR #960 shows that the change I just pushed here should work properly for integration testing.

Shrews (approved):

PR #950, opened 2022-01-05
Fix bug with truncating newline

AlanCoding:
This was followup promised in https://github.com/ansible/ansible-runner/pull/943

#### Remaining concern

Of course I love `.rstrip` for the simplicity of it, but there may be a case where this method receives multiple empty lines, and it is expected to keep them. Like, it gets `foo\n\n\n`, and it is expected to write `foo\n\n`. This whole thing is frustrating and confusing. This scenario is probably unlikely to happen because verbose events only write 1 line at a time anyway. There might still be a case I'm missing that's important, and in that case, this could be adjusted so that only 1 trailing newline is removed from `stdout_chunk`, instead of all newline characters. I expect that would be less performant, so if this satisfies the current needs I'd keep it.

### DIFF
--- a/ansible_runner/utils/__init__.py
+++ b/ansible_runner/utils/__init__.py
@@ -374,7 +374,7 @@ class OutputEventFilter(object):
                 event_data['uuid'] = str(uuid.uuid4())
             self._counter += 1
             event_data['counter'] = self._counter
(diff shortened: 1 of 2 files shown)


eqrx (approved):

Shrews (approved):
Approved. Probably should be backported, too.

Shrews:
recheck

PR #952, opened 2022-01-07
Do some test cleanup

Shrews:
We have been seeing some random test failures due to `No space left on device` errors. Not entirely sure what is causing that, but this is at least a start. This will remove the copy of the test project tree that is made for every test that uses the `project_fixtures` fixture.

Also, remove a now unused pytest fixture (`clear_integration_artifacts`).

Also fix a flaky test (`test_stdout_file_no_write`) that fails randomly under `pexpect` preventing this from merging. The `assert`'s being removed are not actually testing the feature.

### DIFF
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -29,24 +29,6 @@ def rc(tmp_path):
     return rc
 
 
-@pytest.fixture(scope='session')
-def clear_integration_artifacts(request):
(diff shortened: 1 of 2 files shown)


Shrews:
recheck

AlanCoding (approved):

eqrx (approved):

PR #955, opened 2022-01-12
Remove script used for test migration

AlanCoding:
I wrote this years ago to migrate tests in AWX that needed to be ported into ansible-runner. They have diverged significantly since then and it wouldn't be viable to re-run this.


eqrx (approved):

Shrews:
recheck

Shrews (approved):

PR #960, opened 2022-01-17
DNM: Test gha ci

Shrews:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,145 @@
...
+        uses: actions/checkout@v2
+
+      - name: Build Container Image
+        run: ${{ matrix.runtime }} build --rm=true -t quay.io/ansible/ansible-runner:devel -t quay.io/ansible/ansible-runner:latest .
+
+
+  integration:
+    runs-on: ubuntu-20.04
+    name: Integration - ${{ matrix.py_version.name }}
+    needs: images
```
AlanCoding: I have been frustrated by the lack of artifacts support with Github Actions myself. My understanding of "needs:" is that it is strictly temporal. As in, it will make a step wait until a prior check is finished, but it doesn't carry over anything from the file system of the prior job.

I've seen some plugins that implement custom means of passing artifacts, but they're ugly. These tend to push the artifacts to some datastore and then pull it again in the job that uses it. Yes, `needs:` would play a role there to get the timing right, but in this case you would still need to pull and pull from an image registry, if I understand correctly.

Shrews: That might explain why the GHA integration tests did not fail as I expected them too with this change. If this is the case, then I don't think switching to GHA for the tests is a viable thing to do.

webknjaz: FYI I used GH container registry for artifacts in the creator-ee repo, you may want to use it as an example. The trick is to bake in the workflow run ID so that you could publish from one job and pick up from the next stage. See the release workflow specifically for this example.

webknjaz: Or did you mean normal file artifacts? There's official upload-/download-artifacts actions for that.

README.md
```diff
@@ -5,7 +5,7 @@ Ansible Runner
 [![Documentation](https://readthedocs.org/projects/ansible-runner/badge/?version=stable)](https://ansible-runner.readthedocs.io/en/latest/)
 [![Code of Conduct](https://img.shields.io/badge/Code%20of%20Conduct-Ansible-silver.svg)](https://docs.ansible.com/ansible/latest/community/code_of_conduct.html)
 [![Ansible Mailing lists](https://img.shields.io/badge/Mailing%20lists-Ansible-orange.svg)](https://docs.ansible.com/ansible/latest/community/communication.html#mailing-list-information)
-
+[![codecov](https://codecov.io/gh/ansible/ansible-runner/branch/devel/graph/badge.svg?token=CmCcjBz0pQ)](https://codecov.io/gh/ansible/ansible-runner)
```
webknjaz, line 8: No need for a token in public repos.

test/integration/test_runner.py
```diff
@@ -14,6 +14,7 @@
 from test.utils.common import iterate_timeout
 
 
+@pytest.mark.xfail(reason='Test is unstable')
```
webknjaz, line 17: If it's flaky, it should be
```suggestion
@pytest.mark.xfail(reason='Test is unstable', strict=False)
```

Issue #964, opened 2022-01-18
Preserve modification times when streaming directories via process


AlanCoding:
Demonstrating via AWX, I ran a job that gathered facts for a host. The modification time of the files in the fact cache folder differ on the execution node versus the control node.

```
(env) [alancoding@alan-red-hat awx]$ docker exec -it tools_awx_1 python -c "import os; print(os.path.getmtime('/tmp/awx_42_e25zlma1/artifacts/42/fact_cache/HostSelectionPurpose204494140484113530734218352555284941748073844862122634129'))"
1642533390.6791515
(env) [alancoding@alan-red-hat awx]$ docker exec -it tools_receptor_1 python -c "import os; print(os.path.getmtime('/tmp/awx_42_e25zlma1/artifacts/42/fact_cache/HostSelectionPurpose204494140484113530734218352555284941748073844862122634129'))"
1642533390.142149
```

This matters because AWX uses the modification times to make decisions about cache invalidation.

We are only really concerned about preserving timestamps when streaming from the execution node back to the control node (`ansible-runner process`) and less concerned about the other direction.

Specifically, the unstream_dir method is relevant:

https://github.com/ansible/ansible-runner/blob/d3a202f34a27fd982f17465cfbf41f991844542d/ansible_runner/utils/streaming.py#L53

We think that the stack overflow solution here might work:

https://stackoverflow.com/a/9813471

Ping @jbradberry

AlanCoding:
Fix is merged. Further work is tracked in the AWX issue.

PR #966, opened 2022-01-19
Preserve timestamps when unstreaming dirs

AlanCoding:
Connect https://github.com/ansible/ansible-runner/issues/964

That fixes the linked issue with the limitation (discovered by the test) that we might miss a 2 second delta or less. This assumes the computer clock is perfectly synced on the machines that do the compressing and extracting.

Nonetheless, I think this opens up an option for AWX to get the job done. I show a demo in the test with the `very_old.txt` file, where I artificially set the mod and access time to 1 day in the past. It seems like that would work for the AWX use case, assuming that the two computers agree on the day, which doesn't sound like an unreasonable demand.

Ping @jbradberry

### DIFF
--- a/ansible_runner/utils/streaming.py
+++ b/ansible_runner/utils/streaming.py
@@ -69,6 +69,7 @@ def unstream_dir(stream, length, target_directory):
 
         with zipfile.ZipFile(tmp.name, "r") as archive:
             # Fancy extraction in order to preserve permissions
+            # AWX relies on the execution bit, in particular, for inventory
             # https://www.burgundywall.com/post/preserving-file-perms-with-python-zipfile-module
             for info in archive.infolist():
                 out_path = os.path.join(target_directory, info.filename)
@@ -87,6 +88,7 @@ def unstream_dir(stream, length, target_directory):
                 archive.extract(info.filename, path=target_directory)
 
                 # Fancy logic to preserve modification times
+                # AWX uses modification times to determine if new facts were written for a host
                 # https://stackoverflow.com/questions/9813243/extract-files-from-zip-file-and-retain-mod-date
                 date_time = time.mktime(info.date_time + (0, 0, -1))
                 os.utime(out_path, times=(date_time, date_time))
--- a/ansible_runner/utils/streaming.py
+++ b/ansible_runner/utils/streaming.py
@@ -1,3 +1,4 @@
+import time
 import tempfile
 import zipfile
 import os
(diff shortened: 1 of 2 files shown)


AlanCoding:
I'm pretty happy with this in its current form. It does the job well enough for our purposes and tests are passing.

jbradberry (approved):

ansible_runner/utils/streaming.py
```diff
@@ -85,6 +86,11 @@ def unstream_dir(stream, length, target_directory):
 
                 archive.extract(info.filename, path=target_directory)
 
+                # Fancy logic to preserve modification times
```
Shrews, line 90: Can we add a bit more context here to this comment as to _why_ we need to preserve modifications time?

AlanCoding: I thought this would be kind of a general expectation for extracting a zip file.

In the case of runner, we have a control & execution machine. The control machine creates the private_data_dir, the worker runs the job, and it gets streamed back. The modtimes can be used to distinguish if the control machine last modified the file, or if the file was modified during the job run.

In the AWX case, we use modification times to determine if new facts were written for a host. This is a big performance optimization for the case where the inventory has 10k hosts but you run against 1 host. Most playbooks will gather facts, and AWX maintains a fact cache in the database. A job run which should go super fast would get bogged down saving the rather substantial facts data structure for 10k hosts.

I have trouble putting that into a short sentence :)

Shrews: The new Executor team (currently just myself and @eqrx) lack A LOT of context around design decisions in the current code base (much of it having to do with AWX where we have no experience or insight), so it makes maintaining it rather difficult. Just a simple sentence like "AWX uses  modification times to determine if new facts were written for a host" would give us a lot of clarity when we see code like this where the purpose is not immediately clear.

test/unit/utils/test_utils.py
```diff
@@ -147,6 +149,48 @@ def test_transmit_permissions(tmp_path, fperm):
...
+    size_data = json.loads(first_line.strip())
+    unstream_dir(outgoing_buffer, size_data['zipfile'], dest_dir)
+
+    # Assure modification times are internally consistent
+    mod_delta = os.path.getmtime(dest_dir / 'a.txt') - os.path.getmtime(dest_dir / 'b.txt')
+    assert mod_delta >= 1.0
+
+    # Assure modification times are same as original
+    for filename in ('a.txt', 'b.txt'):
+        assert os.path.getmtime(dest_dir / filename) == os.path.getmtime(dest_dir / filename)
```
Shrews: This will never fail. Don't you want to compare `source_dir` to `dest_dir`?

AlanCoding: Yes, great catch, thanks

Shrews (approved):

Shrews:
Approved. Seems like something we should backport after merged.

PR #976, opened 2022-01-24
Remove use of distutils

### LINKED ISSUE #969: Stop using `distutils`
It's deprecated. Python 3.10 raises a `DeprecationWarning` and Python 3.12 won't have it in the stdlib anymore.
Why fix it now and not wait for another 2 years when this will produce an `ImportError`? Because it's nice to solve such problems ahead of time and also since ansible-runner is a library, it also causes these warnings (_or even errors_ with pytest's [`filterwarning = error`] or [`python -Werror`]&nbsp;[^-W-python-option]) for the projects that have Python 3.10 in their test matrixes. In particular, it's been observed in ansible-navigator.

[`filterwarning = error`]: https://docs.pytest.org/en/6.2.x/reference.html#confval-filterwarnings
[`python -Werror`]: https://docs.python.org/dev/library/warnings.html#the-warnings-filter
[^-W-python-option]: Learn more at https://docs.python.org/dev/using/cmdline.html#cmdoption-W

Shrews:
Fixes #969

### DIFF
--- a/ansible_runner/config/runner.py
+++ b/ansible_runner/config/runner.py
@@ -23,7 +23,7 @@ import shlex
 import stat
 import tempfile
 import six
-import distutils.dir_util
+import shutil
 
 from six import string_types, text_type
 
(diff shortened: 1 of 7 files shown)


Shrews:
The `copytree()` portion of code used for directory isolation does not seem to be covered by any tests. I'd like to add a test for that before merging this change. However, I discovered a bug in that bit of code that needs fixed before we can merge this (see issue #977).

eqrx (approved):

PR #986, opened 2022-02-03
fix rpm-builder build

kdelee:
This fixes the rpm-builder build...
the "make rpm" target it still not working for other reasons

would be great to get that fixed + add a PR check so https://github.com/ansible/ansible-runner/pull/985 wouldn't happen

PR #987, opened 2022-02-03
Fix rpm build

simaishi:
Follow-up for #986. There were still multiple rpm build failures (centos 8 stream change, .spec being out-of-date, etc.).  I fixed for el8 only.

@kdelee please try this out


kdelee:
Works for me!!
```
ls rpm-build/ansible-runner-2.1.1.dev30-0.git2022020321.116be0d.el8.noarch.rpm 

rpm-build/ansible-runner-2.1.1.dev30-0.git2022020321.116be0d.el8.noarch.rpm
```

kdelee (approved):

kdelee:
Would be cool to run `make rpm` in https://github.com/ansible/ansible-runner/blob/devel/.github/workflows/ci.yml

doesn't have to be this PR, but once this is fixed it will be an option

Shrews (approved):

PR #995, opened 2022-02-11
Implement ansible's resolved action

matburt:
This supercedes #816 

This also includes some black formatting changes.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,7 +1,7 @@
 import shutil
 
 from pathlib import Path
-from distutils.version import LooseVersion
+from packaging.version import Version
 import subprocess
 
 from ansible_runner import defaults
@@ -60,7 +60,7 @@ def is_pre_ansible212():
             .stdout.strip()
             .decode()
         )
-        if LooseVersion(base_version) < LooseVersion("2.12"):
+        if Version(base_version) < Version("2.12"):
             return True
     except pkg_resources.DistributionNotFound:
         pass
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -1,6 +1,8 @@
 import shutil
 
 from pathlib import Path
+from distutils.version import LooseVersion
+import subprocess
 
 from ansible_runner import defaults
 
@@ -44,6 +46,30 @@ def is_pre_ansible211():
 def skipif_pre_ansible211(is_pre_ansible211):
     if is_pre_ansible211:
         pytest.skip("Valid only on Ansible 2.11+")
+
+
+@pytest.fixture(scope="session")
+def is_pre_ansible212():
+    try:
+        base_version = (
+            subprocess.run(
+                "python -c 'import ansible; print(ansible.__version__)'",
+                capture_output=True,
+                shell=True,
+            )
+            .stdout.strip()
+            .decode()
+        )
+        if LooseVersion(base_version) < LooseVersion("2.12"):
+            return True
+    except pkg_resources.DistributionNotFound:
+        pass
+
+
+@pytest.fixture(scope="session")
+def skipif_pre_ansible212(is_pre_ansible212):
+    if is_pre_ansible212:
+        pytest.skip("Valid only on Ansible 2.12+")
 
 
 # TODO: determine if we want to add docker / podman
--- a/test/integration/test_display_callback.py
+++ b/test/integration/test_display_callback.py
@@ -197,8 +197,32 @@ def test_callback_plugin_task_args_leak(executor, playbook):
     # make sure playbook was successful, so all tasks were hit
     assert not events[-1]['event_data']['failures'], 'Unexpected playbook execution failure'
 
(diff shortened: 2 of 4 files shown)


test/conftest.py
```diff
@@ -56,26 +55,51 @@ def pytest_generate_tests(metafunc):
...
                     ),
                 )
                 for runtime in CONTAINER_RUNTIMES
             )
-            metafunc.parametrize('runtime', args)
+            metafunc.parametrize("runtime", args)
             break
 
 
+@pytest.fixture(scope="session")
```
matburt, line 74: This is the real addition, not sure how I feel about subprocess here but we need to get ansible's internal core version

test/integration/test_display_callback.py
```diff
@@ -174,32 +211,65 @@ def test_callback_plugin_no_log_filters(executor, playbook):
...
-        assert 'PRIVATE' not in json.dumps(event), event
+        assert "PRIVATE" not in json.dumps(event), event
     # make sure playbook was successful, so all tasks were hit
-    assert not events[-1]['event_data']['failures'], 'Unexpected playbook execution failure'
+    assert not events[-1]["event_data"][
+        "failures"
+    ], "Unexpected playbook execution failure"
+
+
+@pytest.mark.parametrize(
```
matburt, line 243: Here's the new test

matburt:
[47ee0e5](https://github.com/ansible/ansible-runner/pull/995/commits/47ee0e5fe63d1a45c133b2df4bd207a6d7ca70e9) keeps us from resurfacing #969

cidrblock:
+1, not because I completely grok it all but because it doesn't use distulils :)

test/conftest.py
```diff
@@ -46,6 +48,30 @@ def skipif_pre_ansible211(is_pre_ansible211):
         pytest.skip("Valid only on Ansible 2.11+")
 
 
+@pytest.fixture(scope="session")
```
matburt, line 51: I'm not crazy about what the `is_pre_ansible211` test does. It doesn't seem like it'll work in the general case.

matburt: I'm also not sure how I feel about this implementation generally but it gets us there quick with whatever variant of Ansible is installed in the environment.

shanemcd (approved):

test/conftest.py
```diff
@@ -1,6 +1,8 @@
 import shutil
 
 from pathlib import Path
+from packaging.version import Version
```
shanemcd, line 4: @simaishi Can you check and see if this is available as an RPM from the standard repos?

matburt: 🤞🏻 it's also used by `ansible-core` so my working assumption was that we had solved it for the necessary dependencies.

simaishi: ansible-core has it bundled, but we do have an access to the standalone rpm as well.

AlanCoding:
This looks most similar to the `task_action` key... makes sense, right? Before, we had task_action, now we have resolved_action in addition to that. But those are handled in separate parts of the code, and I haven't understood why. We seem to get both of them from the same `task` object... so why would they be in different places?

AlanCoding:
To give my comment greater specificity, 

```diff
diff --git a/ansible_runner/display_callback/callback/awx_display.py b/ansible_runner/display_callback/callback/awx_display.py
index aa29f39..0aee596 100644
--- a/ansible_runner/display_callback/callback/awx_display.py
+++ b/ansible_runner/display_callback/callback/awx_display.py
@@ -409,6 +409,7 @@ class CallbackModule(DefaultCallbackModule):
             task=(task.name or task.action),
             task_uuid=str(task._uuid),
             task_action=task.action,
+            resolved_action=getattr(task, 'resolved_action', ''),
             task_args='',
         )
         try:
@@ -554,7 +555,6 @@ class CallbackModule(DefaultCallbackModule):
             name=task.get_name(),
             is_conditional=is_conditional,
             uuid=task_uuid,
-            resolved_action=getattr(task, 'resolved_action', '')
         )
         with self.capture_event_data('playbook_on_task_start', **event_data):
             super(CallbackModule, self).v2_playbook_on_task_start(task, is_conditional)
```

this would seem more clear and still passes the test

shanemcd:
recheck

test/conftest.py
Shrews: The `is_pre_ansible211` fixture should work in the general case. The ansible package  used in our tests will be `ansible-core` for 2.11 or higher. 2.10 is `ansible-base`, and 2.9 and earlier is `ansible`.

Shrews (approved):

AlanCoding:
@Shrews in the future, I want to make sure that we do integration testing with AWX before merging any change intended to be consumed by AWX. In this case, I did, and those were passing.

Shrews:
@AlanCoding Ok. @shanemcd had asked me about merging this today, so I assumed that was already done.

Issue #810, opened 2021-09-01
Support for cgroupsv2


Sakirnth-Z:
Hi,

At the moment ansible-runner only works with cgroupsv1. cgroupsv1 filesystem is not mounted by default anymore in the new version of Debian (Debian Bullseye). Therefore cgroupsv2 should be supported now in ansible-runner.

Could you please adapt the code that it will work with cgroupsv2.

Here is a link to cgroupsv2 documentation: 
https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html

And this is the original bug report on Debian, here you can find more informations:  
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=993371

Thank you very much and cheeers
Saki

chrismeyersfsu:
The code in runner that interfaces directly in cgroups was used only in the Resource Profiling feature in AWX. The dependent Resource Profiling feature has been removed and now we just need to remove the corresponding code from runner.

Are you using the Resource Profiling feature?

Sakirnth-Z:
Thanks for your answer. Therefore it's dead code now?

Personally I do not use this feature. But it was reported in debian bugs, so I guess some people are using it.

Cheers,
Saki

samdoran:
Further discussion is needed to determine if we want to keep the resource profiling feature or remove it entirely.

Akasurde:
resolved_by_pr #997

Sakirnth-Z:
Thank you for your work. In Debian I patched the version 2.1.3 with this patch.

Greetings
Sakirnth

PR #997, opened 2022-02-14
Purge deprecated resource profiling capability

matburt:


matburt:
This should also close #810

shanemcd (approved):

PR #1000, opened 2022-02-16
Adding --omit-env-files option

john-westcott-iv:
Adding an option to not write out files into the env folder as these can contain sensitive information. 
Changing write fifo to take either bytes or a string.

Initially based off of https://github.com/ansible/ansible-runner/pull/812

### DIFF
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -218,11 +218,11 @@ DEFAULT_CLI_ARGS = {
             ),
         ),
         (
-            ("--no-store-env",),
+            ("--omit-env-files",),
             dict(
-                action="store_false",
-                dest='store_env',
-                help="Set to false to prevent the writing of the env directory"
+                action="store_true",
+                dest='suppress_env_files',
+                help="Add flag to prevent the writing of the env directory"
             ),
         ),
         (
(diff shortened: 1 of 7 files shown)


ansible_runner/__main__.py
```diff
@@ -217,6 +217,15 @@
                      "(status and stdout still included for other events)"
             ),
         ),
+        (
+            ("--no-store-env",),
+            dict(
+                default="True",
```
shanemcd: The default value here is `True` for "not" storing the env, but your code in `__init__.py` says `kwargs.get('store_env', True)`.

I'm personally not a fan of CLI options that start with `--no-`... they are confusing. I think I'd prefer to change this to `--store-env`.

john-westcott-iv: One of my goals in writing this was to maintain backwards compatibility which is `kwargs.get('store_env', True)`. Initially I was going to do an inputted parameter like: `--store-env <True|False>` but that ended up seeming like extra typing since it defaults to `True` you would only ever be typing `--store-env False`. If I just go with `--store-env` the default (no parameter) would break this backwards compatibility. Because of these two factors I ended up going with `--no-` flag. Line 223 should be removed; it really does nothing and was from the initial go at the <True|False> form of the parameter. If we don't want to go with `--no` I wouldn't be against renaming the parameter to something like `--skip-env` and propagating that through the code (i.e. `kwargs.get('skip_env', False)`. Or we could go back to the text boolean parameter option of `--store-env <True|False>`. Or, if we are ok with changing the default behavior, we could also do `--store-env` to go back to the previous default behavior. Let me know what you think.

shanemcd: Ok, what you've said here makes sense. I was mostly confused about `default="True"`. Remove that and we are good.

shanemcd (approved):

Shrews (approved):

PR #1005, opened 2022-02-18
make callback event_dict fields consitent

valbendan:
add `name` field to `event_dict` to make callback event_dict fields more consitent (intended to fix missing `name` field)

### DIFF
--- a/ansible_runner/display_callback/callback/awx_display.py
+++ b/ansible_runner/display_callback/callback/awx_display.py
@@ -708,6 +708,7 @@ class CallbackModule(DefaultCallbackModule):
         # NOTE: Not used by Ansible 2.x.
         event_data = dict(
             task=task,
+            name=task.get_name(),
         )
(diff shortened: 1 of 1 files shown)


AlanCoding:
Without more information, I'm hesitant about merging this diff. What consistency does this add? Other cases we use `name` includes this, which is the _play_ name.

https://github.com/ansible/ansible-runner/blob/a3a7d1f003996fe7c757436b021d54f4a84aa4c6/ansible_runner/display_callback/callback/awx_display.py#L506-L508

I do see other cases where we put in the task name, but all of those cases look rather duplicated with the `task` key. For a problem that like, I would prefer to put up a notice to deprecate the ambiguous field in favor of the other field.

Akasurde:
@valbendan Could you please address the https://github.com/ansible/ansible-runner/pull/1005#issuecomment-1101444040? Thanks.

PR #1027, opened 2022-03-16
Fix image build to override root homedir in /etc/passwd

### LINKED ISSUE #1024: `ansible-runner` base image improperly overrides `$HOME`, breaking various Ansible things
The `ansible-runner` base image doesn't actually change the `root` user's homedir, it just overrides the `HOME` envvar to `/home/runner` while leaving the actual homedir in `/etc/passwd` as `/root`. Anything that attempts to resolve root's homedir by asking the system what it should be (eg, Ansible, all over the place via `echo ~root`) will get the wrong answer. For places where Ansible expands the directory control-side and uses the expanded value, it's still "wrong" but not broken, but for things like, eg, `async_status` which expect the homedir to resolve properly (and in a matching way) on the target as well as remotely, things will be broken badly. 

In particular, this issue breaks `async_status` for all tasks started over the `local` connection, since the status module is looking for the job status file in `~/.ansible_async/(jid)`, but the control-side pre-creates it in `~root/.ansible_async(jid)`. If properly overridden, those two resolve to the same value, but since it's not properly overridden, the status module can't find the JID file. This effectively means any `async` task that uses the local connection (`connection: local`, `local_action`, probably networking) will be broken on any EE based on the runner default EE.

Probably just need to `usermod -d /home/runner root` early in the Dockerfile, but it might be more complicated since it's root...

nitzmahone:
fixes #1024 

* using only `HOME` envvar to override homedir causes mismatches with anything that asks for it a different way (eg, `echo ~root`)
* kicking the can down the road on "why are we overriding /root as root's homedir anyway?"

### DIFF
--- a/utils/entrypoint.sh
+++ b/utils/entrypoint.sh
@@ -1,19 +1,30 @@
 #!/usr/bin/env bash
 
-# We need to fix a number of problems here that manifest under different container runtimes. If we're running
-# as a legit default user that has an entry in /etc/passwd and a valid homedir that's not `/`, we're all good.
-# If the username/uid we're running under is not represented in /etc/passwd and/or doesn't have a homedir that's not
-# `/` (eg, the container was run with --user and some dynamic unmapped UID from the host with primary GID 0), we need to
-# correct that. Some things (eg podman/cri-o today) already create an /etc/passwd entry on the fly in this case, but
-# they set the homedir to `/`, which causes potential collisions with mounted/mapped volumes. For consistency, we'll
-# just always set every dynamically-created user's homedir to `/home/runner`, which we've already configured in a way
-# that should always work (eg, ug+rwx and all dirs owned by the root group).
+# We need to fix a number of problems here that manifest under different container runtimes, as well as paper over some
+# things to simplify runner's containerized launch behavior. Since runner currently always expects to bind-mount its
+# callback plugins under ~/.ansible, it must have prior knowledge of the user's homedir before the container is launched
+# in order to know where to mount in the callback dir. In all cases, we must get a consistent answer from $HOME
+# and anything that queries /etc/passwd for a homedir (eg, `~root`), or lots of things (including parts of Ansible
+# core itself) will be broken.
 
-# if current user is not listed in /etc/passwd, add an entry with username==uid, primary gid 0, and homedir /home/runner
+# If we're running as a legit default user that has an entry in /etc/passwd and a valid homedir, we're all good.
 
-# if current user is in /etc/passwd but $HOME == `/`, rewrite that user's homedir in /etc/passwd to /home/runner and
-# export HOME=/home/runner for this session only- all new sessions, eg created by exec, should set HOME to match the
-# current value in /etc/passwd going forward.
+# If the username/uid we're running under is not represented in /etc/passwd or the current user's homedir is something
+# other than /home/runner (eg, the container was run with --user and some dynamic unmapped UID from the host with
+# primary GID 0), we need to correct that in order for ansible-runner's callbacks to function properly. Some things
+# (eg podman/cri-o today) already create an /etc/passwd entry on the fly in this case, but they set the homedir to
+# WORKDIR, which causes potential collisions with mounted/mapped volumes. For consistency, we'll
+# just always set the current user's homedir to `/home/runner`, which we've already configured in a way
+# that should always work with known container runtimes (eg, ug+rwx and all dirs owned by the root group).
+
+# If current user is not listed in /etc/passwd, add an entry with username==uid, primary gid 0, and homedir /home/runner
+
+# If current user is in /etc/passwd but $HOME != `/home/runner`, rewrite that user's homedir in /etc/passwd to
+# /home/runner and export HOME=/home/runner for this session only. All new sessions (eg podman exec) should
+# automatically set HOME to the value in /etc/passwd going forward.
+
+# Ideally in the future, we can come up with a better way for the outer runner to dynamically inject its callbacks, or
+# rely on the inner runner's copy. This would allow us to restore the typical POSIX user homedir conventions.
 
 # if any of this business fails, we probably want to fail fast
 if [ -n "$EP_DEBUG" ]; then
@@ -23,20 +34,18 @@ else
   set -e
 fi
 
-# FIXME junk output
-if ! getent passwd $(whoami || id -u) ; then
+# current user might not exist in /etc/passwd at all
+if ! $(whoami &> /dev/null) || ! getent passwd $(whoami || id -u) &> /dev/null ; then
   if [ -n "$EP_DEBUG" ]; then
-    echo "hacking missing uid $(id -u) into /etc/passwd"
+    echo "adding missing uid $(id -u) into /etc/passwd"
   fi
   echo "$(id -u):x:$(id -u):0:container user $(id -u):/home/runner:/bin/bash" >> /etc/passwd
   export HOME=/home/runner
 fi
 
-# FIXME junk output
 MYHOME=`getent passwd $(whoami) | cut -d: -f6`
 
-# FIXME: we also want to check the case of a generated user who podman set their homedir to WORKDIR; maybe anything with a high UID, or ?
-if [ "$MYHOME" != "$HOME" ] || [ $(id -u) -ge 500 ] && [ "$MYHOME" != "/home/runner" ]; then
+if [ "$MYHOME" != "$HOME" ] || [ "$MYHOME" != "/home/runner" ]; then
   if [ -n "$EP_DEBUG" ]; then
     echo "replacing homedir for user $(whoami)"
   fi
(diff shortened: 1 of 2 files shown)


nitzmahone:
argh, looks like we'll also need to fix this hack: https://github.com/ansible/ansible-runner/blob/devel/utils/entrypoint.sh#L14-L15

ssbarnea (commented):
I agree that this is a serious bug but at the same time, I doubt that changing room home directory to make it use another folder is ever a good idea.

I do see two paths here: switch to non root user or go for using `/root`.

AlanCoding:
I'd like @shanemcd to be looped into this

nitzmahone:
So would I- I'd love to know the history on this, since the commit history on the container definition isn't telling me much other than "we've dorked around with this an awful lot".

ssbarnea:
I am curious about the history too but not for blaming reasons.

To be honest I would prefer to go for a standard setup:

either:

a) root user with its default home directory
b) non root user with its home directory in either /home/username or /username - and ensuring that HOME == tilde  

Black magic is never a good idea, can hit you back when you expect the least.

AlanCoding:
My own understanding is that there are two requirements which are challenging to satisfy simultaneous (1) rootless podman usage and (2) ability to use in OCP. I would like to note that integration tests are already failing. I would like to have more integration testing with more of the platform, as this is the type of change where we would like to test as much as possible in advance.

shanemcd:
The primary point of complexity here is supporting arbitrary UIDs. We need to support this for 2 main use cases:

- By default, the 'restricted' SCC in OCP will make your pod run with a uid like `1000680000`. 
- When ansible-runner is used with docker, we pass [`--user`](https://github.com/ansible/ansible-runner/blob/1ab1ed5905b9d7b46906b6c40ab7e87ec6256ba7/ansible_runner/config/_base.py#L541) so that files written inside of the EE are visible to the host user.

In response to the changes made so far in this PR:

If we remove `ENV HOME` from the Dockerfile and run as some random uid, HOME will be set to `/`:

```
$ docker run -ti --user 1000680000 quay.io/shanemcd/ansible-runner:homedir_fix bash
bash-4.4$ echo $HOME 
/
```

This causes container group jobs to fail with an error like:

```
  File "/usr/local/bin/ansible-playbook", line 65, in <module>
    import ansible.constants as C
    import ansible.constants as C
  File "/usr/local/lib/python3.8/site-packages/ansible/constants.py", line 180, in <module>
    config = ConfigManager()
  File "/usr/local/lib/python3.8/site-packages/ansible/config/manager.py", line 304, in __init__
    self.update_config_data()
  File "/usr/local/lib/python3.8/site-packages/ansible/config/manager.py", line 604, in update_config_data
    raise AnsibleError("Invalid settings supplied for %s: %s\n" % (config, to_native(e)), orig_exc=e)
ansible.errors.AnsibleError: Invalid settings supplied for DEFAULT_LOCAL_TMP: Unable to create local directories(/.ansible/tmp): [Errno 13] Permission denied: b'/.ansible'
```

One alternative would be to remove `ENV HOME` from the `Dockerfile` and add some conditional logic to the entrypoint before we call [`exec`](https://github.com/ansible/ansible-runner/blob/1ab1ed5905b9d7b46906b6c40ab7e87ec6256ba7/utils/entrypoint.sh#L49). But this would not work if anyone ever wanted to `podman exec` into an EE, because those sessions do not run through the entrypoint.

As awful as it looks, I dont think there would be an issue with changing the root user's home directory as long as `$HOME` and `/etc/passwd` both match. If anyone can point out a scenario where this wouldn't work, I'm always ok with being wrong. But I think it is fairly straight forward to say HOME inside of an EE is _always_ `/home/runner`.

shanemcd:
Tried that here and it seems to work: https://github.com/shanemcd/ansible-runner/pull/2/checks (The failing test was quay.io throwing a 500 error.)

nitzmahone:
I think this should now cover the various cases- still some cleanup to do, it's still spewing noise on startup because I suck at bash, but I think the overall approach should ensure that we've always got a workable homedir that's consistent everywhere, whether it's a dynamic uid or a real user.

shanemcd:
In addition to the noise at startup, the tests are also failing because of how we mount runner's display plugin into the container from the host. This code was also written with the assumption that HOME is always `/home/runner`:

https://github.com/ansible/ansible-runner/blob/1ab1ed5905b9d7b46906b6c40ab7e87ec6256ba7/ansible_runner/utils/__init__.py#L71

shanemcd:
If we want to keep trying to go down this route, I think it would work if we just changed the code to mount the callback under both `/home/runner/.ansible/plugins/callback/` and `/root/.ansible/plugins/callback/`.

shanemcd:
Actually... we can just do this:

```diff
diff --git a/ansible_runner/utils/__init__.py b/ansible_runner/utils/__init__.py
index 214b0a7..c8869e2 100644
--- a/ansible_runner/utils/__init__.py
+++ b/ansible_runner/utils/__init__.py
@@ -68,7 +68,6 @@ def callback_mount(copy_if_needed=False):
     if copy_if_needed is set, and the install is owned by another user,
     it will copy the plugin to a tmpdir for the mount in anticipation of SELinux problems
     '''
-    container_dot_ansible = '/home/runner/.ansible'
     rel_path = ('callback', '',)
     host_path = os.path.join(get_plugin_dir(), *rel_path)
     if copy_if_needed:
@@ -78,7 +77,7 @@ def callback_mount(copy_if_needed=False):
             register_for_cleanup(tmp_path)
             host_path = os.path.join(tmp_path, 'callback')
             shutil.copytree(callback_dir, host_path)
-    container_path = os.path.join(container_dot_ansible, 'plugins', *rel_path)
+    container_path = os.path.join('/usr/share/ansible', 'plugins', *rel_path)
     return (host_path, container_path)
```

shanemcd:
After trying my proposed change locally, some of the tests are now intermittently failing with:

```
[WARNING]: Error accessing plugin paths: [Errno 13] Permission denied:
b'/usr/share/ansible/plugins/callback'
ERROR! Invalid callback for stdout specified: awx_display
```

This might be a race condition because we are using the `:Z` label: 

https://github.com/ansible/ansible-runner/blob/1ab1ed5905b9d7b46906b6c40ab7e87ec6256ba7/ansible_runner/config/_base.py#L502

<br>

From `man podman run`:

> The z option tells Podman that two containers share the volume content. As a result, Podman labels the content with a shared content label. Shared volume labels allow all containers to read/write content.  The Z option tells Podman to label the content with a private unshared label.

shanemcd:
I've confirmed that the tests pass when changing `:Z` to `:z`.

So now, my current concern here is that these latest findings will not only require a change at the EE level, but would also require changes in ansible-runner itself. This would prevent users from being able to use the newer EEs with the versions of ansible-runner 2.x already out in the wild. Unless there is a valid reason about why the more surgical solution I proposed above will not work, I do not think this warrants all of the cascading work that will fall out of making this change.

nitzmahone:
> I do not think this warrants all of the cascading work that will fall out of making this change

One way or another, we still have to fix the HOME/passwd mismatches by the time Ansible runs (which your change alone doesn't)- the current test suite doesn't have any tests that reflect the failure we're actually trying to fix. But yeah, I definitely want to limit the blast radius of whatever this fix ends up looking like to avoid changing the outer runner's volume mount behavior, because IIUC we don't currently have a good way to "sniff" the inner runner version or validate that the assumptions each will make are compatible (something else I have some ideas about- we have to deal with that soon).

There might be some latent weirdness in certain `become` cases if we just make *every* user's homedir `/home/runner` (including root), but it'd certainly be a lot less broken than it is now, and we could probably fix it in core if we hit something. So I can adjust the entrypoint changes to rewrite every user's homedir to `/home/runner` and bomb out/warn/fix+subshell/something if the user's primary gid isn't `0` (which is the only way we can make that work reasonably robustly, and there will always be problems with things explicitly creating things with owner-only modes).

This also works with the `exec` case BTW, at least if we're more aggressive about passwd rewrites in the entrypoint, because it won't let you `exec` under a UID that's not in /etc/passwd, so if we ensure that every user + a potential dynamic uid in `run` are properly reflected in `/etc/passwd`, we never need to set `HOME` at the container level (or if we do it's redundant), but by the time entrypoint has done its thing on the initial `run`, anything that asks for a homedir after that will always get the right answer, no matter how it asks, which solves the original problem.

ziegenberg:
Now that https://github.com/ansible-community/toolset is deprecated, more and more people tend to turn towards https://github.com/ansible/creator-ee, which builds on top of ansible-runner. More issues related to #1024 and this PR are raised. Is there anything that can be done to move things forward?

eqrx:
Hey, can I help with with bringing this forward?:)

utils/entrypoint.sh
```diff
@@ -1,27 +1,60 @@
...
-    # sometime in the future. Seems to be fixed on Fedora 32
-    # If we are running in rootless podman, this file cannot be overwritten
-    ROOTLESS_MODE=$(cat /proc/self/uid_map | head -n1 | awk '{ print $2; }')
-    if [[ "$ROOTLESS_MODE" -eq "0" ]]; then
-cat << EOF > /etc/passwd
-root:x:0:0:root:/root:/bin/bash
-runner:x:`id -u`:`id -g`:,,,:/home/runner:/bin/bash
-EOF
-    fi
+# We need to fix a number of problems here that manifest under different container runtimes, as well as paper over some
```
ssbarnea: Good idea to remove "paper over" idiom, as it it likely to be not known to many readers.

nitzmahone:
@shanemcd the only additional change I can think of right now that might be needed here is to restore the `runner` group and try to force all users into it as a secondary group, but I'm not sure if it's necessary or not... Any other testing you want to do on other container hosts, or anything else?

shanemcd (approved):

PR #1037, opened 2022-04-14
Add stable-2.13 job

GomathiselviS:
Signed-off-by: GomathiselviS <gomathiselvi@gmail.com>

### DIFF
--- a/.zuul.d/jobs.yaml
+++ b/.zuul.d/jobs.yaml
@@ -59,6 +59,53 @@
 # =============================================================================
 
 - job:
+    name: ansible-runner-build-container-image-stable-2.13
+    parent: ansible-build-container-image
+    provides:
+      - ansible-runner-stable-2.13-container-image
+    requires:
+      - python-base-container-image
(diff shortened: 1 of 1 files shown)


goneri (approved):

goneri:
Looks good, but we need to migrate the project on the new platform first.

Shrews (changes requested):
This job does nothing unless you refer to it in the `project.yaml`. Because the SF zuul cannot run these jobs, we've disabled all of them in `check|gate|post`. Only the `periodic` pipeline remains as it appears to still be running on the older zuul.

ryanmerolle:
what is still blocking this?

Shrews:
> what is still blocking this?

This changes the Zuul jobs. Zuul is no longer used for runner CI, therefore this change is pointless.

Closing this issue.

ryanmerolle:
@Shrews So will another PR bring a tag forward for stable-2.13 and eventually stable-2.14?

PR #1049, opened 2022-04-28
Set ZIP strict_timestamps to false

bpedersen2:
Since python 3.8 timestamps before 1980 are disallowed by default,
but seem to occur in streams.

Fixes: awx#12126

Signed-off-by: Björn Pedersen <bjoern.pedersen@frm2.tum.de>

### DIFF
--- a/ansible_runner/utils/streaming.py
+++ b/ansible_runner/utils/streaming.py
@@ -13,7 +13,7 @@ from pathlib import Path
 def stream_dir(source_directory, stream):
     with tempfile.NamedTemporaryFile() as tmp:
         with zipfile.ZipFile(
-            tmp.name, "w", compression=zipfile.ZIP_DEFLATED, allowZip64=True
+            tmp.name, "w", compression=zipfile.ZIP_DEFLATED, allowZip64=True, strict_timestamps=False
         ) as archive:
             if source_directory:
                 for dirpath, dirs, files in os.walk(source_directory):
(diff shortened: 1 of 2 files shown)


bpedersen2:
Hint: As python >=3.8 seems to be the minimum supported version we don't need a feature gate.

AlanCoding:
I wrote a test for this fix: https://github.com/bpedersen2/ansible-runner/pull/1

I confirmed that it fails with the error you give in the issue, and passes with your fix here. We would like to capture test coverage when merging something like this. If you merge that commit or add it manually, this will have my support to merge.

jbradberry (approved):

AlanCoding (approved):

Shrews (approved):

weiyentan:
@bpedersen2 @AlanCoding how do we get these changes? In what tag did the fix come in?
I used this image in my AWX environment
 EE_BASE_IMAGE: 'quay.io/ansible/ansible-runner:stable-2.12-latest' and i am still getting this error when having contents in my collections.

jinja2ninja:
I'm getting the same issue as @weiyentan on stable-2.12-devel

weiyentan:
@fahcsim i am testing with 21

Shrews:
FYI, the images on quay.io are not being updated, at the moment. We are working on a new container image release pipeline, which will likely include a new image tagging scheme. We ask for patience while we develop this process, which we hope to have finished soon. In the meantime, it is easy enough to build your own image based on the current release. For the `2.2` release, something along the lines of:

```
git clone https://github.com/ansible/ansible-runner.git
git checkout -b release_2.2 origin/release_2.2
docker build -t my_tag_name .
```

Use the `release_2.1` branch if you prefer that release.

weiyentan:
Thank you for your confirmation. What will set need to change in the image?

PR #1067, opened 2022-05-06
Update Event emitter plugin docs

Akasurde:
##### SUMMARY

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
docs/external_interface.rst


eqrx (approved):

Shrews:
I would be more in favor of deleting both of these sections than updating them. We do not maintain either of these plugins. Perhaps we could change the doc to just point to them as examples, and let the plugin docs explain how to use them? Not sure of the best way to deal with this, TBH.

Shrews (approved):

PR #1080, opened 2022-05-20
Add execution environment image detail

matburt:
Making this a draft to decide on what kind of data should go in there if it's beyond just the image id. I feel like this is what is needed to support https://github.com/ansible/awx/issues/10409


shanemcd:
I'm a little stuck on the fact this will only work for traditional vm/bare-metal based installs... and even in that case we dont know for sure if we have the image locally before we end up invoking `podman run`.

matburt:
That's fair... there's an alternative workflow here where AWX resolves the proper image before calling the Runner task with a very specific image id where it ends up being known at runtime.

PR #1079, opened 2022-05-19
Refactoring interface common routines

matburt:
This should also fix up direct role execution when run from the command line using an execution environment.

Still need to sort out some tests and make sure the cli entrypoint is doing the right thing in all circumstances.

PR #1084, opened 2022-05-31
skip process isolation check for transmit and process

### LINKED ISSUE #1085: while running via the transmit-worker-process, ansible-runner should only check process isolation exec for worker
When trying to run job via `transmit-worker-process` while using `process_isolation_executable` the kwargs pass into 
`transmit` will contain the `process_isolation=true` this cause `check_isolation_executable_installed` to run during all 3 of the phases (transmit, worker, process) 

https://github.com/ansible/ansible-runner/blob/a349921448f84dc507abbd87aca163291f2c3bdc/ansible_runner/interface.py#L86-L90

In order for the `transmit-worker-process`  chain to use `process_isolation_executable` the argument must be passed to `transmit` and have `transmit` pass the argument down the chain. Therefore we can not just simply remove the argument.

In AWX `ansible-runner transmit` runs on host that `dispatcher` runs while `ansible-runner worker` runs on host that the `receptor` runs

It is possible for the host for `dispatcher` does not have the `process_isolation_executable` (since the job is not actually run on the host) and only the host for `receptor` have the `process_isolation_executable`

https://github.com/ansible/awx/blob/4dc956c76f31accc3f57599d5e6df1735ffa2c16/awx/main/tasks/receptor.py#L391

TheRealHaoLiu:
https://github.com/ansible/ansible-runner/issues/1085

### DIFF
--- a/ansible_runner/interface.py
+++ b/ansible_runner/interface.py
@@ -83,12 +83,6 @@ def init_runner(**kwargs):
         if logfile:
             output.set_logfile(logfile)
 
-    if kwargs.get("process_isolation", False):
-        pi_executable = kwargs.get("process_isolation_executable", "podman")
-        if not check_isolation_executable_installed(pi_executable):
-            print(f'Unable to find process isolation executable: {pi_executable}')
-            sys.exit(1)
-
     event_callback_handler = kwargs.pop('event_handler', None)
     status_callback_handler = kwargs.pop('status_handler', None)
     artifacts_handler = kwargs.pop('artifacts_handler', None)
@@ -117,6 +111,12 @@ def init_runner(**kwargs):
                                          finished_callback=finished_callback,
                                          **kwargs)
             return stream_processor
+
+    if kwargs.get("process_isolation", False):
+        pi_executable = kwargs.get("process_isolation_executable", "podman")
+        if not check_isolation_executable_installed(pi_executable):
+            print(f'Unable to find process isolation executable: {pi_executable}')
+            sys.exit(1)
 
     kwargs.pop('_input', None)
     kwargs.pop('_output', None)
--- a/test/integration/test_transmit_worker_process.py
+++ b/test/integration/test_transmit_worker_process.py
@@ -168,6 +168,66 @@ class TestStreamingUsage:
 
         self.check_artifacts(str(process_dir), job_type)
 
+    def test_process_isolation_executable_not_exist(self, tmp_path, mocker):
+        """Case transmit should not fail if process isolation executable does not exist and
+        worker should fail if process isolation executable does not exist
+        """
+        mocker.patch.object(ansible_runner.interface, 'check_isolation_executable_installed', return_value=False)
+
+        job_kwargs = self.get_job_kwargs('run')
+        job_kwargs['process_isolation'] = True
+        job_kwargs['process_isolation_executable'] = 'does_not_exist'
+
+        outgoing_buffer_file = tmp_path / 'buffer_out'
+        outgoing_buffer_file.touch()
+        outgoing_buffer = outgoing_buffer_file.open('b+r')
+
+        transmitter = ansible_runner.interface.run(
+            streamer='transmit',
+            _output=outgoing_buffer,
+            **job_kwargs,
+        )
+
+        # valide process_isolation kwargs are passed to transmitter
+        assert transmitter.kwargs['process_isolation'] == job_kwargs['process_isolation']
+        assert transmitter.kwargs['process_isolation_executable'] == job_kwargs['process_isolation_executable']
+
+        # validate that transmit did not fail due to missing process isolation executable
+        assert transmitter.rc in (None, 0)
+
+        # validate that transmit buffer is not empty
+        outgoing_buffer.seek(0)
+        sent = outgoing_buffer.read()
+        assert sent  # should not be blank at least
+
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -28,6 +28,7 @@ rpm-build/
 .tox/
 .venv
 venv/
+.vscode/
```
shanemcd: Can you split this out into a different commit?

Shrews: This should be in the user's `$HOME/.gitignore_global` file, IMO, not a per-repo ignore.

ansible_runner/interface.py
```diff
@@ -83,7 +83,7 @@ def init_runner(**kwargs):
         if logfile:
             output.set_logfile(logfile)
 
-    if kwargs.get("process_isolation", False):
+    if kwargs.get("process_isolation", False) and kwargs.get("streamer") not in ('transmit', 'process'):
```
matburt: I have this in my PR also: https://github.com/ansible/ansible-runner/pull/1079/files#diff-179ccecedba1c6ccf73b8c02f0321b252c018226e41ce82887a93ace762d2fbbR55

This line causes an astonishing amount of problems when interfacing with Runner from the CLI

matburt:
Fix that vscode comment and then I think we should merge this asap.

ansible_runner/interface.py
matburt: I think the reason it does this has to do with the rest of my PR, which is that the CLI entrypoint clashes with the module level entrypoint when managing process isolation paths.

My work on #1079  focuses on realigning those two paths.

TheRealHaoLiu:
`.gitignore` change amended as per suggestion

TheRealHaoLiu:
code change after peer programing with @jbradberry moved the location of the check

jbradberry (approved):

Shrews:
Manually tested this to verify that `worker` will bail if the executable is not found. I'd like to have a test for this before merging it. Thanks!

TheRealHaoLiu:
> Manually tested this to verify that `worker` will bail if the executable is not found. I'd like to have a test for this before merging it. Thanks!

testcase added

test/integration/test_transmit_worker_process.py
```diff
@@ -168,6 +168,54 @@ def process_method(results_sockfile_read):
...
+            _output=outgoing_buffer,
+            **job_kwargs,
+        )
+
+        # valide process_isolation kwargs are passed to transmitter
+        assert transmitter.kwargs['process_isolation'] == job_kwargs['process_isolation']
+        assert transmitter.kwargs['process_isolation_executable'] == job_kwargs['process_isolation_executable']
+
+        # validate that transmit did not fail due to missing process isolation executable
+        assert transmitter.rc in (None, 0)
```
Shrews, line 196: Is `None` really a valid value here?

test/integration/test_transmit_worker_process.py
```diff
@@ -168,6 +168,54 @@ def process_method(results_sockfile_read):
...
+        assert b'kwargs' in sent
+
+        # validate kwargs in buffer contain correct process_isolation and process_isolation_executable
+        for line in sent.decode('utf-8').split('\n'):
+            if "kwargs" in line:
+                kwargs = json.loads(line).get("kwargs", {})
+                assert kwargs['process_isolation'] is True
+                assert kwargs['process_isolation_executable'] == process_isolation_executable
+                break
+
```
Shrews, line 213: Similar to the `transmit` check above, you should check that `worker` will fail if the executable is missing.

Shrews (changes requested):

test/integration/test_transmit_worker_process.py
```diff
@@ -168,6 +168,54 @@ def process_method(results_sockfile_read):
...
         self.check_artifacts(str(process_dir), job_type)
 
+    @pytest.mark.parametrize("process_isolation_executable", ['exist', 'does_not_exist'])
+    def test_process_isolation_executable_transmit(self, tmp_path, process_isolation_executable, mocker):
+        """
+        Case transmit should not fail if process isolation executable does not exist
+        """
+        def mock_check_isolation_executable_installed(executable):
+            if executable == "exist":
+                return True
```
Shrews:
```suggestion
                return True
            return False
```

test/integration/test_transmit_worker_process.py
TheRealHaoLiu: https://github.com/ansible/ansible-runner/blob/devel/test/integration/test_transmit_worker_process.py#L74-L75

TheRealHaoLiu:
testcase commit amended base on @Shrews's feedback

Shrews (approved):
Thanks for the PR! If you want this in any of the previously released 2.x versions, you'll need to supply a backport to the appropriate release branches. We follow the same basic backport process as Ansible (https://docs.ansible.com/ansible/latest/community/development_process.html#backport-process).

TheRealHaoLiu:
> Thanks for the PR! If you want this in any of the previously released 2.x versions, you'll need to supply a backport to the appropriate release branches. We follow the same basic backport process as Ansible (https://docs.ansible.com/ansible/latest/community/development_process.html#backport-process).

no need for backport thanks

TheRealHaoLiu:
backport to 2.2 https://github.com/ansible/ansible-runner/pull/1094

PR #1089, opened 2022-06-03
update callback_dir copy and mounting behavior

### LINKED ISSUE #1088: fail to create container when ansible-runner is installed in /usr
When using process_isolation_executable (in this specific case `podman`) via `ansible-runner run` and `ansible-runner worker` 
ansible-runner will mount `ansible_runner/display_callback/callback` on the host to `/home/runner/.ansible/plugins/callback` in the container 

https://github.com/ansible/ansible-runner/blob/c143c07a9355cf4f598717ad853bb4fd008dbca7/ansible_runner/config/_base.py#L508-L509

When ansible-runner is installed in /usr directory (in this specific case `/usr/lib/python3.9/site-packages/ansible_runner`) ansible-runner fail to create container due to not able to mount from `/usr` directory (forbidden by podman)

https://github.com/ansible/ansible-runner/blob/c143c07a9355cf4f598717ad853bb4fd008dbca7/ansible_runner/utils/__init__.py#L73

`copy_if_needed` param does not trigger due if `/usr` directory is own by current user (when running as root)

error log:
```
Error: error preparing container 382fd4e33f875ba82753d6c1969d2caa4c32a38045ac1faed66948c04507cc85 for attach: relabeling content in /usr is not allowed
```

job_args
```
podman run
  --rm
  --tty
  --interactive 
  --workdir /runner/project 
  --env-file /var/lib/awx/job_execution/awx_59_1l624cim/artifacts/59/env.list 
  --quiet 
  --name ansible_runner_59 
  --user=root
  --network slirp4netns:enable_ipv6=true 
  -v /var/lib/awx/job_execution/awx_59_1l624cim/:/runner/:Z 
  -v /usr/lib/python3.9/site-packages/ansible_runner/display_callback/callback/:/home/runner/.ansible/plugins/callback/:Z 
  -v /etc/pki/ca-trust/:/etc/pki/ca-trust/:O 
  -v /usr/share/pki/:/usr/share/pki/:O 
  -v /var/lib/awx/projects/_6__demo_project/:/var/lib/awx/projects/_6__demo_project/:z 
  -v /var/lib/awx/projects/.__awx_cache/_6__demo_project/:/var/lib/awx/projects/.__awx_cache/_6__demo_project/:z 
  brew.registry.redhat.io/rh-osbs/ansible-automation-platform-22-ee-supported-rhel8@sha256:472c8b36f517c7d074d3ef3487f128e18baa8b363ea18da4a299ef9ac52566f9 
  ansible-playbook -t update_git,install_roles,install_collections 
    -i /runner/inventory/hosts 
    -e @/runner/env/extravars project_update.yml
```

 offending line that cause `podman run` failure
```
   -v /usr/lib/python3.9/site-packages/ansible_runner/display_callback/callback/:/home/runner/.ansible/plugins/callback/:Z 
```

TheRealHaoLiu:
https://github.com/ansible/ansible-runner/issues/1088

- enhance callback_mount copy_if_needed check to include unmountable location `/usr`
- add option to specify location when copying callback plugin
- if callback dir copy is needed copy to artifact directory rather than /tmp

Signed-off-by: Hao Liu <haoli@redhat.com>

### DIFF
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -505,7 +505,7 @@ class BaseConfig(object):
         self._update_volume_mount_paths(new_args, "{}".format(self.private_data_dir), dst_mount_path="/runner", labels=":Z")
 
         # Mount the stdout callback plugin from the ansible-runner code base
-        mount_paths = callback_mount(copy_if_needed=True)
+        mount_paths = callback_mount(copy_if_needed=True, copy_dir="{}/artifacts".format(self.private_data_dir))
         self._update_volume_mount_paths(new_args, mount_paths[0], dst_mount_path=mount_paths[1], labels=":Z")
 
         if self.container_auth_data:
--- a/ansible_runner/utils/__init__.py
+++ b/ansible_runner/utils/__init__.py
@@ -61,22 +61,26 @@ def is_dir_owner(directory):
     return bool(current_user == callback_owner)
 
 
-def callback_mount(copy_if_needed=False):
+def callback_mount(copy_if_needed=False, copy_dir=None):
     '''
     Return a tuple that gives mount points for the standard out callback
     in the form of (<host location>, <location in container>)
-    if copy_if_needed is set, and the install is owned by another user,
-    it will copy the plugin to a tmpdir for the mount in anticipation of SELinux problems
+    if copy_if_needed is set, and the install is owned by another user
+    or in location that is not mountable such as /usr,
+    it will copy the plugin to a tmpdir or copy_dir for the mount in anticipation of SELinux problems
     '''
     container_dot_ansible = '/home/runner/.ansible'
     rel_path = ('callback', '',)
     host_path = os.path.join(get_plugin_dir(), *rel_path)
     if copy_if_needed:
         callback_dir = get_callback_dir()
-        if not is_dir_owner(callback_dir):
-            tmp_path = tempfile.mkdtemp(prefix='ansible_runner_plugins_')
-            register_for_cleanup(tmp_path)
-            host_path = os.path.join(tmp_path, 'callback')
+        if callback_dir.startswith('/usr') or not is_dir_owner(callback_dir):
+            if copy_dir is None:
+                tmp_path = tempfile.mkdtemp(prefix='ansible_runner_plugins_')
+                register_for_cleanup(tmp_path)
+                host_path = os.path.join(tmp_path, 'callback')
+            else:
+                host_path = os.path.join(copy_dir, 'callback')
             shutil.copytree(callback_dir, host_path)
     container_path = os.path.join(container_dot_ansible, 'plugins', *rel_path)
     return (host_path, container_path)


ansible_runner/utils/__init__.py
```diff
@@ -61,22 +61,26 @@ def is_dir_owner(directory):
...
     container_dot_ansible = '/home/runner/.ansible'
     rel_path = ('callback', '',)
     host_path = os.path.join(get_plugin_dir(), *rel_path)
     if copy_if_needed:
         callback_dir = get_callback_dir()
-        if not is_dir_owner(callback_dir):
-            tmp_path = tempfile.mkdtemp(prefix='ansible_runner_plugins_')
-            register_for_cleanup(tmp_path)
-            host_path = os.path.join(tmp_path, 'callback')
+        if callback_dir.startswith('/usr') or not is_dir_owner(callback_dir):
```
TheRealHaoLiu, line 77: add check for if the source is from an unmountable dir (`/usr`)

ansible_runner/utils/__init__.py
```diff
@@ -61,22 +61,26 @@ def is_dir_owner(directory):
...
+            if copy_dir is None:
+                tmp_path = tempfile.mkdtemp(prefix='ansible_runner_plugins_')
+                register_for_cleanup(tmp_path)
+                host_path = os.path.join(tmp_path, 'callback')
+            else:
+                host_path = os.path.join(copy_dir, 'callback')
```
TheRealHaoLiu, line 83: does not change how `callback_mount` when call like before
`callback_mount(copy_if_needed=False)` will still behave like before

ansible_runner/config/_base.py
```diff
@@ -505,7 +505,7 @@ def wrap_args_for_containerization(self, args, execution_mode, cmdline_args):
         self._update_volume_mount_paths(new_args, "{}".format(self.private_data_dir), dst_mount_path="/runner", labels=":Z")
 
         # Mount the stdout callback plugin from the ansible-runner code base
-        mount_paths = callback_mount(copy_if_needed=True)
+        mount_paths = callback_mount(copy_if_needed=True, copy_dir="{}/artifacts".format(self.private_data_dir))
```
TheRealHaoLiu, line 508: `$private_data_dir/artifacts` is a good candidate for copying the callback plugin to for the following reasons
1. clean up after run like tmpdir so no accumulation of garbage
2. have to be (and is therefore always) mountable by the container that are created

ansible_runner/utils/__init__.py
```diff
@@ -61,22 +61,26 @@ def is_dir_owner(directory):
...
     host_path = os.path.join(get_plugin_dir(), *rel_path)
     if copy_if_needed:
         callback_dir = get_callback_dir()
-        if not is_dir_owner(callback_dir):
-            tmp_path = tempfile.mkdtemp(prefix='ansible_runner_plugins_')
-            register_for_cleanup(tmp_path)
-            host_path = os.path.join(tmp_path, 'callback')
+        if callback_dir.startswith('/usr') or not is_dir_owner(callback_dir):
+            if copy_dir is None:
+                tmp_path = tempfile.mkdtemp(prefix='ansible_runner_plugins_')
```
AlanCoding, line 79: Probably better to get rid of this entirely. You have a good case that the artifacts dir is a better location, and if it doesn't cause any problems with testing, then we should use that location in all cases, and I'd say `copy_dir` shouldn't be an argument here.

ansible_runner/utils/__init__.py
AlanCoding: I'm curious if we could have some more descriptive criteria than `.startswith`. What is it about the /usr folder that makes it un-mountable, and can we check that criteria more directly?

ansible_runner/utils/__init__.py
TheRealHaoLiu: what is the best way to get the correct private_data_dir from the context of this function?

reason why i preserve the previous method signature is because i do not know what else is relying on this method (internal or imported by other project)

TheRealHaoLiu:
https://github.com/ansible/ansible-runner/pull/1093 alternative fix proposal

TheRealHaoLiu:
closed due to alternative proposal https://github.com/ansible/ansible-runner/pull/1093  being accepted

PR #1094, opened 2022-06-06
[backport][2.2] skip process isolation check for transmit and process

TheRealHaoLiu:
backport of https://github.com/ansible/ansible-runner/pull/1084

skip process isolation check for transmit and process

moving down the process isolation executable so that transmit and
process no longer checks for its existance

ansible-runner worker will still run the check since it will invoke
ansible-runner run with process-isolation-executable and it will still
be checked properly

Co-Authored-By: Jeff Bradberry <685957+jbradberry@users.noreply.github.com>
Signed-off-by: Hao Liu <haoli@redhat.com>

### DIFF
--- a/ansible_runner/interface.py
+++ b/ansible_runner/interface.py
@@ -83,12 +83,6 @@ def init_runner(**kwargs):
         if logfile:
             output.set_logfile(logfile)
 
-    if kwargs.get("process_isolation", False):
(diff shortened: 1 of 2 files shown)


Shrews:
Thanks.

Please use `git cherry-pick -x` for future backports. Example of this is in the [ansible backporting docs](https://docs.ansible.com/ansible/latest/community/development_process.html#backporting-merged-prs-in-ansible-core).

PR #1105, opened 2022-07-19
Enable pylint

ssbarnea:
This is fixes two basic problems while enabling pylint. All the other
rules that are excluded are because existing violations. These are
supposed to be gradually fixed while removing them from the 'disable'
list.


mjtrangoni:
@ssbarnea @eqrx what was the triage here? Is pylint not desired?

Akasurde:
Pylint is enabled with recent changes, closing this.

PR #1106, opened 2022-07-20
Consolidate how we run linters

ssbarnea:
Ensure we use pre-commit as linter orchestrator as this ensures that
we always use the same linter version.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,3 +4,9 @@ repos:
     rev: v2.0.0
     hooks:
       - id: flake8
+  - repo: https://github.com/adrienverge/yamllint
+    rev: v1.27.1
+    hooks:
+      - id: yamllint
+        args:
+          - --strict
--- a/tox.ini
+++ b/tox.ini
@@ -19,11 +19,10 @@ commands = pytest {posargs}
 [testenv:linters]
 description = Run code linters
 basepython = python3.8
+deps =
+    pre-commit
 commands=
-    flake8 --version
-    flake8 docs ansible_runner test
-    yamllint --version
-    yamllint -s .
+    pre-commit run -a
 
 [testenv:unit{,-py38,-py39,-py310}]
 description = Run unit tests


Shrews:
What is this "consolidating" with?  I'm not sure I see the improvement here.  If the only reason is to guarantee we always use the same version (and I'm not convinced we should), we can just set the version in `test/requirements.txt`.

ssbarnea:
First, requirements.txt is for runtime requirements, not test ones. pre-commit takes care of version locking itself and can automatically make pull requests to update the linters, see how it works on devtools project, example https://github.com/ansible/ansible-lint/pull/2229

Also you do not want to get different results when you run 'pre-commit' or 'tox -e lint'. At this moment that is what it happens.

You can ask @cidrblock or we can even setup a meeting, I would love to explain why we ended up with that approach.

In regards to pinning test dependencies we also went for a very interesting approach, using contraint files and having dependabot and/or pip-compile update them and create pull-requests for them. The main advantage of this is that the release of breaking dependency will not break your CI/CD pipelines, it will only produce a pull-request to update contraints that fails, but other pending changes will not be affected by it.

Shrews:
> First, requirements.txt is for runtime requirements, not test ones. pre-commit takes care of version locking itself and can automatically make pull requests to update the linters, see how it works on devtools project, example [ansible/ansible-lint#2229](https://github.com/ansible/ansible-lint/pull/2229)

That's not accurate. Our `tox.ini` installs the requirements from `test/requirements.txt` into the test tox environment before the tests run. The top-level `requirements.txt` file (also installed to each tox test environment) is for the runtime requirements.

> 
> Also you do not want to get different results when you run 'pre-commit' or 'tox -e lint'. At this moment that is what it happens.

I'm not sure why you see that. We don't tend to use any pre-commits in our workflow, so none of the Executor team has experienced that.  If it's because of the existing `.pre-commit-config.yaml` file in our repo, I'd be more in favor of just removing that file. Not even sure why that exists, tbh.

> 
> You can ask @cidrblock or we can even setup a meeting, I would love to explain why we ended up with that approach.
> 
> In regards to pinning test dependencies we also went for a very interesting approach, using contraint files and having dependabot and/or pip-compile update them and create pull-requests for them. The main advantage of this is that the release of breaking dependency will not break your CI/CD pipelines, it will only produce a pull-request to update contraints that fails, but other pending changes will not be affected by it.

Appreciate the info on the devtools workflow. Maybe a meeting between the Executor and Devtools teams should be done first if you want to explain your workflow and encourage Executor to adopt it. I'm not sure that just submitting PRs to change the current Executor workflow is the best way to proceed, without first having that discussion.

@Akasurde @eqrx As the other members of the Executor team, your input on this would be appreciated.

eqrx (dismissed):

.pre-commit-config.yaml
```diff
@@ -4,3 +4,9 @@ repos:
     rev: v2.0.0
     hooks:
       - id: flake8
+  - repo: https://github.com/adrienverge/yamllint
```
eqrx, line 7: I would agree with @Shrews and remove this file altogether since it is not used

eqrx (changes requested):
Yeah, would agree that having a discussion about this would be best

Akasurde:
I agree with @Shrews.

PR #1112, opened 2022-08-09
Linting fixes

Akasurde:
##### SUMMARY

* Added whitespace after the keyword
* Pinned `flake8` to 4.0.1 for now
* Commented `ENV PIP_OPTS=--no-build-isolation` 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ansible_runner/__main__.py
ansible_runner/cleanup.py

### DIFF
--- a/ansible_runner/cleanup.py
+++ b/ansible_runner/cleanup.py
@@ -79,9 +79,9 @@ def is_alive(dir):
 
     try:
         os.kill(pid, signal.SIG_DFL)
-        return(0)
+        return 0
     except OSError:
-        return(1)
+        return 1
 
 
 def project_idents(dir):
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,9 @@ RUN if [ "$ANSIBLE_BRANCH" != "" ] ; then \
 # NOTE(pabelanger): For downstream builds, we compile everything from source
 # over using existing wheels. Do this upstream too so we can better catch
 # issues.
-ENV PIP_OPTS=--no-build-isolation
+
+# Commenting due to failures in CI
+# ENV PIP_OPTS=--no-build-isolation
 RUN assemble
 
 FROM $PYTHON_BASE_IMAGE
--- a/ansible_runner/__main__.py
+++ b/ansible_runner/__main__.py
@@ -869,21 +869,21 @@ def main(sys_args=None):
                     else:
                         sys.stderr.write(exc)
                     return 1
-            return(res.rc)
+            return res.rc
 
     try:
         with open(pidfile, 'r') as f:
             pid = int(f.readline())
     except IOError:
-        return(1)
+        return 1
 
     if vargs.get('command') == 'stop':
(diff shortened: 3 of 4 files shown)


Shrews:
Remove the `pypsrp[kerberos,credssp]` and `pywinrm[kerberos,credssp]` entries from `tools/requirements.txt` to get the integration tests working and we'll merge these two fixes as one change. Those unneeded entries were pulling in `pykrb5` which now requires `Cython`, thus causing the errors.

I'll work on removing other unneeded entries in that and other requirements files for us in a separate PR, if any.

Shrews (dismissed):
This will probably require a backport. Not sure if that will cause issues.

tools/requirements.txt
```diff
@@ -2,8 +2,6 @@ ansible-core
 ncclient
 paramiko
 pyOpenSSL<20.0.0  # Transient dependency for credssp support.
-pypsrp[kerberos,credssp]
-pywinrm[kerberos,credssp]
```
Shrews: Although this fixes the immediate problem with integration tests, some EEs based on our image might be depending on having these packages available. Therefore, we can't remove them.

Speaking with nitz on this, he's identified a line in our `Dockerfile` that seems to be causing the issue. If you comment out the `ENV PIP_OPTS=--no-build-isolation` line in `Dockerfile`, the tests should succeed. The tests will take longer now, but at least we won't break anyone.

Shrews (changes requested):

Shrews:
Have you identified what changed (flake8 update maybe?) to cause these linting issues to appear? It might be worth considering pinning the linting tools to a specific version to prevent this from happening (and then update the version on major releases).

ansible_runner/cleanup.py
```diff
@@ -79,9 +79,9 @@ def is_alive(dir):
 
     try:
         os.kill(pid, signal.SIG_DFL)
```
eqrx, line 81: While we are working on linting, could we please please rework this? in this line the feature of glibc is used that signal 0 checks if the caller has permissions to send a signal to the PID. for some reason somebody passed the default signal handler which evaluates `0` in there.

so `os.kill(pid, 0)` would be more correct but the problem still remains that this line uses a feature that is NOT part of the `os.kill` contract and it checks for permission of the pid, not existence. (so if the process does exist but runner is not allowed to send a signal which can happen for multiple reasons, it just assumes it is not there)

eqrx (dismissed):

ansible_runner/cleanup.py
Shrews: Let's not try to fix random things within this PR. The linting change will likely have to be backported to get our CI working correctly across all branches. I'm not convinced a change to this line should be backported yet.

eqrx: fair!

Shrews (approved):
LGTM. Probably needs backported.

PR #1123, opened 2022-08-24
Make sure we close stdout/stderr in Runner.run()

Shrews:
If we treat warnings as errors with pytest (via `-Werror` or config file change), tests using `subprocess` code path will fail because we get a warning about an unclosed file on `stderr`.

Ultimately, the entire `Runner.run()` method probably needs reworked so that it is not so complex, and we can use better/safer coding methods, such as context managers on the open file handles. As it is now, using context managers on these file handles is a major pain since there is so much code between opening them and where we need to guarantee they are closed.

Provided test will fail without the code changes to close `stdout`/`stderr`.

### DIFF
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -276,9 +276,7 @@ class Runner(object):
                     stderr_response = stderr_response.decode()
                 stderr_handle.write(stderr_response)
 
-            stdout_handle.flush()
             stdout_handle.close()
-            stderr_handle.flush()
             stderr_handle.close()
         else:
             try:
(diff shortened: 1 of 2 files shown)


ansible_runner/runner.py
```diff
@@ -275,6 +275,11 @@ def run(self):
                 if isinstance(stderr_response, bytes):
                     stderr_response = stderr_response.decode()
                 stderr_handle.write(stderr_response)
+
+            stdout_handle.flush()
```
eqrx: Is a flush needed right before a close?

Shrews: Fair point. I was doing a blind copy-paste of existing code there. The `flush()` call shouldn't be needed, so I'll remove those lines.

eqrx (approved):
Good catch, thanks for doing this:)

PR #1133, opened 2022-09-16
remove any pipe files before zip/unzipping

fosterseth:
We are seeing cases where unstream_dir() is hanging when attempting to read `ssh_key_data`. This file is supposed to be cleaned up by this point, but maybe not if things go awry.

This change just looks for files that are type "pipe" and removes them.

To test I changed the "rm" command here to "ls", so the pipe is never removed.

```python3
ssh_key_cleanup_command = 'rm -f {}'.format(ssh_key_path)
```
https://github.com/ansible/ansible-runner/blob/devel/ansible_runner/config/_base.py#L612

Project updates can still succeed, no hangs.

### DIFF
--- a/ansible_runner/utils/streaming.py
+++ b/ansible_runner/utils/streaming.py
@@ -22,11 +22,6 @@ def stream_dir(source_directory, stream):
                         relpath = ""
                     for fname in files + dirs:
                         full_path = os.path.join(dirpath, fname)
-                        if stat.S_ISFIFO(os.stat(full_path).st_mode):
-                            # skip any pipes, as python hangs when attempting
-                            # to read them.
-                            # i.e. ssh_key_data that was never cleaned up
-                            continue
                         # Magic to preserve symlinks
                         if os.path.islink(full_path):
                             archive_relative_path = os.path.relpath(dirpath, source_directory)
@@ -37,6 +32,11 @@ def stream_dir(source_directory, stream):
                             permissions |= 0xA000
                             zip_info.external_attr = permissions << 16
                             archive.writestr(zip_info, os.readlink(full_path))
+                        elif stat.S_ISFIFO(os.stat(full_path).st_mode):
+                            # skip any pipes, as python hangs when attempting
+                            # to open them.
+                            # i.e. ssh_key_data that was never cleaned up
+                            continue
                         else:
                             archive.write(
                                 os.path.join(dirpath, fname), arcname=os.path.join(relpath, fname)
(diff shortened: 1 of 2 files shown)


ansible_runner/utils/streaming.py
```diff
@@ -22,6 +22,11 @@ def stream_dir(source_directory, stream):
                         relpath = ""
                     for fname in files + dirs:
                         full_path = os.path.join(dirpath, fname)
+                        if stat.S_ISFIFO(os.stat(full_path).st_mode):
+                            # must remove any pipes before reading
+                            # i.e. ssh_key_data that was never cleaned up
+                            os.remove(full_path)
```
nitzmahone: Any reason we need to actually *delete* the fifo instead of just skipping it for the zip? Neither the stream nor unstream directories are guaranteed to be exclusively "ours", so who knows what might be laying around in there that something else is relying on...

ansible_runner/utils/streaming.py
```diff
@@ -79,6 +84,11 @@ def unstream_dir(stream, length, target_directory):
 
                 is_symlink = mode[:1] == 'l'
                 if os.path.exists(out_path):
+                    if stat.S_ISFIFO(os.stat(out_path).st_mode):
+                        # must remove any pipes before reading
+                        # i.e. ssh_key_data that was never cleaned up
+                        os.remove(out_path)
```
nitzmahone, line 93: This one makes more sense to delete as a safety, esp since it will only trigger when the archive contains a file with the same name/path- seems a lot less likely to stomp an arbitrary socket file laying around that might be in use...

nitzmahone (changes requested):
LGTM, other than deleting any socket file encountered in the send tree seems a little draconian (rather than just skipping it).

ansible_runner/utils/streaming.py
fosterseth: changed so that it skips instead of removes. We still remove in the unstream_dir though.

ansible_runner/utils/streaming.py
```diff
@@ -22,6 +22,11 @@ def stream_dir(source_directory, stream):
                         relpath = ""
                     for fname in files + dirs:
                         full_path = os.path.join(dirpath, fname)
+                        if stat.S_ISFIFO(os.stat(full_path).st_mode):
```
Akasurde: Could you please add a test for this change?

AlanCoding:
I am a little worried about performance of running python stat on every file in a large repo, and it might be good to characterize the time cost of doing that compared to the total cost of extracting the zip data.

fosterseth:
to-do: move logic below the symlink checks, otherwise we get a file not found error

AlanCoding:
https://gist.github.com/AlanCoding/90708be38abab46d594ee18a729e86f2

Initial data:

```
Took 0.021636962890625 to crawl all of .
Took 0.059778451919555664 to crawl all of . while checking for symlinks
That includes 10133 objects
Took 0.07620024681091309 to crawl all of . with stat
Took 2.5405492782592773 to actually stream to in-memory buffer
```

Idea here is that this PR adds the "stat" computation. The symlink check (in addition to being necessary for this work) is pre-existing. The added time from the stat comes to `0.01642`

Compared to the total streaming time, that's like `0.01642/2.5405492782592773`, coming to about 0.6% of total time. This will be higher for repos with more small files, and smaller for repos with fewer large files.

AlanCoding (approved):

AlanCoding:
https://github.com/fosterseth/ansible-runner/pull/1

Here are 2 proposed tests for the cases being fixed here. Obviously, I confirmed that they fail in devel, those are the failures I give in that link.

I have not confirmed that your work here makes them pass, but obviously, if the fix works, it should.

AlanCoding:
Looks good, we did integration testing with AWX as well and that came back looking fine.

nitzmahone (approved):

PR #1136, opened 2022-09-28
Add stable-2.13 container builds

twouters:
I just copied 2.12 stuff and changed it to 2.13 and verified with the changes to enable stable-2.12 so this isn't actually tested.

There's another pull request but it appears abandoned.

I'll prepare a PR for 2.14 too in case this one gets accepted.

### DIFF
--- a/.zuul.d/jobs.yaml
+++ b/.zuul.d/jobs.yaml
@@ -59,6 +59,53 @@
 # =============================================================================
 
 - job:
+    name: ansible-runner-build-container-image-stable-2.13
+    parent: ansible-build-container-image
+    provides:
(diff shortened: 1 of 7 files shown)


Shrews:
Zuul is not used with the `ansible-runner` repo any more. As such, container building is temporarily disabled. We are currently discussing our container needs and attempting to identify a new container building methodology.

fsdrw08:
> Zuul is not used with the `ansible-runner` repo any more. As such, container building is temporarily disabled. We are currently discussing our container needs and attempting to identify a new container building methodology.

any update about the new version container?

PR #1148, opened 2022-10-25
Remove --ipc=host podman option

### LINKED ISSUE #984: `ipc=host` for podman fails when `/dev/mqueue` does not exist
Found while troubleshooting an ansible-navigator issue on WSL2/Ubuntu https://github.com/ansible/ansible-navigator/issues/610

(there is no /dev/mqueue by default on Ubuntu for WSL2)

This is half issue/half question.... 

Is it used by runner? (some basic playbook runs in navigator work fine with https://github.com/ansible/ansible-runner/blob/88130033e6b4791952236a8b1eb6f7ac823649e3/ansible_runner/config/_base.py#L473 commented out, but the testing was not extensive)
Can it be removed?
Can it be added as a parameter for navigator to use?

In the mean time, we can exit navigator and ask the user to build the directory.

Shrews:
Fixes #984

This removes the `--ipc=host` option from the `podman` command line. This does not seem to be needed, possibly an artifact of a long-removed feature, and adds an unnecessary requirement of having an existing `/dev/mqueue`.

### DIFF
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -490,7 +490,6 @@ class BaseConfig(object):
             if 'podman' in self.process_isolation_executable:
                 # container namespace stuff
                 new_args.extend(["--group-add=root"])
-                new_args.extend(["--ipc=host"])
 
             self._ensure_path_safe_to_mount(self.private_data_dir)
             # Relative paths are mounted relative to /runner/project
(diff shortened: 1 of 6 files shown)


Shrews:
DO NOT MERGE without approval from Controller team.

AlanCoding:
I would like @shanemcd :eyes: on this. I will try to cross-reference against how we're doing this in AWX.

Shrews:
Closing this, will need an alternate solution. Confirmed the option is needed with SELinux enabled hosts and the SSH_AUTH_SOCK file is on a separately mounted volume.

PR #1151, opened 2022-11-01
Add 'skip_ipc_opt' API parameter to avoid --ipc=host option on podman run command

### LINKED ISSUE #984: `ipc=host` for podman fails when `/dev/mqueue` does not exist

Shrews:
The `--ipc=host` option to `podman run` is added to bypass some issues accessing SSH_AUTH_SOCK on systems running SELinux. This can cause issues on systems without a `/dev/mqueue` device. We add a new API parameter called `skip_ipc_opt` that can be set to `True` to avoid using the `--ipc` option.

Fixes #984

### DIFF
--- a/ansible_runner/interface.py
+++ b/ansible_runner/interface.py
@@ -203,7 +203,7 @@ def run(**kwargs):
     :param bool check_job_event_data: Check if job events data is completely generated. If event data is not completely generated and if
                                  value is set to 'True' it will raise 'AnsibleRunnerException' exception,
                                  if set to 'False' it log a debug message and continue execution. Default value is 'False'
-    :param bool skip_ipc_opt: Disables the --ipc=host option to podman run command.
+    :param bool skip_ipc_opt: Disables the --ipc=host option to podman run command. Default value is 'False'.
 
     :returns: A :py:class:`ansible_runner.runner.Runner` object, or a simple object containing ``rc`` if run remotely
     '''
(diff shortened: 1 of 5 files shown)


ansible_runner/interface.py
```diff
@@ -203,6 +203,7 @@ def run(**kwargs):
     :param bool check_job_event_data: Check if job events data is completely generated. If event data is not completely generated and if
                                  value is set to 'True' it will raise 'AnsibleRunnerException' exception,
                                  if set to 'False' it log a debug message and continue execution. Default value is 'False'
+    :param bool skip_ipc_opt: Disables the --ipc=host option to podman run command.
```
Akasurde:
```suggestion
    :param bool skip_ipc_opt: Disables the --ipc=host option to podman run command. The default value is 'False'.
```

Shrews:
Tests in this change are not actually running due to issue #1153

Shrews:
@nitzmahone has suggested he'd rather see us move toward a more lightweight probing model, rather than an API change, for this type of stuff. Seems like probably a more elegant way to handle this, so closing this PR.

PR #1165, opened 2022-11-17
Don't flush fileobj

### LINKED ISSUE #1164: ANSIBLE_VERBOSITY>=3 does not work with ansible 2.14 and ansible-runner
Playbook run gets stuck when running with ansible 2.14 and ansible-runner 2.3.1 with verbosity>=3. This seems to be due to a regression in  awx_display.py after ansible PR https://github.com/ansible/ansible/pull/77056. 

```
$cat test.yaml 
- hosts: all
  connection: local
  tasks:
    - name: check file
      ansible.builtin.file:
        path: /etc/foo.conf
        state: absent

$ cat inventory.yaml
[all]
test-[1:48].example.com

$export ANSIBLE_INVENTORY=inventory.yaml
$export ANSIBLE_VERBOSITY="3"
$ansible-runner run -p test.yaml $HOME
```
```
traceback:
.....
<test-19.example.com> EXEC /bin/sh -c 'echo ~ramishra && sleep 0'
Using module file /home/ramishra/.local/lib/python3.10/site-packages/ansible/modules/file.py
<test-19.example.com> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo /home/ramishra/.ansible/tmp `"&& mkdir "` echo /home/ramishra/.ansible/tmp/ansible-tmp-1668673584.1542976-1435398-231198689550583 `" && echo ansible-tmp-1668673584.1542976-1435398-231198689550583="` echo /home/ramishra/.ansible/tmp/ansible-tmp-1668673584.1542976-1435398-231198689550583 `" ) && sleep 0'
<test-17.example.com> PUT /home/ramishra/.ansible/tmp/ansible-local-14328044ytahafl/tmpawosdm87 TO /home/ramishra/.ansible/tmp/ansible-tmp-1668673584.0746303-1435377-8803172912834/AnsiballZ_file.py
<test-17.example.com> EXEC /bin/sh -c 'chmod u+x /home/ramishra/.ansible/tmp/ansible-tmp-1668673584.0746303-1435377-8803172912834/ /home/ramishra/.ansible/tmp/ansible-tmp-1668673584.0746303-1435377-8803172912834/AnsiballZ_file.py && sleep 0'
<test-16.example.com> EXEC /bin/sh -c '/usr/bin/python3 /home/ramishra/.ansible/tmp/ansible-tmp-1668673584.0498953-1435368-188608545559309/AnsiballZ_file.py && sleep 0'
<test-20.example.com> ESTABLISH LOCAL CONNECTION FOR USER: ramishra
<test-20.example.com> EXEC /bin/sh -c 'echo ~ramishra && sleep 0'
Using module file /home/ramishra/.local/lib/python3.10/site-packages/ansible/modules/file.py
<test-21.example.com> ESTABLISH LOCAL CONNECTION FOR USER: ramishra
<test-21.example.com> EXEC /bin/sh -c 'echo ~ramishra && sleep 0'
<test-17.example.com> EXEC /bin/sh -c '/usr/bin/python3 /home/ramishra/.ansible/tmp/ansible-tmp-1668673584.0746303-1435377-8803172912834/AnsiballZ_file.py && sleep 0'
<test-18.example.com> PUT /home/ramishra/.ansible/tmp/ansible-local-14328044ytahafl/tmpl2xxribm TO /home/ramishra/.ansible/tmp/ansible-tmp-1668673584.1266525-1435390-274928110267104/AnsiballZ_file.py
...
```

rabi:
Let the system flush it as expected with PR[1].

[1] https://github.com/ansible/ansible/pull/77056

Fixes: #1164

### DIFF
--- a/ansible_runner/display_callback/callback/awx_display.py
+++ b/ansible_runner/display_callback/callback/awx_display.py
@@ -221,7 +221,7 @@ class EventContext(object):
     def get_end_dict(self):
         return {}
 
-    def dump(self, fileobj, data, max_width=78, flush=False):
+    def dump(self, fileobj, data, max_width=78):
         b64data = base64.b64encode(json.dumps(data).encode('utf-8')).decode()
         with self.display_lock:
             # pattern corresponding to OutputEventFilter expectation
@@ -231,8 +231,6 @@ class EventContext(object):
                 escaped_chunk = u'{}\x1b[{}D'.format(chunk, len(chunk))
                 fileobj.write(escaped_chunk)
             fileobj.write(u'\x1b[K')
-            if flush:
-                fileobj.flush()
 
     def dump_begin(self, fileobj):
         begin_dict = self.get_begin_dict()
@@ -240,7 +238,7 @@ class EventContext(object):
         self.dump(fileobj, {'uuid': begin_dict['uuid']})
 
     def dump_end(self, fileobj):
-        self.dump(fileobj, self.get_end_dict(), flush=True)
+        self.dump(fileobj, self.get_end_dict())
 
 
 event_context = EventContext()


cjeanner:
I can confirm this patch solves the issue we're facing. Thank you @rabi !

nitzmahone:
Thanks for tracking this deadlock down and submitting a patch! We want to spend a little time reproducing this one and questioning some of the underlying assumptions that got us into this state before we merge, as we're still not confident  it's not just moving the problem around, especially since threads are involved in the callback dispatch on the core side. The situation in recent versions of core should generally be much better around display deadlocks on stdout, since core is doing it all from the main process now (instead of letting forks also write to stdout), but that also means that it's a lot more "bursty" on the writes, which may be part of what triggered this.

nitzmahone:
I beat on this one a bit this afternoon, and it's as I feared- removing the flush definitely reduces lock contention by allowing the dump to return more quickly, but it's still not too hard to trigger a deadlock since there are two locks and at least two threads involved. I'd say we should merge it anyway since it at least improves the deadlock situation, but I need to spend some time working through the implications of deferring the flush (since sometimes it's a TTY and sometimes it's a file that has realtime-ish constraints). An intermediate possibility might be to defer the flush until after the lock's been released. Anyway, we'll keep playing with it and hopefully get to a robust fix. Thanks!

rabi:
> Anyway, we'll keep playing with it and hopefully get to a robust fix. Thanks!

@nitzmahone Thanks for looking at the PR. I proposed the patch based on comment[1] which suggests to leave it to the system to flush stdout and has actually solved our issue of tripleo moving to ansible 2.14. If there are more issues with multiple re-entrant locks, that could probably be handled separately I guess(?).

[1] https://github.com/ansible/ansible/blob/devel/lib/ansible/utils/display.py#L262-L263

nitzmahone:
Purely from an interactive core perspective, I'd agree there is little appreciable benefit to the explicit flush, especially when it's going to a TTY (hence the comment you mention in display.py), but we need to ensure that runner still always sees what it needs on stdout by the time the corresponding event gets generated. Removing the flush from under the lock changes the guaranteed behavior that's there now, which could screw up a lot of things for the event->stdout correlation unless it gets re-plumbed to sample it another way. I've been looking at that for the longer-term so we can just dump the locks entirely, but it's not something we'd do as a bugfix/backport.

nitzmahone:
fixed by https://github.com/ansible/ansible/pull/79522 - thanks!

PR #1181, opened 2023-01-17
Add hostvars mapping in awx_display callback

### LINKED ISSUE #1137: Object of type HostVarsVars is not JSON serializable
I'm getting error 

```
[WARNING]: Failure using method (v2_runner_on_ok) in callback plugin
(<ansible.plugins.callback.awx_display.CallbackModule object at
0x7f54a8551c70>): Object of type HostVarsVars is not JSON serializable
```

when running the below playbook in AWX (ansible-runner)  

```
- name: test 
  hosts: localhost
  tasks:
    - debug: 
        msg: 
         - "{{ item.inventory_hostname}}"
      loop: "{{ ansible_play_hosts
        | default([])
        | map('extract', hostvars) 
        | list 
        }}"
```

ansible-runner --version
2.1.4.dev2

python --version
Python 3.8.13

Akasurde:
##### SUMMARY

Fixes: #1137

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ansible_runner/display_callback/callback/awx_display.py

### DIFF
--- a/src/ansible_runner/display_callback/callback/awx_display.py
+++ b/src/ansible_runner/display_callback/callback/awx_display.py
@@ -39,6 +39,7 @@ from copy import copy
 # Ansible
 from ansible import __version__ as ansible_version_str
 from ansible import constants as C
+from ansible.module_utils.common._collections_compat import Mapping
 from ansible.plugins.callback import CallbackBase
 from ansible.plugins.loader import callback_loader
 from ansible.utils.display import Display
@@ -107,6 +108,9 @@ class AnsibleJSONEncoderLocal(json.JSONEncoder):
             return {'__ansible_vault': encrypted_form}
         if isinstance(o, (datetime.date, datetime.datetime)):
             return o.isoformat()
+        if isinstance(o, Mapping):
+            # hostvars and other objects
+            return dict(o)
         return super().default(o)
 
 


Akasurde:
cc @sivel

AlanCoding:
We should add some regression test coverage, but :+1: this looks good.

nitzmahone (commented):
This needs to delegate to core's encoder- it's already missing a bunch of object types that core outputs can contain today (and are handled by core's encoder), and once data tagging ships, it'll be hopelessly broken.

AlanCoding:
The encoder class is pretty self-aware that it duplicates the Ansible-core version, in its doc string:

> The class AnsibleJSONEncoder exists in Ansible core for this function this performs a mostly identical function via duck typing

We should really fix this soon, so I want to formalize the suggestion. Is the idea to replace the `AnsibleJSONEncoderLocal` class inside of `ansible_runner.display_callback.callback.awx_display` with instead an import to this?

```python
from ansible.module_utils.common.json import AnsibleJSONEncoder
```

nitzmahone:
Yeah, probably something along those lines, at least for this specific case, but especially once data tagging lands, we'll need to be much more specific about what metadata we actually want to preserve in the callback output.

I'm still uncertain about the value of e.g., preserving the internal representation of an inline vaulted var in callback output. Shouldn't we just mask the value? I don't know what people typically do with the serialized output details like this beyond just looking at it in the UI, and we're only providing special handling for an uncommon subset of vault stuff today... Data tagging will provide us with a whole lot more choices to make about this, but also a lot more flexibility.

PR #1186, opened 2023-01-27
Give more detail when we cannot process a non-JSON streamed line

AlanCoding:
The error handling for this in AWX has been really unforgivably bad. This abuses `result_traceback` somewhat, but this practice already existed in AWX. More information is better than less, is the main idea I'm following here.

### DIFF
--- a/ansible_runner/streaming.py
+++ b/ansible_runner/streaming.py
@@ -244,8 +244,13 @@ class Processor(object):
             try:
                 line = self._input.readline()
                 data = json.loads(line)
-            except (json.decoder.JSONDecodeError, IOError):
-                self.status_callback({'status': 'error', 'job_explanation': 'Failed to JSON parse a line from worker stream.'})
+            except (json.decoder.JSONDecodeError, IOError) as exc:
+                self.status_callback({
+                    'status': 'error',
+                    'job_explanation': (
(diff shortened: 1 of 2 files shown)


fosterseth (approved):

ansible_runner/streaming.py
```diff
@@ -244,8 +244,11 @@ def run(self):
             try:
                 line = self._input.readline()
                 data = json.loads(line)
-            except (json.decoder.JSONDecodeError, IOError):
-                self.status_callback({'status': 'error', 'job_explanation': 'Failed to JSON parse a line from worker stream.'})
+            except (json.decoder.JSONDecodeError, IOError) as exc:
```
Akasurde, line 247: @AlanCoding Thanks for the patch. Could you please add a test for this change?

AlanCoding: Sorry for the delay. I agree.

I just pushed a change where I modified some existing tests. There was effectively already coverage, but I added more detail to the assertions.

Shrews (approved):

Shrews:
Thanks. Backport PRs are requested (`git cherry-pick -x`, please) if you want this in 2.2 or 2.3 releases.

AlanCoding:
Right now I would say "no" to a 2.3 backport. There are a number of error-handling related fixes, and I don't think this would be the most major.

PR #1191, opened 2023-02-09
Emit periodic keepalive events from `Worker`

### LINKED ISSUE #1187: Send "keep alive" (empty) messages to avoid idle timeouts when streaming
The problem we want to address is that we have `ansible-runner worker` running, talking to an `ansible-runner process` agent. Due to the infrastructure set up between them (usually K8S), people can have issues with timeout triggers dropping the connection and doing other bad things.

This proposes that ansible-runner, itself, sends a do-nothing messages from worker-->process at a periodic interval, configured by that user. The default is probably to not send these messages.

We may also provide a do-nothing callback in the process code.

nitzmahone:
fixes #1187

This approach is less "fancy" than some others, but also doesn't require yet another interposing output handle.

Open issues:
- [x] short-circuit handling: Ideally there'd be a short-circuit in the `Process` layer for this, or any other "malformed"/bogus event, but as-is, `Process` will bomb out with a KeyError on any dictionary missing `uuid` or `counter`. Ideally, these keepalives could be as simple as a newline, dot, or `{}`, but unless/until we can be sure that an older `Process` isn't going to barf on them, we have to have a larger payload to keep it happy.

- [x] other consumers: In cases where the event actually gets persisted by `Process`, will anything else be upset by the empty `0-0.json` event it writes as? (_EDIT_: consensus was to go for a "fail-fast" approach instead)

- [x] unit tests: there are a number of potential output corruption races this attempts to prevent, but simulating them with unit tests to ensure they stay "prevented" would probably be a Good Thing. (_EDIT_: added crazy way-subsecond keepalive tests to try and flush out problems and races here)

- [x] tests: Don't want to gold-plate this until  we're sure we like this approach. Also, the transmit/worker/process tests need some refactoring to be able to introduce delays so the keepalives have a chance to fire, and also to ensure that they don't fire when they're not supposed to. 

This specifically is *not* trying to solve the generic "I want to do periodic random stuff in runner" issue- it seems like that would be better handled in a future asyncio layer where such things are trivial to do without extra threading overhead.

### DIFF
--- a/ansible_runner/streaming.py
+++ b/ansible_runner/streaming.py
@@ -8,7 +8,6 @@ import sys
 import tempfile
 import uuid
 import traceback
-from builtins import ValueError
 
 import ansible_runner
 from ansible_runner.exceptions import ConfigurationError
--- a/ansible_runner/streaming.py
+++ b/ansible_runner/streaming.py
@@ -65,7 +65,7 @@ class Transmitter(object):
 
 
 class Worker:
-    def __init__(self, _input=None, _output=None, keepalive_seconds: int | None = None, **kwargs):
+    def __init__(self, _input=None, _output=None, keepalive_seconds: float | None = None, **kwargs):
         if _input is None:
             _input = sys.stdin.buffer
         if _output is None:
@@ -73,7 +73,7 @@ class Worker:
 
         if keepalive_seconds is None:  # if we didn't get an explicit int value, fall back to envvar
             # FIXME: emit/log a warning and silently continue if this value won't parse
-            keepalive_seconds = int(os.environ.get('ANSIBLE_RUNNER_KEEPALIVE_SECONDS', 0))
+            keepalive_seconds = float(os.environ.get('ANSIBLE_RUNNER_KEEPALIVE_SECONDS', 0))
 
         self._keepalive_interval_sec = keepalive_seconds
         self._keepalive_thread: Thread | None = None
--- a/ansible_runner/streaming.py
+++ b/ansible_runner/streaming.py
@@ -1,3 +1,5 @@
+from __future__ import annotations  # allow newer type syntax until 3.10 is our minimum
+
 import codecs
 import json
 import os
@@ -6,10 +8,7 @@ import sys
 import tempfile
 import uuid
 import traceback
-try:
-    from collections.abc import Mapping
-except ImportError:
-    from collections import Mapping
+
 
 import ansible_runner
 from ansible_runner.exceptions import ConfigurationError
@@ -17,6 +16,9 @@ from ansible_runner.loader import ArtifactLoader
 import ansible_runner.plugins
 from ansible_runner.utils import register_for_cleanup
 from ansible_runner.utils.streaming import stream_dir, unstream_dir
+from collections.abc import Mapping
+from functools import wraps
+from threading import Event, RLock, Thread
 
 
 class UUIDEncoder(json.JSONEncoder):
@@ -60,12 +62,18 @@ class Transmitter(object):
         return self.status, self.rc
 
 
-class Worker(object):
-    def __init__(self, _input=None, _output=None, **kwargs):
+class Worker:
+    def __init__(self, _input=None, _output=None, keepalive_seconds: int = 0, **kwargs):
         if _input is None:
             _input = sys.stdin.buffer
         if _output is None:
             _output = sys.stdout.buffer
+
+        self._keepalive_interval_sec = keepalive_seconds
+        self._keepalive_thread: Thread | None = None
+        self._output_event = Event()
+        self._output_lock = RLock()
+
         self._input = _input
         self._output = _output
 
@@ -81,6 +89,64 @@ class Worker(object):
         self.status = "unstarted"
         self.rc = None
 
+    def _begin_keepalive(self):
+        """Starts a keepalive thread at most once"""
+        if not self._keepalive_thread:
+            self._keepalive_thread = Thread(target=self._keepalive_loop)
+            self._keepalive_thread.start()
+
+    def _end_keepalive(self):
+        """Disable the keepalive interval and notify the keepalive thread to shut down"""
+        self._keepalive_interval_sec = 0
+        self._output_event.set()
+
+    def _keepalive_loop(self):
+        """Main loop for keepalive injection thread; exits when keepalive interval is <= 0"""
+        while self._keepalive_interval_sec > 0:
+            # block until output has occurred or keepalive interval elapses
+            if self._output_event.wait(timeout=self._keepalive_interval_sec):
+                # output was sent before keepalive timeout; reset the event and start waiting again
+                self._output_event.clear()
+                continue
+
+            # keepalive interval elapsed; try to send a keepalive...
+            # pre-acquire the output lock without blocking
+            if not self._output_lock.acquire(blocking=False):
+                # something else has the lock; output is imminent, so just skip this keepalive
+                # NB: a long-running operation under an event handler that's holding this lock but not actually moving
+                # output could theoretically block keepalives long enough to cause problems, but it's probably not
+                # worth the added locking hassle to be pedantic about it
+                continue
+
+            try:
+                # were keepalives recently disabled?
+                if self._keepalive_interval_sec <= 0:
+                    # we're probably shutting down; don't risk corrupting output by writing now, just bail out
+                    return
+                # output a keepalive event
+                # FIXME: this could be a lot smaller (even just `{}`) if a short-circuit discard was guaranteed in
+                #  Processor or if other layers were more defensive about missing event keys and/or unknown dictionary
+                #  values...
+                self.event_handler(dict(event='keepalive', counter=0, uuid=0))
+            finally:
+                # always release the output lock (
+                self._output_lock.release()
+
+    def _synchronize_output_reset_keepalive(wrapped_method):
+        """
+        Utility decorator to synchronize event writes and flushes to avoid keepalives splatting in the middle of
+        mid-write events, and reset keepalive interval on write completion.
+        """
+        @wraps(wrapped_method)
+        def wrapper(self, *args, **kwargs):
+            with self._output_lock:
+                ret = wrapped_method(self, *args, **kwargs)
+                # signal the keepalive thread last, so the timeout restarts after the last write, not before the first
+                self._output_event.set()
+                return ret
+
+        return wrapper
+
     def update_paths(self, kwargs):
         if kwargs.get('envvars'):
             if 'ANSIBLE_ROLES_PATH' in kwargs['envvars']:
@@ -93,62 +159,70 @@ class Worker(object):
         return kwargs
 
     def run(self):
-        while True:
-            try:
-                line = self._input.readline()
-                data = json.loads(line)
-            except (json.decoder.JSONDecodeError, IOError):
-                self.status_handler({'status': 'error', 'job_explanation': 'Failed to JSON parse a line from transmit stream.'}, None)
-                self.finished_callback(None)  # send eof line
-                return self.status, self.rc
-
-            if 'kwargs' in data:
-                self.job_kwargs = self.update_paths(data['kwargs'])
-            elif 'zipfile' in data:
+        self._begin_keepalive()
+        try:
+            while True:
                 try:
-                    unstream_dir(self._input, data['zipfile'], self.private_data_dir)
-                except Exception:
-                    self.status_handler({
-                        'status': 'error',
-                        'job_explanation': 'Failed to extract private data directory on worker.',
-                        'result_traceback': traceback.format_exc()
-                    }, None)
+                    line = self._input.readline()
+                    data = json.loads(line)
+                except (json.decoder.JSONDecodeError, IOError):
+                    self.status_handler({'status': 'error', 'job_explanation': 'Failed to JSON parse a line from transmit stream.'}, None)
                     self.finished_callback(None)  # send eof line
                     return self.status, self.rc
-            elif 'eof' in data:
-                break
 
-        self.kwargs.update(self.job_kwargs)
-        self.kwargs['quiet'] = True
-        self.kwargs['suppress_ansible_output'] = True
-        self.kwargs['private_data_dir'] = self.private_data_dir
-        self.kwargs['status_handler'] = self.status_handler
-        self.kwargs['event_handler'] = self.event_handler
-        self.kwargs['artifacts_handler'] = self.artifacts_handler
-        self.kwargs['finished_callback'] = self.finished_callback
+                if 'kwargs' in data:
+                    self.job_kwargs = self.update_paths(data['kwargs'])
+                elif 'zipfile' in data:
+                    try:
+                        unstream_dir(self._input, data['zipfile'], self.private_data_dir)
+                    except Exception:
+                        self.status_handler({
+                            'status': 'error',
+                            'job_explanation': 'Failed to extract private data directory on worker.',
+                            'result_traceback': traceback.format_exc()
+                        }, None)
+                        self.finished_callback(None)  # send eof line
+                        return self.status, self.rc
+                elif 'eof' in data:
+                    break
 
-        r = ansible_runner.interface.run(**self.kwargs)
-        self.status, self.rc = r.status, r.rc
+            self.kwargs.update(self.job_kwargs)
+            self.kwargs['quiet'] = True
+            self.kwargs['suppress_ansible_output'] = True
+            self.kwargs['private_data_dir'] = self.private_data_dir
+            self.kwargs['status_handler'] = self.status_handler
+            self.kwargs['event_handler'] = self.event_handler
+            self.kwargs['artifacts_handler'] = self.artifacts_handler
+            self.kwargs['finished_callback'] = self.finished_callback
 
-        # FIXME: do cleanup on the tempdir
+            r = ansible_runner.interface.run(**self.kwargs)
+            self.status, self.rc = r.status, r.rc
+
+            # FIXME: do cleanup on the tempdir
+        finally:
+            self._end_keepalive()
 
         return self.status, self.rc
 
+    @_synchronize_output_reset_keepalive
     def status_handler(self, status_data, runner_config):
         self.status = status_data['status']
         self._output.write(json.dumps(status_data).encode('utf-8'))
         self._output.write(b'\n')
         self._output.flush()
 
+    @_synchronize_output_reset_keepalive
     def event_handler(self, event_data):
         self._output.write(json.dumps(event_data).encode('utf-8'))
         self._output.write(b'\n')
         self._output.flush()
 
+    @_synchronize_output_reset_keepalive
     def artifacts_handler(self, artifact_dir):
         stream_dir(artifact_dir, self._output)
         self._output.flush()
 
+    @_synchronize_output_reset_keepalive
     def finished_callback(self, runner_obj):
         self._output.write(json.dumps({'eof': True}).encode('utf-8'))
         self._output.write(b'\n')
@@ -210,6 +284,7 @@ class Processor(object):
             self.status_handler(status_data, runner_config=self.config)
 
     def event_callback(self, event_data):
+        # FIXME: this needs to be more defensive to not blow up on "malformed" events or new values it doesn't recognize
         full_filename = os.path.join(self.artifact_dir,
                                      'job_events',
                                      '{}-{}.json'.format(event_data['counter'],
@@ -254,6 +329,9 @@ class Processor(object):
                 self.artifacts_callback(data)
             elif 'eof' in data:
                 break
+            # FIXME: add a short-circuit here to minimize the overhead of keepalives?
+            # elif data.get('event') == 'keepalive':
+            #     continue
             else:
                 self.event_callback(data)
 
--- a/ansible_runner/streaming.py
+++ b/ansible_runner/streaming.py
@@ -337,9 +337,9 @@ class Processor(object):
                 self.artifacts_callback(data)
             elif 'eof' in data:
                 break
-            # FIXME: add a short-circuit here to minimize the overhead of keepalives?
-            # elif data.get('event') == 'keepalive':
-            #     continue
+            elif data.get('event') == 'keepalive':
+                # just ignore keepalives
+                continue
             else:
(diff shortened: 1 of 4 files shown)


ansible_runner/streaming.py
```diff
@@ -81,6 +89,64 @@ def __init__(self, _input=None, _output=None, **kwargs):
...
+        """Disable the keepalive interval and notify the keepalive thread to shut down"""
+        self._keepalive_interval_sec = 0
+        self._output_event.set()
+
+    def _keepalive_loop(self):
+        """Main loop for keepalive injection thread; exits when keepalive interval is <= 0"""
+        while self._keepalive_interval_sec > 0:
+            # block until output has occurred or keepalive interval elapses
+            if self._output_event.wait(timeout=self._keepalive_interval_sec):
+                # output was sent before keepalive timeout; reset the event and start waiting again
```
AlanCoding, line 114: So that I'm understanding this comment, entering this `if` block corresponds to the `self._output_event.set()` action which would get triggered at the end of any other write?

ansible_runner/streaming.py
```diff
@@ -81,6 +89,64 @@ def __init__(self, _input=None, _output=None, **kwargs):
...
+            try:
+                # were keepalives recently disabled?
+                if self._keepalive_interval_sec <= 0:
+                    # we're probably shutting down; don't risk corrupting output by writing now, just bail out
+                    return
+                # output a keepalive event
+                # FIXME: this could be a lot smaller (even just `{}`) if a short-circuit discard was guaranteed in
+                #  Processor or if other layers were more defensive about missing event keys and/or unknown dictionary
+                #  values...
+                self.event_handler(dict(event='keepalive', counter=0, uuid=0))
```
AlanCoding, line 136: I would prefer to lose both of `counter=0, uuid=0`, since neither value makes sense or adds any value. I absolutely want to keep it as a single-line JSON, but the eof event is really just `{"eof": true}`

ansible_runner/streaming.py
```diff
@@ -93,62 +159,70 @@ def update_paths(self, kwargs):
...
         self._output.flush()
 
+    @_synchronize_output_reset_keepalive
     def artifacts_handler(self, artifact_dir):
         stream_dir(artifact_dir, self._output)
         self._output.flush()
 
+    @_synchronize_output_reset_keepalive
     def finished_callback(self, runner_obj):
         self._output.write(json.dumps({'eof': True}).encode('utf-8'))
```
AlanCoding, line 234: you'd probably want to do `self._end_keepalive()` before this line, because eof should be the final line. Wouldn't break anything I know of to have this violated, but better not risk it. If so, this method shouldn't need the decorator either.

ansible_runner/streaming.py
```diff
@@ -81,6 +89,64 @@ def __init__(self, _input=None, _output=None, **kwargs):
...
+        while self._keepalive_interval_sec > 0:
+            # block until output has occurred or keepalive interval elapses
+            if self._output_event.wait(timeout=self._keepalive_interval_sec):
+                # output was sent before keepalive timeout; reset the event and start waiting again
+                self._output_event.clear()
+                continue
+
+            # keepalive interval elapsed; try to send a keepalive...
+            # pre-acquire the output lock without blocking
+            if not self._output_lock.acquire(blocking=False):
```
jjwatt, line 120: Just FYI: This will never be false because `_output_lock` is a `RLock` which is reentrant (not "read" as a surprising number of developers seem to think!) e.g.,
```python
>>> from threading import RLock
>>> lock = RLock()
>>> lock.acquire(blocking=False)
True
>>> lock.acquire(blocking=False)
True
>>> lock.acquire(blocking=False)
True
>>> lock.acquire(blocking=False)
True
>>> lock.release()
>>> lock.release()
>>> lock.release()
>>> lock.release()
>>> lock.release()
  File "<stdin>", line 1, in <module>
RuntimeError: cannot release un-acquired lock
```
You can grab it all you want but it's never released until you call `release()` the same number of times you called `acquire()`. It's meant for preventing deadlock by letting the same thread grab a lock more than once or for multiple threads to hold a lock.

nitzmahone: I think you might be misunderstanding what's going on in that method (remember it's running on a dedicated thread)- I specifically chose an `RLock` to allow it to be "best effort" pre-acquired by the thread, then "inherited" by the call to the event writer from the same thread (which will re-entrantly reacquire the lock). If something else has picked up the lock, it means the main thread is in the middle of writing an event, in which case the keepalive is redundant and should be skipped. Without this lightweight pre-acquisition + reentrant property of RLock, the keepalive thread would block on the event dispatch until the main thread released the lock and immediately splat a redundant keepalive (which in some cases, like during shutdown, could also technically break the output by emitting a keepalive after the `eof` event).

ansible_runner/streaming.py
nitzmahone: I'd love to make it super tiny (even `{}`), but I don't know at what point it's safe to assume that an old broken `Process` impl won't encounter this and barf on it- if the overall keepalive feature doesn't end up being backported, maybe that's a reasonable approach, but if it is, we'll need to have something (at least for a while) that won't break an older processor implementation.

ansible_runner/streaming.py
jjwatt: I think I see what you mean. It's somewhat recursive ("inherited," "same thread"). That makes sense. I don't know enough about the codebase to say any more about that. However, I think my other comment is still valid, right? This conditional on 114 will never be false. I also don't think it will ever block even if `blocking=True`. Thoughts on that?

jjwatt: Sorry, I guess I mean "will always be False" with the `not`. I just mean that `aquire()` will always return `True` with `Rlock`, won't it?

nitzmahone: Nope, not when the main thread already holds the lock, which is the case I'm trying to correct for:

* (on main thread) idle for a long time
* (on keepalive thread) keepalive timeout expires
* (on main thread) acquire lock, start event write
* (on keepalive thread) acquire lock to start a keepalive write, block because the lock is held by main thread
* (on main thread) complete event write, release lock
* (on keepalive thread) unblock and acquire lock, write superfluous keepalive

this fixes that problem, because the extra non-blocking lock acquire on the keepalive thread will return `False` (since the main thread holds the lock), so we know there's another write already in progress and that we don't need a keepalive

nitzmahone: (if you don't believe me, set the keepalive interval to 1s and inject a 5s delay in one of the event writers, which is what I did when I was verifying it :laughing: )

ansible_runner/streaming.py
nitzmahone: Since `_end_keepalive()` doesn't currently `join`, it's probably best to keep this one decorated for the lock (since the keepalive thread could still splat something on its dying gasp). That said, I think it *would* be smart to add a defensive call to `_end_keepalive()` inside this to guarantee that it's definitively impossible to splat a keepalive after the eof message- I'll add that.

ansible_runner/streaming.py
nitzmahone: Correct- when `wait()` returns true, it's because the event was `set()` by something.

ansible_runner/streaming.py
AlanCoding: Since the default is to not send keepalive messages, I'm not worried about new versions of runner with old versions of AWX / controller at all.

Considering the case of an old version of runner with newer versions of AWX... if they don't turn this feature on, it shouldn't be a problem. If they do turn it on, then they will need to have all their EEs upgraded. This should be documented (in the help text for the very setting they are enabling), but I don't see any way around that. Such is the way of introducing any new feature. You can't enable something new while running old code.

ansible_runner/streaming.py
```diff
@@ -60,12 +62,18 @@ def run(self):
...
-    def __init__(self, _input=None, _output=None, **kwargs):
+class Worker:
+    def __init__(self, _input=None, _output=None, keepalive_seconds: int = 0, **kwargs):
         if _input is None:
             _input = sys.stdin.buffer
         if _output is None:
             _output = sys.stdout.buffer
+
+        self._keepalive_interval_sec = keepalive_seconds
+        self._keepalive_thread: Thread | None = None
```
AlanCoding, line 79: I might show my own ignorance with this question - but could you explain the python syntax of this line? I try to reproduce it, and I get something like this

```
----> 1 _keepalive_thread: Thread | None = None

TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

Where there is a highlight around "Thread | None"

ansible_runner/streaming.py
jjwatt: Thanks for the explanation! It's clarifying. I follow what you're saying, but I think I would have to see it running, like you suggested, to really understand the code. I'm going to try to do that, but I wouldn't hold up a PR for it (I'm new and still need to learn some basics). Sounds like you've tested it, and it works the way you expect!

jjwatt: Oh, I see the main idea I was missing: Python's RLock.acquire() does return False if a *different* thread tries to acquire. Cool.

ansible_runner/streaming.py
jjwatt: `from typing import Optional, Union`
https://peps.python.org/pep-0484/#union-types says to use `Optional[T]` as shorthand for `Union[T, None]`, but apparently Python 3.10 added support for `<T> | None`. If we need to stay compatible with 3.9, we should probably stick to `Optional[T]`, so here it should be `self._keepalive_thread: Optional[Thread]`

nitzmahone: It's a Python 3.10+ typing shorthand feature for `Union` (eg, `X | Y | None` is equivalent to `typing.Optional[typing.Union[X, Y]]`), but since I added the `annotations` future, it's allowed and supported back to Python 3.7 from the runtime's perspective, and at least `mypy` and pycharm validate it properly. That future causes the parser to store all type annotations as strings instead of trying to AST them, which allows things like forward references, self-referential type declarations, newer typing syntax in older Pythons, and cheaper runtime overhead.

nitzmahone: > I might show my own ignorance

Please do, anytime, and I promise I'll do the same :wink: !  I'll never think less of anyone for seeking to understand- it drives me bonkers when people are afraid to stop and say "wait, I don't understand that...", because it usually becomes apparent later anyway, and if the thing you don't understand might be foundational for something else, it's a waste of everyone's time to keep moving forward when you can't form the full picture. Thanks for asking!

AlanCoding: Checks out

```python
>>> from __future__ import annotations
>>> _keepalive_thread: Thread | None = None
```

This works, and the result is that `_keepalive_thread` is `None`, but known to be either a `NoneType` or a `Thread`.

ansible_runner/streaming.py
AlanCoding: That answered my question completely. But raised a new question with

> `_end_keepalive()` doesn't currently `join`

Couldn't something go wrong with that? If something was using `Worker` programmatically, will this leave around useless threads? Particularly with multiple calls. While this doesn't seem like a very realistic concern, it's already a concern for ansible-runner tests.

My current reading is that if the parameter `keepalive_seconds` parameter gets its default value of `0`, this will still start a thread that will never time out (but still gets reset, kind of pointlessly, by other reads).

This is still a draft, and I'm not asking you to do the work, but want general oppinions on:
 - If the parameter is 0, could we avoid creating an extra thread entirely, minimally affecting the current code path (which all AWX jobs hit)?
 - For test safety (if nothing else) could we have a means of joining the thread? I might agree that never joining is best for CLI use, but a programmatic hook for testing might be useful.

nitzmahone: > could we avoid creating an extra thread entirely

Yeah, I actually wrote it that way to begin with (as well as no-opping the event and `nullcontext`ing the lock), but since the overhead is relatively low, I chucked that since fewer code path permutations and optimizations is much easier to test (especially where concurrency is involved).

> if the parameter keepalive_seconds parameter gets its default value of 0, this will still start a thread that will never time out 

[this](https://github.com/ansible/ansible-runner/pull/1191/files#diff-81c3df3f520280c8548b6499f287446e289a81ecad30288e0c561d3aa95456c5R105) causes the thread to exit instantly in that case.

I was back-and-forth on the `join()` as well- since I pulse the event in the shutdown request anyway, in *nearly* all cases, the thread should exit instantly, but there is a non-zero possibility that it could wait one keepalive interval. Which does bring up the point that I should mark that thread `daemon=True`. It was originally a futures TPE, but that seemed like more overhead than necessary for the fire-and-forget, esp since we can't just bridge the `cancel` on the future to actually function :( (and I really don't want to introduce asyncio into the runner codebase at this point).

ansible_runner/streaming.py
nitzmahone: @shanemcd and I were talking about how to introduce this as a hotfix to an existing environment via an envvar that could be set in the EE itself, rather than having to plumb it all the way through as a kwarg to `Worker` and (somehow) deal with older EEs that don't understand that kwarg (possibly even just enabling a sane default in *all* our EEs and calling it a day). This is another one of those cases where the payload format design and its consumers didn't consider upgrade-ability, so either we force everything to be upgraded, or we stick with a format that's non-objectionable to existing consumers.

ansible_runner/streaming.py
nitzmahone: @jjwatt if you want to see how I set that up to observe it in a debugger sometime, give me a poke on Slack...

ansible_runner/streaming.py
nitzmahone: * added `ANSIBLE_RUNNER_KEEPALIVE_SECONDS` envvar fallback; explicitly-specified CLI/kwarg value takes precedence
* cleaned up the transmit kwargs so we'll never send `keepalive_seconds` in a payload kwargs bundle (as it's nonsensical there and old runners will barf on it)
* fixed `Process` to short-circuit drop keepalives and drop any malformed events (currently silently; if there's a good place to do so loudly, I'm all ears)

Still need to decide if we want to allow a minimal hotfix deployment for this with only the worker EE- if not, we can truncate the extra crap off the keepalive payload. Another option would be to only do the fatter payload on the 2.3-equivalent version of runner and use the smaller payload for anything new (even maybe `{}`)...

ansible_runner/streaming.py
nitzmahone: * added the defensive `_end_keepalive` call to the finished callback, just in case
* marked the KA thread daemonic; that should prevent it from artificially keeping a process alive if anything goes wrong or a test does something weird
* the keepalive thread state is already inspectable by tests via the worker's `_keepalive_thread` attribute, so doing a `join` in `_end_keepalive` is probably not worth the problems it could cause
* we can cheaply short-circuit the creation of the thread entirely in `_begin_keepalive()` if we want; currently it just immediately exits the thread if KA is disabled

ansible_runner/streaming.py
AlanCoding: If it is a goal to have an EE emit keepalive messages in a way that won't fail the API in some way, then from brief testing with AWX, it seems that the minimal requirements would be that `counter` is greater than 0. Then, that the `event` is a value within the listed event types here:

https://github.com/ansible/awx/blob/0815f935ca359537c023efadf286671e31dc3374/awx/main/models/events.py#L164

And then, to make the UI not mess up, you would ideally want to set start_line and end_line to the same number. What number? I'm not sure.

AlanCoding:
I put up https://github.com/nitzmahone/ansible-runner/pull/1 to suggest a test. You can take it or leave it regarding the patch, but this definitely needs an integration test like that.

ansible_runner/streaming.py
nitzmahone: (consensus was "fail fast", so the keepalive doesn't look like a valid "event" in case it escapes from `Process`, it'll just fail the process altogether)

shanemcd (commented):
I didn't try running it - but the code makes sense. LGTM pending approval from someone that actually tested it.

ansible_runner/streaming.py
```diff
@@ -60,12 +64,22 @@ def run(self):
...
+class Worker:
+    def __init__(self, _input=None, _output=None, keepalive_seconds: float | None = None, **kwargs):
         if _input is None:
             _input = sys.stdin.buffer
         if _output is None:
             _output = sys.stdout.buffer
+
+        if keepalive_seconds is None:  # if we didn't get an explicit int value, fall back to envvar
+            # FIXME: emit/log a warning and silently continue if this value won't parse
+            keepalive_seconds = float(os.environ.get('ANSIBLE_RUNNER_KEEPALIVE_SECONDS', 0))
```
AlanCoding, line 76: This is kind of a painful FIXME. It's not that I want logging (how do you even ship the logs back??), but just a simple `string_to_bool` kind of method. Could we just copy the module here into the ansible-runner utils dir?

https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/parsing/convert_bool.py

nitzmahone: I'm not seeing how the `convert_bool` stuff applies here- we always want a number...

nitzmahone: (I internally changed it to a `float` to make the "way-subsecond acid test" easier, but we always want an actual number)

AlanCoding: aah, sorry. I get it. So a bad value will probably just error, and I think I'm okay with that.

AlanCoding (approved):

hswong3i:
I can install https://github.com/ansible/ansible-runner/releases/tag/2.3.2 with Ubuntu 18.04, which is Python 3.6, but with following runtime error message:
```
$ sudo podman run -ti --rm ubuntu:18.04

root@0772e205b3d1:/# apt update && apt -y install python3-pip

root@0772e205b3d1:/# pip3 install ansible-runner
Collecting ansible-runner
  Downloading https://files.pythonhosted.org/packages/8a/ce/fd842b0d464eafad9c88fce0573e638c806211d579a7cdd5d65b417939b4/ansible_runner-2.3.2-py3-none-any.whl (80kB)
    100% |################################| 81kB 10.6MB/s 
Collecting packaging (from ansible-runner)
  Downloading https://files.pythonhosted.org/packages/05/8e/8de486cbd03baba4deef4142bd643a3e7bbe954a784dc1bb17142572d127/packaging-21.3-py3-none-any.whl (40kB)
    100% |################################| 40kB 17.3MB/s 
Collecting pexpect>=4.5 (from ansible-runner)
  Downloading https://files.pythonhosted.org/packages/39/7b/88dbb785881c28a102619d46423cb853b46dbccc70d3ac362d99773a78ce/pexpect-4.8.0-py2.py3-none-any.whl (59kB)
    100% |################################| 61kB 17.7MB/s 
Collecting python-daemon (from ansible-runner)
  Downloading https://files.pythonhosted.org/packages/89/38/c223036ee8104ae95118d4481b5cacccf4547d7e5b8d1ee1c63d2cd57e5d/python_daemon-2.3.2-py3-none-any.whl
Requirement already satisfied: six in /usr/lib/python3/dist-packages (from ansible-runner)
Collecting pyyaml (from ansible-runner)
  Downloading https://files.pythonhosted.org/packages/36/2b/61d51a2c4f25ef062ae3f74576b01638bebad5e045f747ff12643df63844/PyYAML-6.0.tar.gz (124kB)
    100% |################################| 133kB 17.7MB/s 
Collecting pyparsing!=3.0.5,>=2.0.2 (from packaging->ansible-runner)
  Downloading https://files.pythonhosted.org/packages/6c/10/a7d0fa5baea8fe7b50f448ab742f26f52b80bfca85ac2be9d35cdd9a3246/pyparsing-3.0.9-py3-none-any.whl (98kB)
    100% |################################| 102kB 18.4MB/s 
Collecting ptyprocess>=0.5 (from pexpect>=4.5->ansible-runner)
  Downloading https://files.pythonhosted.org/packages/22/a6/858897256d0deac81a172289110f31629fc4cee19b6f01283303e18c8db3/ptyprocess-0.7.0-py2.py3-none-any.whl
Requirement already satisfied: setuptools in /usr/lib/python3/dist-packages (from python-daemon->ansible-runner)
Collecting lockfile>=0.10 (from python-daemon->ansible-runner)
  Downloading https://files.pythonhosted.org/packages/c8/22/9460e311f340cb62d26a38c419b1381b8593b0bb6b5d1f056938b086d362/lockfile-0.12.2-py2.py3-none-any.whl
Collecting docutils (from python-daemon->ansible-runner)
  Downloading https://files.pythonhosted.org/packages/8d/14/69b4bad34e3f250afe29a854da03acb6747711f3df06c359fa053fae4e76/docutils-0.18.1-py2.py3-none-any.whl (570kB)
    100% |################################| 573kB 3.9MB/s 
Building wheels for collected packages: pyyaml
  Running setup.py bdist_wheel for pyyaml ... done
  Stored in directory: /root/.cache/pip/wheels/95/84/67/ebeac632c63797cfbeb90128ca41073117721540dad526d213
Successfully built pyyaml
Installing collected packages: pyparsing, packaging, ptyprocess, pexpect, lockfile, docutils, python-daemon, pyyaml, ansible-runner
Successfully installed ansible-runner-2.3.2 docutils-0.18.1 lockfile-0.12.2 packaging-21.3 pexpect-4.8.0 ptyprocess-0.7.0 pyparsing-3.0.9 python-daemon-2.3.2 pyyaml-6.0

root@0772e205b3d1:/# ansible-runner --version
  File "/usr/local/bin/ansible-runner", line 7, in <module>
    from ansible_runner.__main__ import main
  File "/usr/local/lib/python3.6/dist-packages/ansible_runner/__init__.py", line 3, in <module>
    from .interface import run, run_async, \
  File "/usr/local/lib/python3.6/dist-packages/ansible_runner/interface.py", line 33, in <module>
    from ansible_runner.streaming import Transmitter, Worker, Processor
  File "/usr/local/lib/python3.6/dist-packages/ansible_runner/streaming.py", line 1
    from __future__ import annotations  # allow newer type syntax until 3.10 is our minimum
    ^
SyntaxError: future feature annotations is not defined
```

Seems like changes from https://github.com/ansible/ansible-runner/pull/1191/files#diff-81c3df3f520280c8548b6499f287446e289a81ecad30288e0c561d3aa95456c5R1 require Python >= 3.7, am I correct?

PR #1192, opened 2023-02-10
Bump cryptography from 35.0.0 to 39.0.1 in /docs

Akasurde:
##### SUMMARY

Bumps cryptography from 35.0.0 to 39.0.1

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
docs/requirements.txt

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -4,7 +4,7 @@ babel==2.9.1
 certifi==2022.12.7
 cffi==1.15.0
 charset-normalizer==2.0.7
-cryptography==35.0.0
+cryptography==39.0.1
 docutils==0.17.1
 idna==3.3
 imagesize==1.3.0


Shrews (changes requested):
These are requirements for _documentation_. Why do we even need this lib? I tested just removing it from this file and docs build just fine. Let's just delete it. Probably wouldn't hurt to go through the rest of the listed requirements and see if we can get it down to only what is actually needed.

Shrews:
So it seems these pins are present to keep the docs job from failing (see PR #888). I suspect the cryptography requirement comes in as something that is required by the `ansible-core` dependency (I don't know how Sam extracted all of these non-doc related deps). Since the docs job does not actually run ansible or even ansible-playbook, I see no reason ever update any of these pins that are not relevant to actually creating docs.

All that to say, I vote we ignore anything from dependencybot for the `docs/requirements.txt` file.

Akasurde:
Closing as per https://github.com/ansible/ansible-runner/pull/1192#issuecomment-1428397814

PR #1198, opened 2023-02-26
Add podman-remote as container engine

rcarrillocruz:
This allows to have podman-remote as container engine option. 
This is to avoid having the images to hardcode linking podman to podman-remote.

### DIFF
--- a/ansible_runner/config/_base.py
+++ b/ansible_runner/config/_base.py
@@ -137,7 +137,7 @@ class BaseConfig(object):
 
         os.makedirs(self.artifact_dir, exist_ok=True, mode=0o700)
 
-    _CONTAINER_ENGINES = ('docker', 'podman')
+    _CONTAINER_ENGINES = ('docker', 'podman', 'podman-remote')
 
     @property
     def containerized(self):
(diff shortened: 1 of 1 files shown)


rcarrillocruz:
@shanemcd let me know what you think, per the conversation we had in chat.

sivel:
I believe this will also require a different way of handling the mounted volumes.  In ansible-test we do this by not using volumes, and instead by copying in and out the things we need.  Otherwise, you are mounting volumes from the remote podman host, and not from where the client is executed.

ansible_runner/config/_base.py
```diff
@@ -137,7 +137,7 @@ def __init__(self,
 
         os.makedirs(self.artifact_dir, exist_ok=True, mode=0o700)
 
-    _CONTAINER_ENGINES = ('docker', 'podman')
+    _CONTAINER_ENGINES = ('docker', 'podman', 'podman-remote')
```
sivel, line 140: Also note on some systems, there is no `podman-remote` CLI, but instead it is `podman --remote`

shanemcd:
@sivel For our use case, the remote host is not really remote - we're talking to the host via a podman socket mounted in a container, so we are able to mount in the directories we need. I'm trying to avoid installing the full podman package into our EE because podman-remote is smaller.

So while I think we can at some point rethink how we mount volumes, I don't think it's _required_ here.

I do think this is worth documenting though. I went and looked to see where we might mention this and I just noticed that these never got updated when we switched the default from bwrap to podman: https://ansible-runner.readthedocs.io/en/stable/intro/?highlight=process_isolation_executable#process-isolation-settings-for-runner

PR #1200, opened 2023-02-27
Differentiate timeout vs idle timeout

TheRealHaoLiu:
currently idle timeout and timeout result in the same status call back this result in the AWX not being able to parse the differences between the two failure condition and make it confusing to debug

### DIFF
--- a/ansible_runner/runner.py
+++ b/ansible_runner/runner.py
@@ -35,6 +35,7 @@ class Runner(object):
         self.status_handler = status_handler
         self.canceled = False
         self.timed_out = False
+        self.idle_timed_out = False
         self.errored = False
         self.status = "unstarted"
         self.rc = None
@@ -337,7 +338,7 @@ class Runner(object):
                 if self.config.idle_timeout and (time.time() - self.last_stdout_update) > self.config.idle_timeout:
                     self.kill_container()
                     Runner.handle_termination(child.pid, is_cancel=False)
-                    self.timed_out = True
+                    self.idle_timed_out = True
 
             stdout_handle.close()
             stderr_handle.close()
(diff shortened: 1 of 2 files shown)


ansible_runner/runner.py
```diff
@@ -350,6 +350,8 @@ def _decode(x):
             self.status_callback('successful')
         elif self.timed_out:
             self.status_callback('timeout')
+        elif self.timed_out:
```
Akasurde: Shouldn't be this `idle_timeout` instead of `timed_out`?

ansible_runner/runner.py
```diff
@@ -337,7 +337,7 @@ def _decode(x):
                 if self.config.idle_timeout and (time.time() - self.last_stdout_update) > self.config.idle_timeout:
                     self.kill_container()
                     Runner.handle_termination(child.pid, is_cancel=False)
-                    self.timed_out = True
+                    self.idle_timed_out = True
```
relrod, line 341: This will need a default like `timed_out` has on line 37. Otherwise if we _don't_ hit this case, we'll crash when we check the variable below.

Shrews (changes requested):
You are missing checks in several places where currently `self.timed_out` is checked for cleanup, return codes, etc.

Also, we might need to consider if this is a breaking change because `status` value is now different, which is sort of an API change. Not sure this can be backported.

AlanCoding:
Really quick 2 cents here - I also think the API change of a new `status` would be disruptive and prone to break things. Not something we want in the error handling! If you look a layer deeper in the call stack, other options present themselves.

https://github.com/ansible/ansible-runner/blob/d392c5150e85b2db842a7c4d206d03d3f7393d53/ansible_runner/runner.py#L101-L109

Instead of using a new `status` value for the call to `status_callback`, it would go much more smoothly to add more data to `status_data` in the `status_handler`. I think you could even safely abuse `job_explanation`, because this failure path is a known scenario and it shouldn't be conflicting with anything else setting that string. This could relatively nice present the timeout _reason_ to the user.

PR #1206, opened 2023-03-02
[devel] Add `python_requires = >= 3.8`

### LINKED ISSUE #1205: v2.3.2 is not Compatible with Python <= 3.7
See https://github.com/ansible/ansible-runner/pull/1191#issuecomment-1451605346

hswong3i:
Fixes https://github.com/ansible/ansible-runner/issues/1205

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -10,6 +10,9 @@ description_file = README.md
 description_content_type = text/markdown
 license_file = LICENSE.md
 
+[options]
+python_requires = >=3.8
+
 [entry_points]
 console_scripts =
     ansible-runner = ansible_runner.__main__:main


Shrews:
Thanks for the contribution.

hswong3i:
Could we have the corresponding new release(s) which also yarn the legacy version from pypi.org?

Else CentOS 7/8, openSUSE Leap 15.4, Debian 10 and Ubuntu 18.04 will still receive incorrect 2.y.z version though pip install...

nitzmahone:
Unfortunately just re-releasing the other versions doesn't help for the unpinned `pip install ansible-runner` case without yanking every incorrect release (which we can't reasonably do at this point). You'd still have to explicitly constrain to an older x.y version anyway to avoid getting the (incorrect) newer x.y+1 that will appear to be viable for Python 3.6, and if you're going to use a `<` constraint anyway, there's no additional benefit to updating and re-releasing old packages.

nitzmahone:
Oh actually, I forgot they added support for yanking-without-deletion awhile back... So yeah, actually, maybe we can do that.

nitzmahone:
... the one caveat to this is that you have to have a new-enough `pip` that understands yanked releases. I can't remember when they added that, but almost certainly the OS-packaged version of `pip` included in any OS that shipped with Python 3.6 will *not* understand yanked releases and happily install the wrong one. But I'm happy to try and clean it up within reason for folks that have a decent pip anyway, since it was definitely "our bad" on that one.

hswong3i:
Yes please do so, my OBS repo have latest possible pip for legacy OS, see https://build.opensuse.org/package/show/home:alvistack/pypa-pip-21.3.1

I had also package ansible-runner 1.4.9 for those Python 3.6 legacy OS, and manually patching 2.3.2 with python_requires = >= 3.8, see https://build.opensuse.org/package/show/home:alvistack/ansible-ansible-runner-2.3.2

nitzmahone:
I didn't realize how long this had been broken- technically since the origin of the project, but materially since 2.0 bumped the implied minimum Python from 2.7 to 3.8 without adding the packaging metadata. By my count, that's 15 releases as of right now (not including alpha/beta) that would need to be yanked and either 3 or 15 releases that need to be re-packaged with the additional metadata to fix EOL or near-EOL Pythons (that also have a newer pip) finding the right `ansible-runner` automatically.

I'm committed to making sure this works going forward (including yanking releases in case we forget again), but given that we went 18 months this way without anyone apparently noticing, and the potential (wasted) churn that a bunch of new tags and releases on dead 2.x branches and PyPI could cause, I'm thinking we just have to say "sorry, our bad, we'll do better next time", and that runner 1.x users on dead/dying Pythons will need to do `pip install 'ansible-runner<2.0'` on their installations.

hswong3i:
Python 3.6 is EOL for sure, but CentOS 7/8 are still support release... And people with such OS have no choice to install with pip and now will fail if we don't yark those legacy 2.y.x release.

Technically we could do something better than just a sorry message from document: yark those legacy 2.y.z release, publish new 2.0/2.1/2.2/2.3 release (i.e. 4 releases) to pipy.org, and problem will get fixed for those with new enough pip user.

PR #1213, opened 2023-03-02
Add license and classifiers metadata

### LINKED ISSUE #1196: Add additional license metadata to setup.cfg
In addition to `license_file`, we should also include the license in `license` and `classifiers` to better disclose the license within the `Meta` section of pypi.org.  As of now, there is some indication in the readme, but no machine parsable information is available via the pypi API.

Shrews:
Fixes #1196

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,7 +9,7 @@ project_urls =
 description_file = README.md
 description_content_type = text/markdown
 license_file = LICENSE.md
-license = Apache
+license = Apache Software License, Version 2.0
 classifiers =
     Development Status :: 5 - Production/Stable
     Environment :: Console
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,6 +9,24 @@ project_urls =
 description_file = README.md
 description_content_type = text/markdown
 license_file = LICENSE.md
+license = Apache
+classifiers =
+    Development Status :: 5 - Production/Stable
+    Environment :: Console
+    Intended Audience :: Developers
+    Intended Audience :: Information Technology
+    Intended Audience :: System Administrators
+    License :: OSI Approved :: Apache Software License
+    Natural Language :: English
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -9,6 +9,24 @@ project_urls =
 description_file = README.md
 description_content_type = text/markdown
 license_file = LICENSE.md
+license = Apache
```
sivel: There is no standard here, and the way the pypi UI shows this is basically `license (classifier-license)`.

But after looking through the list of packages with the apache classifier, I've come to believe `Apache Software License, Version 2.0` covers the most.

```suggestion
license = Apache Software License, Version 2.0
```

sivel (approved):

hswong3i:
@Shrews shall we also backport this for 2.0 and 2.1? Therefore we could release a new version to pypi with yank the incompatible legacy versions?

Shrews:
@hswong3i The 2.1 and earlier releases will receive only security fixes. Those branches will not receive this change.

hswong3i:
Yes as a human being we all respecting the community policy for just patching for security for 2.0/2.1 release, but as a platform and tool as pypi.org and pip couldn't understand this.

If we only hotfix 2.2/2.3 release, pypi.org and pip will still report 2.1.x as suitable and available version for installation, which not sync with what we claimed in document for human...

Shall we give an exception for solving a technical issue by technical but not by documentation only?

hswong3i:
P.S. as a rpm/deb packager with OBS, I had already EXCLUDE all 2.x build for legacy OS with Python 3.6 but just provide 1.4.9. So basically I am respecting community policy and already solve the problem for my own daily use case under my own controlled environment.

But this didn't solve the unexpected behavior if people follow our document and install with pip directly. We could patch the document with sorry, but patch the code could solve the root cause...

Issue #1218, opened 2023-03-09
project_dir does not work


Shrews:
Setting `project_dir` in the `interface.run()` method does not appear to work. Example:

```
r = ansible_runner.run(
        private_data_dir='demo',
        playbook='test.yml',
        project_dir='my_project',
)
```

This should find the `demo/my_project/test.yml` playbook, but instead fails with:

`ERROR! the playbook: test.yml could not be found`

jkugler:
Also, it appears to not modify the directory it passes to podman as the working directory. Even with "project_dir" specified, podman is still invoked with:
```
execve("/usr/bin/podman", ["/usr/bin/podman", "run", "--rm", "--tty", "--interactive", "--workdir", "/runner/project", "-v", ...
```
and if the 'project' directory does not exist in `private_data_dir`,  will throw an error of `Error: workdir "/runner/project" does not exist on container ...`

Akasurde:
resolved_by_pr https://github.com/ansible/ansible-runner/pull/1222

Issue #1143, opened 2022-10-12
Automatic generated folders make it impossible to run different playbooks in multithreaded environment.


joaopfg:
The automatic generated folders (`artifacts`, `inventory` and `project`) make it impossible to run different playbooks (with ansible-runner) in different threads.

I can of course remove these folders before and after every command but it doesn't solve the problem when two commands are run almost at the same time.

Is there some work around for this ?

Also, I don't see the reason for these folders to be dumped into the filesystem. Couldn't they be stored as python objects and carried through the code ? Is there a deeper reason for them to be dumped into the filesystem ?

urbanchef:
Any updates on this?

joaopfg:
@urbanchef The work around I implemented in my code is to generate these folders with names in a one to one correspondence with the thread that generated them. Like this, I can manage their lifecycle in each thread. Ugly solution though. I hope someone gets inspired to make a PR getting rid of these folders. I don't have time to dig deeper into that now.

urbanchef:
FYI, I create a separate private_data_dir for each playbook run given that project_dir remains the same. I delete private_data_dir some time after. All the details I mentioned [here](https://github.com/ansible/ansible-runner/pull/1222#issuecomment-1833301463)

PR #1222, opened 2023-03-17
Handle custom project_dir

Akasurde:
##### SUMMARY

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ansible_runner/runner.py
test/fixtures/projects/project_dir_test/env/envvars
test/fixtures/projects/project_dir_test/inventory/inv_1
test/fixtures/projects/project_dir_test/my_project/debug.yml
test/integration/test_interface.py


Akasurde:
Closing and re-opening to retrigger the CI.

jkugler:
Could we also add a test to verify that it works when

1. there is no "project" directory and 
2. you specify a custom project_dir (see comment on the original bug for details: https://github.com/ansible/ansible-runner/issues/1218#issuecomment-1462576456)

Akasurde:
> Could we also add a test to verify that it works when
> 
>     1. there is no "project" directory and
> 
>     2. you specify a custom project_dir (see comment on the original bug for details: [project_dir does not work #1218 (comment)](https://github.com/ansible/ansible-runner/issues/1218#issuecomment-1462576456))

Yes sure. Thanks for the review.

test/integration/test_interface.py
```diff
@@ -190,6 +190,68 @@ def test_multiple_inventories(project_fixtures):
...
+    private_data_dir = project_fixtures / 'project_dir_test'
+
+    res = run(
+        private_data_dir=private_data_dir,
+        playbook='debug.yml',
+    )
+    stdout = res.stdout.read()
+
+    assert res.rc == 0, stdout
+    assert 'project' in stdout
```
Shrews: This will also match `my_project`. Best to be more explicit and use `In project` instead.

test/integration/test_interface.py
```diff
@@ -190,6 +190,68 @@ def test_multiple_inventories(project_fixtures):
...
+        private_data_dir=private_data_dir,
+        playbook='debug.yml',
+        process_isolation_executable=runtime,
+        process_isolation=True,
+        container_image=defaults.default_container_image,
+    )
+    stdout = res.stdout.read()
+
+    assert res.rc == 0, stdout
+    assert 'project' in stdout
```
Shrews: Again: s/project/In project/

ansible_runner/runner.py
```diff
@@ -193,6 +193,8 @@ def run(self):
                 )
         else:
             cwd = self.config.cwd
+            if self.config.project_dir:
+                cwd = os.path.join(self.config.cwd, self.config.project_dir)
```
Shrews: I don't believe this is going to be the fix. `project_dir`, apparently, _can_ be a full path that exists outside of the private_data_dir (see `test/unit/config/test_runner.py::test_runner_config_project_dir()` test for an example).

Shrews (changes requested):

docs/python_interface.rst
```diff
@@ -262,6 +262,22 @@ Usage examples
...
 
+.. code-block:: python
+
+  import ansible_runner
+  # Specifying custom container working directory while using Podman
+  # `/tmp/demo` contains `my_project` directory which contains all the Ansible content
+  res = ansible_runner.run(
+    private_data_dir='/tmp/demo',
+    host_pattern='localhost',
+    container_workdir='/runner/my_project',
```
jkugler: Is this correct? This Issue/PR is about getting it to properly respect `project_dir`.  Is we want to add this example, maybe name it something besides `my_project` to not create confusion with a custom `project_dir` setting?

urbanchef:
If I understand it correctly, this change enforces project_dir to be under private_data_dir even if project_dir is set and is absolute path? 

I believe **project_dir should not be forced to be under private_data_dir**, because then there is no way to work around issue #1143. Consider a case: I have a runner thread that monitors a queue periodically for playbooks to run. When there's a job submitted the runner thread creates a separate unique private_data_dir for each run as well as runs the playbook. This way project_dir remains unchanged, because main ansible playbook code base resides there. The code base consists of dozens of ansible roles and playbooks. Otherwise, I would need to copy all that code into a separate private_data_dir for each playbook run which is clearly not the way I want to go.

One more thing I would like to stress upon is I don't need to reuse anything dumped to filesystem. Be it envs, cmdline, passwords, etc... I handle this programmatically. That is why I also tried to set suppress_env_files=True in settings so that env is not dumped to filesystem. But it didn't work. 

Here's the part of runner thread code I use to run playbooks. And it works! I use ansible-runner==2.3.1, ansible==4.10.0 (ansible-core==2.11.12)
```
settings = {'suppress_ansible_output': True,
                    'suppress_output_file': True,
                    'suppress_env_files': True}

unique_private_data_dir_for_current_run = os.path.join('/opt/ansible/private_data_dirs', f"{unique_identifier}")
ansible_runner_kwargs = {"project_dir": '/opt/ansible/project',
                                 "private_data_dir": unique_private_data_dir_for_current_run,
                                "playbook": playbook,
                                "inventory": json.dumps(inventory),
                                "extravars": run_data,
                                "verbosity": verbosity,
                                "envvars": {'ANSIBLE_STDOUT_CALLBACK': 'mycustomstdoutcallback'},
                                "settings": settings,
                                "cmdline": cmdline,
                                "passwords": passwords
                                 }
result = ansible_runner.run(**ansible_runner_kwargs)

# remove because not needed and some files (passwords) under unique_private_data_dir_for_current_run/env 
# directory may contain sensitive information. For example, ansible vault password. 
shutil.rmtree(unique_private_data_dir_for_current_run)
```

pschiffe:
Hello, just hit this issue today where I couldn't change `--workdir /runner/project` arg of podman run within `ansible-runner run` command

PR #1224, opened 2023-03-21
Remove pkg_resources use

### LINKED ISSUE #1223: `pkg_resources` is deprecated as an API
https://github.com/pypa/setuptools/blob/main/pkg_resources/__init__.py#L121

Found it was throwing some warnings in navigator's tests.

Akasurde:
##### SUMMARY

* pkg_resources is deprecated in favor of importlib.metadata

Fixes: #1223

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ansible_runner/__init__.py
ansible_runner/__main__.py
ansible_runner/utils/importlib_compat.py

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ packaging
 python-daemon
 pyyaml
 six
+importlib-metadata; python_version < "3.10.0"
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,4 @@ packaging
 python-daemon
 pyyaml
 six
-importlib-metadata; python_version < "3.10.0"
+importlib-metadata >= 4.6, < 4.7; python_version < '3.10'
--- a/ansible_runner/__init__.py
+++ b/ansible_runner/__init__.py
@@ -1,5 +1,4 @@
-import pkg_resources
-
+from .utils.importlib_compat import importlib_metadata
 from .interface import run, run_async, \
                         run_command, run_command_async, \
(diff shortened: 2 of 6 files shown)


Akasurde:
cc @cidrblock

requirements.txt
```diff
@@ -3,3 +3,4 @@ packaging
 python-daemon
 pyyaml
 six
+importlib-metadata; python_version < "3.10.0"
```
nitzmahone: Since the standalone `importlib-metadata` project maintains several different branches that roughly track what's in a given Python stdlib, we should probably pin to a specific branch just for consistent behavior (and also to ensure that it has the features we want). Core does the same thing for the standalone `importlib-resources`- [looking at their compat docs](https://github.com/python/importlib_metadata#compatibility), this should be what we want:

```suggestion
importlib-metadata >= 4.6, < 4.7; python_version < '3.10'
```

nitzmahone (changes requested):
LGTM once we have at least a minimum version; also, I'm assuming we'll be asked to backport this, which has downstream packaging implications that p12n will need to be aware of. If that ends up being a problem, another option might be a utility method that would just allow the 3.8/3.9-embedded version of the lib to be used natively and have it manually filter the groups if we're < Python 3.10 (and delegate the filtering on newer Pythons)- basically polyfill the missing functionality that was added in 3.10 to avoid needing the extra dep on the standalone lib.

nitzmahone (approved):

PR #1234, opened 2023-04-17
Drop pbr in favor of setuptools_scm

sivel:
Drop pbr in favor of setuptools_scm


sivel:
Example version:

```
$ ansible-runner --version
2.3.1.dev20+gdffe21c
```

Shrews:
FYI, there are more references to pbr that will need to be removed:

 * docs/requirements.txt
 * docs/requirements.in
 * docs/conf.py

sivel:
I think we're also going to want to drop requirements.txt when we do this, which I believe will require us to bump to tox 4 which is happening elsewhere

setup.py
```diff
@@ -2,7 +2,4 @@
 
 from setuptools import setup
 
-setup(
-    setup_requires=['pbr'],
-    pbr=True,
-)
+setup()
```
sivel: Technically, this file could probably be deleted, but is just left minimal for backwards compat

nitzmahone:
Might want to crib from some of the things I figured out on https://github.com/ansible/ansible-builder/pull/494 - specifically the archive metadata hackery and making sure CI can actually poop out a proper version...

Shrews (approved):

.github/workflows/ci.yml
```diff
@@ -26,6 +26,8 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v3
+        with:
```
nitzmahone, line 29: not even sure this is necessary for runner- these tests succeeded without it for me

nitzmahone (approved):
LGTM, and worked for the various simulated release activities I tried on this branch

.github/workflows/ci.yml
sivel: It's not, but it does produce a better version output in some cases when tox installs the wheel, instead of just `0.0.1`.

PR #1238, opened 2023-04-24
Ensure that UTC ISO8601 datetimes include timezone info

sivel:
This PR modifies UTC datetime generation to always include a `datetime.timezone.utc` `tzinfo`, which ensures that datetimes emitted by runner are correctly indicated to be UTC.

ISO8601 states that missing timezone info indicates localtime, and not UTC.

This should not impact AWX, as the code already defensively corrects this issue:

https://github.com/ansible/awx/blob/2ce9440bab6f14c0eb2f1b85206f1f92dc88fe0d/awx/main/models/events.py#L416-L427


sivel:
cc @shanemcd @AlanCoding

Issue #466, opened 2020-06-23
Remove Dependency on psutil


matburt:
We depend on psutil for one reason, to handle ansible termination:

https://github.com/ansible/ansible-runner/blob/2505afdcc152fab54d416144b9d0a05d083e8256/ansible_runner/runner.py#L394

This is a pretty heavy dependency and I'd like to remove it either by investigating a more reliable way to signal Ansible to exit or finding a lighter weight mechanism that doesn't have a direct C library dependency. Whatever replacement strategy is found needs to support the same sort of behavior we have today which is aggressive.

ryanpetrello:
<img src="https://media1.tenor.com/images/a4d18ffe91a44757e8da9dd73330f4a8/tenor.gif?itemid=7633730" />

AlanCoding:
This is also used in the callback plugin.

https://github.com/ansible/ansible-runner/blob/983f2fa66f612757975e48467d05b90b11e09851/ansible_runner/display_callback/cleanup.py#L35

I'm not entirely clear on the reasons why.

AlanCoding:
Also should document:

https://github.com/ansible/ansible-runner/blob/49d4c9b68f62a006ff43f58c97ab455c15356b43/Dockerfile#L8

It seems that psutil is the reason that the _system_ package `python3-devel` is needed. So if someone removes psutil, remove this too.

AlanCoding:
Also link https://github.com/ansible/ansible/pull/70667

wenottingham:
See: https://github.com/ansible/ansible-runner/pull/555

PR #1256, opened 2023-06-07
Improve 'worker --worker-info' portability

Shrews:
The `--worker-info` option to the `worker` command depended on `/proc/meminfo` existing. We can use the `psutil` package (a new requirement) to make this more portable.

This also improves code coverage by removing unnecessary, custom capacity querying methods, and adding a test for the `--worker-info` option.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,6 +31,7 @@ python_requires = >=3.9
 install_requires =
     pexpect>=4.5
     packaging
+    psutil
     python-daemon
     pyyaml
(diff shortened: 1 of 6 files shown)


Shrews:
May need to reconsider this given the history in https://github.com/ansible/ansible-runner/issues/466

Shrews:
Since no one is asking for this "portability", I'm going to close this and NOT pull `psutil` back in as a dependency.

PR #1259, opened 2023-06-20
Add some type hinting

Shrews:
We now have a `mypy` linting job thanks to PR #1254, so let's improve code health by adding some type hints.

### DIFF
--- a/src/ansible_runner/cleanup.py
+++ b/src/ansible_runner/cleanup.py
@@ -1,3 +1,6 @@
+from __future__ import annotations
+
+import argparse
 import datetime
 import glob
 import os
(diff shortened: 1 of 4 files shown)


src/ansible_runner/cleanup.py
```diff
@@ -143,7 +151,7 @@ def cleanup_dirs(pattern, exclude_strings=(), grace_period=GRACE_PERIOD_DEFAULT)
     return ct
 
 
-def cleanup_images(images, runtime='podman'):
+def cleanup_images(images: list, runtime: str) -> int:
```
Akasurde, line 154:
```suggestion
def cleanup_images(images: list, runtime: str='podman') -> int:
```
Missing default value?

Akasurde (approved):
Rest LGTM

src/ansible_runner/cleanup.py
Shrews: Nope. I changed the calls to `cleanup_images()` to make sure this value is never `None` by using the default runtime at that point.

Akasurde: Ok. Got it.

PR #1267, opened 2023-07-11
Enable pylint check C0209

Shrews:
Enables pylint check C0209 (consider-using-f-string) and changes all non-f-string formats.

Also, change yamllint config to ignore files specified in `.gitignore` so local runs of the `linters` tox environment are more likely to succeed.

### DIFF
--- a/src/ansible_runner/config/runner.py
+++ b/src/ansible_runner/config/runner.py
@@ -134,8 +134,7 @@ class RunnerConfig(BaseConfig):
         if self.sandboxed and self.directory_isolation_path is not None:
             self.directory_isolation_path = tempfile.mkdtemp(prefix='runner_di_', dir=self.directory_isolation_path)
             if os.path.exists(self.project_dir):
-                output.debug("Copying directory tree from {} to {} for working directory isolation".format(self.project_dir,
-                                                                                                           self.directory_isolation_path))
+                output.debug(f"Copying directory tree from {self.project_dir} to {self.directory_isolation_path} for working directory isolation")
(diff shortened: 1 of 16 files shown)


src/ansible_runner/config/runner.py
```diff
@@ -270,35 +269,35 @@ def generate_ansible_command(self):
...
                 extra_vars_list = []
                 for k in self.extra_vars:
-                    extra_vars_list.append("\"{}\":{}".format(k, json.dumps(self.extra_vars[k])))
+                    extra_vars_list.append(f"\"{k}\":{json.dumps(self.extra_vars[k])}")
 
                 exec_list.extend(
                     [
                         '-e',
-                        '{%s}' % ','.join(extra_vars_list)
+                        f'{{{",".join(extra_vars_list)}}}'
```
Akasurde, line 283:
```suggestion
                        ",".join(extra_vars_list)
```

Akasurde (approved):

src/ansible_runner/config/runner.py
Shrews: Your suggestion isn't the same, though. The intent is to pass the extra vars as inline JSON so the surrounding parens are required here.

Shrews: The triple `{` and `}` do look weird, but the first two are how you send a literal paren character in a f-string. The third is for the interpolated value.

Akasurde: TIL

PR #1285, opened 2023-07-27
Updating UTC handing to use UTC aware objects

radez:
Python 3.12 will throw DeprecationWarning: datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.now(datetime.UTC)

### DIFF
--- a/src/ansible_runner/display_callback/callback/awx_display.py
+++ b/src/ansible_runner/display_callback/callback/awx_display.py
@@ -70,7 +70,7 @@ CENSORED = "the output has been hidden due to the fact that 'no_log: true' was s
 
 
 def current_time():
-    return datetime.datetime.utcnow().replace(tzinfo=datetime.timezone.utc)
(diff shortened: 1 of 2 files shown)


src/ansible_runner/display_callback/callback/awx_display.py
```diff
@@ -70,7 +70,7 @@
 
 
 def current_time():
-    return datetime.datetime.utcnow().replace(tzinfo=datetime.timezone.utc)
+    return datetime.datetime.now(datetime.UTC)
```
Akasurde:
```suggestion
    return datetime.datetime.now(datetime.timezone.utc)
```

Alias `datetime.UTC` is added in Python 3.11

Akasurde (changes requested):

Shrews:
Superseded by https://github.com/ansible/ansible-runner/pull/1348 which adds proper 3.12 testing.

Issue #1292, opened 2023-08-10
SSH Agent Forwarding With MacOS and Docker Desktop


timway:
This impacts only Docker Desktop on MacOS when using `ansible-runner` in process isolation mode.

The `ENV` path of `cli_mounts` in `_handle_automounts` does not work with Docker Desktop on MacOS. Docker Desktop wants to share a socket that is local to its internals and not on the MacOS system; `/run/host-services/ssh-auth.sock`. Because this file doesn't exist it fails the first check in `_update_volume_mount_paths`. Because it does not get added to the list of volumes it cannot be consumed by the container.

https://github.com/ansible/ansible-runner/blob/aef73cf7265b807c877697e74585f74d58459366/src/ansible_runner/config/_base.py#L372-L374

This is documented by Docker Desktop at https://docs.docker.com/desktop/networking/#ssh-agent-forwarding and I have a similar bug open with `ansible-navigator` via https://github.com/ansible/ansible-navigator/issues/1591 as that also checks the file path in a similar way.

The socket is mounted as:

```bash
$ ls -lahZ /run/host-services/ssh-auth.sock 
srwxr-xr-x 1 root root ? 0 Aug  7 14:00 /run/host-services/ssh-auth.sock
```

It is only writable by `root` so this feature requires `root` inside the container. I know setting the default container user to root could be problematic so it will require discussion to determine whether `root` inside the container is a good default because it enables `ssh-agent` to work or if that should be a required step by those wishing to use `ssh-agent` and is there for simply documented.

PR #1293, opened 2023-08-10
Exclude `/run/host-services` When Using Docker Desktop on MacOS

### LINKED ISSUE #1593: SSH Agent Forwarding With VS Code Remote-SSH
##### ISSUE TYPE

- Bug Report

##### SUMMARY

Symbolic links are not followed when determining the volume mount paths for the SSH agent forwarding bits.

Running on Mac OS I add an SSH key with a passphrase into my SSH agent. Leveraging VS Code with the Remote-SSH plugin I forward the agent into remote host. This works fine, the remote host can run `ssh-add -l` and I see the key and am able to `ssh` into the managed node I want to use with `ansible-navigator`.

The VS Code Remote-SSH plugin sets up `SSH_AUTH_SOCK` to point to a file within the `tmpdir` configuration value in the plugin which seems to be `/run/user/<uid-of-remote-host-user>`. It essentially just creates a symbolic link to the normal file where the forwarded agent is placed by `ssh`.

##### ANSIBLE-NAVIGATOR VERSION

```bash
$ rpm -qi ansible-navigator
Name        : ansible-navigator
Version     : 3.4.1
Release     : 1.el8ap
Architecture: noarch
Install Date: Tue 08 Aug 2023 04:01:18 PM UTC
Group       : Unspecified
Size        : 1922971
License     : ASL 2.0
Signature   : RSA/SHA256, Thu 03 Aug 2023 05:49:14 PM UTC, Key ID 199e2f91fd431d51
Source RPM  : ansible-navigator-3.4.1-1.el8ap.src.rpm
Build Date  : Thu 03 Aug 2023 05:45:17 PM UTC
Build Host  : s390-064.build.eng.bos.redhat.com
Relocations : (not relocatable)
Packager    : Red Hat, Inc. <http://bugzilla.redhat.com/bugzilla>
Vendor      : Red Hat, Inc.
URL         : https://github.com/ansible/ansible-navigator
Summary     : A text-based user interface (TUI) for the Red Hat Ansible Automation Platform
Description :
A text-based user interface (TUI) for the Red Hat Ansible Automation Platform```

##### CONFIGURATION

##### LOG FILE

VS Code Remote-SSH Log

```bash
[20:16:06.380] Updating $SSH_AUTH_SOCK: ln -f -s "/tmp/ssh-xkYbL0MKH1/agent.14010" "/run/user/1000/vscode-ssh-auth-sock-114491209"
```

##### STEPS TO REPRODUCE

* Using MacOS, add an SSH key to the SSH agent with `ssh-add -i ...` command
* Install VS Code and the Remote-SSH plugin
* Use Remote-SSH to connect to a RHEL 8 system capable of running `ansible-navigator` (from the AAP 2.4 repos in my case)
* Run `ansible-navigator exec bash` and inspect the SSH agent with `ssh-add -l` and notice it fails to connect.

##### EXPECTED RESULTS

I'd expect the SSH agent is forwarded into the VM correctly.

##### ACTUAL RESULTS

It fails to connect to the SSH agent specified by `SSH_AUTH_SOCK` in the environment.

##### ADDITIONAL INFORMATION

I know this is a bit of an esoteric configuration but I keep trying different ways to use SSH keys particularly ones protected by passphrases with `ansible-navigator` starting from Mac OS and am just continuously coming up empty.

timway:
* Docker Desktop takes care of setting up ssh-agent forwarding from MacOS but does in a way that is only allowed to `root` inside the container
* As documented at the time of commit at https://docs.docker.com/desktop/networking/#ssh-agent-forwarding
* Move the `None` check ahead of the path checking with a specific error message to clearly differentiate that error from path checking errors

Partially addresses #1292 by exempting paths starting with `/run/host-services` in `containerized` mode on `docker`  from the source path check.

I'd like to hear feedback on making `--user root` the default for `docker` possibly when a volume mount has a path sourced in `/run/host-services` on `docker` or in all cases on `docker`.

A similar fix like this one will be required in order to resolve https://github.com/ansible/ansible-navigator/issues/1593 after/if this is merged into `ansible-runner` as well.

### DIFF
--- a/src/ansible_runner/config/_base.py
+++ b/src/ansible_runner/config/_base.py
@@ -397,9 +397,14 @@ class BaseConfig:
                                    dst_mount_path: str | None = None,
                                    labels: str | None = None
                                    ) -> None:
+        if src_mount_path is None:
+            logger.debug("Source volume mount path was not provided")
+            return
 
-        if src_mount_path is None or not os.path.exists(src_mount_path):
-            logger.debug("Source volume mount path does not exist: %s", src_mount_path)
+        if not os.path.exists(src_mount_path) and not (
+            self.containerized and self.process_isolation_executable == "docker" and os.path.abspath(src_mount_path).startswith("/run/host-services")
+        ):
+            logger.debug("Source volume mount path does not exit {0}".format(src_mount_path))
             return
 
         # ensure source is abs


timway:
Just checking in on this issue. What should I be doing to advance this?

I wrote a very targeted fix for the issue but I noticed it may be better to look holistically at the project. Would it be better to design an interface various process isolation methods could implement? That would make it easier to implement the various scenarios that runner is getting leveraged in these days. I'd be willing to take a shot at the initial design and see where it goes if necessary. I'd need to know where maintainers see things before spending substantial time on something like that though.

wmudge:
@timway - would be a +1 if this could be an option for `ansible-runner`, as other container engines like Colima provide the SSH socket in other paths. See https://github.com/abiosoft/colima/issues/127

erictoner:
+1 On this; this is absolutely needed for Docker if you want any kind of SSH forwarding from the host machine.

PR #1311, opened 2023-09-27
Allow shutil to copy into existing isolation directory

### LINKED ISSUE #1309: Directory isolation copying fails in sandboxed environments
The call to `shutil.copytree` when copying the project directory to the `directory_isolation_path` fails since `shutil.copytree`'s default functionality when the target directory already exists (which is created by the call to `tempfile.mkdtemp`) is to raise `FileExistsError` (ref: https://docs.python.org/3/library/shutil.html).

https://github.com/ansible/ansible-runner/blob/0594cea6e415af0d186b611e95e04195317cf816/src/ansible_runner/config/runner.py#L138C60-L138C60

We're running Python 3.11.5 in our environment.

christophert:
Fixes #1309 by adding `dirs_exist_ok=True` to the `copytree` call. Tested locally and the directory correctly populates with project directory files.

### DIFF
--- a/src/ansible_runner/config/runner.py
+++ b/src/ansible_runner/config/runner.py
@@ -135,7 +135,7 @@ class RunnerConfig(BaseConfig):
             self.directory_isolation_path = tempfile.mkdtemp(prefix='runner_di_', dir=self.directory_isolation_path)
             if os.path.exists(self.project_dir):
                 output.debug(f"Copying directory tree from {self.project_dir} to {self.directory_isolation_path} for working directory isolation")
-                shutil.copytree(self.project_dir, self.directory_isolation_path, symlinks=True)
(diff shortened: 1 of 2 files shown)


Shrews:
We need to dig into the history of this. It seems there is a `directory_isolation_cleanup` configuration setting that doesn't appear to be exposed that is supposed to control cleaning up of the isolation directory. I'm not certain we should _always_ just blindly overwrite it based on the existence of this setting.

christophert:
It looks like that setting would control cleaning it up once the run was completed. This change allows `copytree` to copy into the newly made directory by `tempfile.mkdtemp` in line 135, not any existing project isolation directory created by previous runs.

Shrews (approved):
`dirs_exist_ok` added in Python 3.8, so this should be ok.

PR #1312, opened 2023-10-02
Add lib symlink to bwrap call

### LINKED ISSUE #1310: bwarp can't execute any binaries due to missing /lib symlink
When testing sandbox execution, we were facing an issue where `bwrap` couldn't execute any commands:
```bash
gbwrap: execvp ansible-playbook: No such file or directory
```

In debug mode, we pulled the command out ourselves and tested it with a system binary which also resulted in the same error:
```bash
bwrap --die-with-parent --unshare-pid --dev-bind /dev dev --proc /proc --dir /tmp --ro-bind /bin /bin --ro-bind /etc /etc --ro-bind /usr /usr --ro-bind /opt /opt --symlink usr/lib64 /lib64 --bind /private_data_dir /private_data_dir --chdir /tmp/runner/runner_di_abcdefg id
bwrap: execvp id: no such file or directory
```

My system information:
```bash
$ uname -a
Linux fedora 6.0.11-200.fc36.aarch64 #1 SMP PREEMPT_DYNAMIC Fri Dec 2 20:16:22 UTC 2022 aarch64 GNU/Linux
$ bwrap --version
bubblewrap 0.7.0
```

I was able to get it working by adding `--symlink usr/lib /lib` to the arguments for `bwrap`, though I'm not sure if this has a wider ranging effect on x86_64 systems. It should be fine since most modern Linux systems symlink /lib to /usr/lib as well.

https://github.com/ansible/ansible-runner/blob/0594cea6e415af0d186b611e95e04195317cf816/src/ansible_runner/config/runner.py#L342C12-L342C12

christophert:
Fixes #1310 by adding a /usr/lib to /lib symlink in calls to `bwrap` for sandboxing.

### DIFF
--- a/src/ansible_runner/config/runner.py
+++ b/src/ansible_runner/config/runner.py
@@ -339,6 +339,7 @@ class RunnerConfig(BaseConfig):
             '--ro-bind', '/etc', '/etc',
             '--ro-bind', '/usr', '/usr',
             '--ro-bind', '/opt', '/opt',
+            '--symlink', 'usr/lib', '/lib',
             '--symlink', 'usr/lib64', '/lib64',
         ])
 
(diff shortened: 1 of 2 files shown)


Akasurde:
FYI this change is only required in case of `aarch64`

Shrews (approved):
I don't see any problem with this. It actually makes sense to symlink these dirs since we are already doing the lib64 variant.

PR #1315, opened 2023-10-03
Fix pylint running on test source files

### LINKED ISSUE #1313: Pylint not running against test files
We invoke `pytest` against our `test` directory in `tox.ini`, however, it isn't actually linting the test source files because we do not have `__init__.py` files in the `unit` or `integration` subdirectories (pylint accepts a list of python modules and packages, not directories). We need to add `__init__.py` to any subdirectories with sources files and fix the failures we are missing.

Shrews:
Fixes #1313

### DIFF
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -1,12 +1,14 @@
 import json
 import os
 import subprocess
-import yaml
 import pathlib
-import pytest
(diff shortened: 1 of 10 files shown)


Akasurde (approved):
LGTM

PR #1342, opened 2024-03-06
Untag instead of force remove image for podman

TheRealHaoLiu:
⚠️ Warning: this PR will cause the behavior to deviate from name of the param `--remove-images` that runs this code ⚠️ 

This is a behavioral change (and hopefully for the better) now cleanup_images  will behave the same for podman and docker

`docker rmi` will just untag while
`podman rmi` will untag and remove layers and cause runing container to be killed
for podman we use `untag` to achieve the same behavior

this only untag the image and does not delete the image prune_images need to be call to delete

### DIFF
--- a/src/ansible_runner/cleanup.py
+++ b/src/ansible_runner/cleanup.py
@@ -152,15 +152,29 @@ def cleanup_dirs(pattern: str, exclude_strings: list | None = None, grace_period
 
 
 def cleanup_images(images: list, runtime: str) -> int:
-    """Note: docker will just untag while podman will remove layers with same command"""
+    """
+    `docker rmi` will just untag while
+    `podman rmi` will untag and remove layers and cause runing container to be killed
+    for podman we use `untag` to achieve the same behavior
+
+    NOTE: this only untag the image and does not delete the image prune_images need to be call to delete
+    """
     rm_ct = 0
     for image_tag in images:
         stdout = run_command([runtime, 'images', '--format="{{.Repository}}:{{.Tag}}"', image_tag])
         if not stdout:
             continue
         for discovered_tag in stdout.split('\n'):
-            stdout = run_command([runtime, 'rmi', discovered_tag.strip().strip('"'), '-f'])
-            rm_ct += stdout.count('Untagged:')
+            if runtime == 'podman':
+                try:
+                    stdout = run_command([runtime, 'untag', image_tag])
+                    if not stdout:
+                        rm_ct += 1
+                except Exception:
+                    pass  # best effort untag
+            else:
+                stdout = run_command([runtime, 'rmi', discovered_tag.strip().strip('"'), '-f'])
+                rm_ct += stdout.count('Untagged:')
     return rm_ct
 
 


nitzmahone:
I get why people want *builder* to do `rmi` (though still ill-advised), but why is *runner* managing images at all?  At least under modern podman, just running container instances shouldn't be creating new images.

shanemcd:
@nitzmahone I was asking myself the same thing. I _think_ it was added to ansible-runner because that's the only thing other than receptor we install on execution nodes, which is where we need to run this.

AlanCoding:
My 2 cents - runner is effectively acting as a command allow-list here. If we could ship an arbitrary python file or bash script, it could be done that way, but perhaps, less securely. In the receptor mesh, the control nodes are only allowed to run `ansible-runner worker` commands. So we have to go through ansible-runner for anything AWX needs to run.

src/ansible_runner/cleanup.py
```diff
@@ -152,15 +152,29 @@ def cleanup_dirs(pattern: str, exclude_strings: list | None = None, grace_period
...
         stdout = run_command([runtime, 'images', '--format="{{.Repository}}:{{.Tag}}"', image_tag])
         if not stdout:
             continue
         for discovered_tag in stdout.split('\n'):
-            stdout = run_command([runtime, 'rmi', discovered_tag.strip().strip('"'), '-f'])
-            rm_ct += stdout.count('Untagged:')
+            if runtime == 'podman':
+                try:
+                    stdout = run_command([runtime, 'untag', image_tag])
+                    if not stdout:
```
jbradberry, line 171: I don't get it.  Why are we incrementing the count if there is no output?

TheRealHaoLiu: that's what podman does... it doesn't output anything when u run `podman untag` it will however output an error if it fail to untag

jbradberry: Why do the condition at all, then?  Rely on the logic getting derailed by the exception.

chadmf:
LGTM!!!

src/ansible_runner/cleanup.py
chadmf: agree with @jbradberry lets keep it simple

CFSNM (approved):

src/ansible_runner/cleanup.py
TheRealHaoLiu: it shouldn't be needed but it is possible, since this is best effort and not "critical" i rather the behavior to be move on instead of bail

gundalow: How about we get this merged as-in and do an ansible-runner release. Then we can can create a tech-debt task to follow up on this.

chadmf (approved):
LGTM!

src/ansible_runner/cleanup.py
jbradberry: @gundalow going to have to push back against that.  It's basically a one-line change, and if it's worth doing at all it should be done up-front.  If it merges as-is, it more or less means that we don't care about the possibility of podman's behavior changing and a follow-up task would just sink out of sight.

jbradberry: (I'm not opposed to merging as-is, but we shouldn't fool ourselves that we'll ever address such a minor concern unless it bites us in the future.)

AlanCoding (approved):

Shrews (approved):

PR #1345, opened 2024-03-13
Address event correlation issues in the awx_display callback

sivel:


src/ansible_runner/display_callback/callback/awx_display.py
```diff
@@ -311,7 +295,15 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
-Display.display = display_with_context(Display.display)
+for _display_attr in ('banner', 'deprecated', 'display', 'error',
+                      'system_warning', 'verbose', 'warning'):
```
sivel, line 311: The display events "supported" by awx are only:

```
        (0, 'debug', _('Debug'), False),
        (0, 'verbose', _('Verbose'), False),
        (0, 'deprecated', _('Deprecated'), False),
        (0, 'warning', _('Warning'), False),
        (0, 'system_warning', _('System Warning'), False),
        (0, 'error', _('Error'), True),
```

This would potentially apply to `unclassified_display` too.  Although the previous monkeypatching didn't exclude `banner`.

src/ansible_runner/utils/__init__.py
```diff
@@ -386,14 +386,14 @@ def _emit_event(self,
             event_data = self._current_event_data
             stdout_chunks = [buffered_stdout]
         elif buffered_stdout:
-            event_data = dict({'event': 'verbose'})
+            event_data = dict({'event': 'unclassified_display'})
```
sivel, line 389: See https://github.com/ansible/ansible-runner/pull/1345/files#r1523986425

src/ansible_runner/display_callback/callback/awx_display.py
```diff
@@ -352,6 +344,18 @@ def __init__(self):
...
+        base_config = C.config.get_configuration_definition(
+            DefaultCallbackModule._load_name, plugin_type='callback'
+        )
+        my_config = C.config.get_configuration_definition(
+            self._load_name, plugin_type='callback'
+        )
+        C.config.initialize_plugin_configuration_definitions(
+            'callback', self._load_name, base_config | my_config
+        )
+        return super().set_options(*args, **kwargs)
```
sivel, line 369: Not related, but since I'm in here making a mess, this fixes the options issue with callbacks not matching the default from a config perspective.

src/ansible_runner/utils/__init__.py
sivel: All of the CI failures appear to be related to this, where `verbose` is being excluded, or depended on. I'm not sure we can make this change, but I'd prefer to not just classify any unknown display event as verbose if possible.

PR #1379, opened 2024-07-01
Add Python 3.13 testing

Shrews:
Add testing environment for future Python 3.13 release.

- Python 3.13 seems to cause the `test_keepalive_setting()` integration tests to fail on a fairly consistent basis at the assert for the number of keepalive events (we exceed the hardcoded value of `5`). Speculation is that there is a change in this version that causes these tests to fail more consistently on a low-powered machine with test parallelization. Increasing the upper "wobble" value to account for this.
- Fixes an issue exposed by Python 3.13 with the `Base64IO` context manager and an undeclared attribute within the `close()` method.

Co-authored-by: Adam Williamson <awilliam@redhat.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,9 @@ jobs:
           - name: '3.12'
             tox_env: integration-py312
 
+          - name: '3.13.0-alpha - 3.13.0'
+            tox_env: integration-py313
+
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -121,6 +124,9 @@ jobs:
 
           - name: '3.12'
             tox_env: unit-py312
+
+          - name: '3.13.0-alpha - 3.13.0'
+            tox_env: unit-py313
 
     steps:
       - name: Checkout
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
           - name: '3.12'
             tox_env: integration-py312
 
-          - name: '3.13.0-alpha - 3.13.0'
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -65,6 +65,9 @@ jobs:
           - name: '3.12'
             tox_env: integration-py312
 
+          - name: '3.13.0-alpha - 3.13.0'
```
webknjaz: Optional improvement possibility. Ref https://github.com/ansible/ansible-builder/pull/685/files#r1688458186

.github/workflows/ci.yml
```diff
@@ -122,6 +125,9 @@ jobs:
           - name: '3.12'
             tox_env: unit-py312
 
+          - name: '3.13.0-alpha - 3.13.0'
```
webknjaz: Optional improvement possibility. Ref https://github.com/ansible/ansible-builder/pull/685/files#r1688458186

test/requirements.txt
```diff
@@ -8,4 +8,6 @@ pytest-xdist==2.5.0
 types-pyyaml
 flake8==6.1.0
 yamllint==1.32.0
+# cffi pre-release to make cryptography work with python 3.13 (remove when 3.13 is released)
+cffi==1.17.0rc1
```
webknjaz: Perhaps, may this conditional?
```suggestion
cffi==1.17.0rc1; python_version == "3.13"
```

Shrews: Oooh, yes, excellent idea.

webknjaz (approved):

PR #1383, opened 2024-07-26
fix: pin python-daemon<3.0.0 to avoid build errors

infinitewarp:
Builds of projects that require `ansible-runner` and thereby also require `python-daemon>=3.0.0` in environments with modern versions of `setuptools` are failing with this error:

```
ModuleNotFoundError: No module named 'setuptools.extern'
```

The problematic `extern` import in `python-daemon` was introduced in its `3.0.0` release and has since been fixed in that project's `main` branch, but there is not yet any release including that fix. Simply pinning an older version appears to resolve this build issue and does not appear to introduce any regressions.

After `python-daemon` releases a new version, assuming they do, then we probably should change this from `<3.0.0` to `>3.0.1`.

See also:

* https://pagure.io/python-daemon/issue/92
* https://github.com/quipucords/quipucords/pull/2714

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,7 +32,8 @@ python_requires = >=3.9
 install_requires =
     pexpect>=4.5
     packaging
-    python-daemon
+    # python-daemon 3.0.0 and 3.0.1 break builds with modern setuptools
+    python-daemon<3.0.0
     pyyaml
     # enable `groups` arg to entry_points missing in 3.9 stdlib importlib.metadata
     importlib-metadata>= 4.6,< 6.3; python_version<'3.10'


infinitewarp:
Note: the maintainer of `python-daemon` released a new version `3.0.2` in response to the aforementioned issue, but then they quickly yanked it because it introduced new worse issues with `setuptools`. I'll keep an eye on the `python-daemon` releases and update this PR with a working version if if becomes available soon.

Shrews:
It looks like you've pinned the version outside of runner in the `quipucords` repo, which is the proper fix to get you past this, IMO. I can see where we'd need to avoid the broken version of `python-daemon` here in runner, but I'd rather wait until a good fix has been released by that package so we can pin this once and not have to remember to modify it later.

We should also have a discussion internally about moving the pin on setuptools in this repo to a more recent version.

Shrews:
Looks to me like `python-daemon` version `3.0.1` is working as expected (PR #1385 in the Python 3.12 and 3.13 tests where `setuptools==72.1.0` and `python-daemon==3.0.1` are used). Both `3.0.0` and `3.0.2` have been yanked. Perhaps this pin is unnecessary at this point?

PR #1386, opened 2024-08-07
Add warning and trace that res is dropped

bcoca:
lets stop silently dropping data and giving no warning


Shrews:
@AlanCoding Since this touches `awx_display.py`, will let you or someone from your team check it out before approving.

PR #1397, opened 2024-09-24
Improve argument streaming for remote job execution (part 1)

### LINKED ISSUE #1324: Account for streaming payload splatting version incompatibilities
Right now payloads are splatted directly into the streaming classes, and ultimately into `ansible_runner.interface.run`.  Knowing that we need to, support a potential version mismatch of runner inside and outside of an EE, this blind splatting can lead to `TypeError` if we send too many args.

This likely needs to be accounted for in a more robust way.

Shrews:
# Changes to Remote Job Execution Argument Streaming

## Reason For Change

Remote job execution involves a `Transmitter` node streaming the job keyword arguments as a JSON object to a `Worker` node. Adding new keyword arguments to the `interface.run()` or `interface.run_async()` methods can cause errors within the remote job streaming interface when an older `Worker` node receives the new, unrecognized keyword argument from a newer `Transmitter` node (worker/transmitter version mismatch).

For more background information on the streaming process, see [this page of the documentation](https://ansible.readthedocs.io/projects/runner/en/2.4.0/remote_jobs/).

Fixes #1324

## Solution

The total fix will require a multi-stage solution:

 - First stage of the solution (implemented in this PR) will have the `Transmitter` process stream only keyword arguments that are actually specified and have a value different from their default value. This does not fix the case when a new keyword argument is introduced and used with a non-default value, but does fix older `Worker` nodes from failing when the new keyword is not used since the new keyword will not be sent in the argument stream.
 - The second stage (not implemented here) will have the `Worker` process gracefully fail on unrecognized keywords.

## Change Details

- [x]  `RunnerConfig` object will become the main source for supplying job parameters instead of `kwargs`.
    - Turns the `BaseConfig` and `RunnerConfig` classes into dataclasses to allow us to properly define arguments, their data types, and their default values.
        - Since the `__init__()` method is now autogenerated, care is taken to keep the same name for a few class attributes that did not have the same name as their argument. This is done through the use of Python class `property` decorators. Internally, the attribute is referenced by the non-alias version.
    - By default, all `RunnerConfig` attributes are streamable to a `Worker`. Attributes that are NEVER streamable will have explicit dataclass metadata to mark it as such (example, `field(metadata={MetaValues.TRANSMIT: False})`).
- [x] `interface.run()` and `interface.run_async()` changed to accept a `RunnerConfig` object.
    - Can now be used in one of two ways:
        - Old-style: method takes only `kwargs` parameters; backwards compatible
        - New-style: method takes a `RunnerConfig` object and a few select parameters.
    - Non-public methods lower in the call hierarchy (`init_runner()`, `dump_artifacts()`, etc) are modified to allow for receiving a `RunnerConfig` object instead of keyword arguments.
- [x] `Transmitter` modified to query the `RunnerConfig` object to retrieve the keyword arguments to transmit.
- [x] Includes a new porting guide in the documentation describing the change.
    - Parameter docs for `interface.run()` method moved to `BaseConfig` and `RunnerConfig` docs.
## TODO

- [ ] Testing with Controller?

### DIFF
--- a/src/ansible_runner/__main__.py
+++ b/src/ansible_runner/__main__.py
@@ -44,7 +44,8 @@ from yaml import safe_dump, safe_load
 from ansible_runner import run
 from ansible_runner import output
 from ansible_runner import cleanup
-from ansible_runner.utils import dump_artifact, Bunch, register_for_cleanup
+from ansible_runner._internal._dump_artifacts import dump_artifact
+from ansible_runner.utils import Bunch, register_for_cleanup
 from ansible_runner.utils.capacity import get_cpu_count, get_mem_in_bytes, ensure_uuid
 from ansible_runner.utils.importlib_compat import importlib_metadata
 from ansible_runner.runner import Runner
@@ -822,14 +823,11 @@ def main(sys_args=None):
             else:
                 vargs['inventory'] = abs_inv
 
-    output.configure()
-
-    # enable or disable debug mode
-    output.set_debug('enable' if vargs.get('debug') else 'disable')
-
-    # set the output logfile
+    debug = bool(vargs.get('debug'))
+    logfile = ''
     if ('logfile' in args) and vargs.get('logfile'):
-        output.set_logfile(vargs.get('logfile'))
+        logfile = vargs.get('logfile')
+    output.configure(debug, logfile)
 
     output.debug('starting debug logging')
 
--- a/src/ansible_runner/__main__.py
+++ b/src/ansible_runner/__main__.py
@@ -45,6 +45,7 @@ from ansible_runner import run
 from ansible_runner import output
 from ansible_runner import cleanup
 from ansible_runner._internal._dump_artifacts import dump_artifact
+from ansible_runner.defaults import default_process_isolation_executable
 from ansible_runner.utils import Bunch, register_for_cleanup
 from ansible_runner.utils.capacity import get_cpu_count, get_mem_in_bytes, ensure_uuid
 from ansible_runner.utils.importlib_compat import importlib_metadata
@@ -884,8 +885,8 @@ def main(sys_args=None):
                     "project_dir": vargs.get('project_dir'),
                     "artifact_dir": vargs.get('artifact_dir'),
                     "roles_path": [vargs.get('roles_path')] if vargs.get('roles_path') else None,
-                    "process_isolation": vargs.get('process_isolation'),
-                    "process_isolation_executable": vargs.get('process_isolation_executable'),
+                    "process_isolation": bool(vargs.get('process_isolation')),
+                    "process_isolation_executable": vargs.get('process_isolation_executable') or default_process_isolation_executable,
                     "process_isolation_path": vargs.get('process_isolation_path'),
                     "process_isolation_hide_paths": vargs.get('process_isolation_hide_paths'),
                     "process_isolation_show_paths": vargs.get('process_isolation_show_paths'),
--- a/src/ansible_runner/config/_base.py
+++ b/src/ansible_runner/config/_base.py
@@ -62,6 +62,11 @@ class BaseExecutionMode(Enum):
     GENERIC_COMMANDS = 2
 
 
+# Metadata string values
+class MetaValues(Enum):
+    STREAMABLE = 'streamable'
+
+
 @dataclass
 class BaseConfig:
     """The base configuration object.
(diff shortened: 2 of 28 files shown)


src/ansible_runner/config/_base.py
```diff
@@ -61,96 +62,154 @@ class BaseExecutionMode(Enum):
     GENERIC_COMMANDS = 2
 
 
+# Metadata string values
+class MetaValues(Enum):
+    # When True, value is emitted from Transmitter to send to Worker.
+    TRANSMIT = 'transmit'
+
+
+@dataclass
```
nitzmahone, line 71: Suggest conditionally adding `kw_only=True` to the decorator when >= 3.10 (until it becomes the minimum, then just hardcode), and document that keyword args are required. For an object with this many fields, posargs in init are kind of a maintenance nightmare for both us and users. Nothing we can do for 3.9, but at least we can dodge the ordering bullet for the future.

I see you're using posargs in some of the tests for `private_data_dir`, but even that one might potentially not be required in the future if some of the things people have been asking for come to pass.

src/ansible_runner/config/runner.py
```diff
@@ -60,52 +64,179 @@ class RunnerConfig(BaseConfig):
     >>> r = Runner(config=rc)
     >>> r.run()
 
+    This class inherites all the initialization parameters of the `BaseConfig` parent class, plus:
```
nitzmahone:
```suggestion
    This class inherits all the initialization parameters of the `BaseConfig` parent class, plus:
```

src/ansible_runner/config/runner.py
```diff
@@ -44,13 +48,13 @@ class ExecutionMode():
     RAW = 3
 
 
+@dataclass
```
nitzmahone, line 51: Same suggestion around conditionally setting `kw_only=True` when >= 3.10

nitzmahone (commented):
Didn't get all the way through the fine-toothed comb review before I ran out of days, but looking great- I think this solves the most important parts of the internal arg-splatting problem well.

src/ansible_runner/config/_base.py
Shrews: Let's discuss this when you return.

I did it this way to maintain full backward compatibility. For the cases where users instantiate their own `RunnerConfig` object, the first parameter (for `private_data_dir`) is positional. Making it `kw_only=True` will break that particular use case.

nitzmahone: After some discussion, I agree what you've got is probably the best option without incurring a lot of grief to add a runtime deprecation warning on usage of posargs.

Shrews: After discussion, going to leave this as-is for max backward compatibility but add an entry to the porting guide about how this positional parameter to RunnerConfig will be deprecated in the future.

Shrews:
Have to revamp this a bit. The config objects come into play for `transmit` and `process` phases with this change, but did not originally. This could cause data transformations that didn't happen before.

Shrews:
~~I think this needs modified to _at _least__ make sure that the `BaseConfig` object initialization does not do any active changes (making directories and the like) that will not be needed for the streaming objects. Possibly need to delay path modifications (e.g., setting absolute paths) as well.~~

src/ansible_runner/__main__.py
```diff
@@ -822,14 +824,11 @@ def main(sys_args=None):
...
-    # enable or disable debug mode
-    output.set_debug('enable' if vargs.get('debug') else 'disable')
-
-    # set the output logfile
+    debug = bool(vargs.get('debug'))
+    logfile = ''
     if ('logfile' in args) and vargs.get('logfile'):
-        output.set_logfile(vargs.get('logfile'))
+        logfile = vargs.get('logfile')
+    output.configure(debug, logfile)
```
Akasurde, line 831: Since you are here, on line 841, can you please change code to use `os.path.isdir` instead?

src/ansible_runner/config/runner.py
```diff
@@ -398,7 +529,7 @@ def wrap_args_for_sandbox(self, args):
 
     def handle_command_wrap(self):
         # wrap args for ssh-agent
-        if self.ssh_key_data:
+        if self.ssh_key:
             debug('ssh-agent agrs added')
```
Akasurde, line 533:
```suggestion
            debug('ssh-agent arguments added')
```

src/ansible_runner/__main__.py
Shrews: Not sure I understand... change `os.makedirs()` to `os.path.isdir()`???

PR #1410, opened 2025-01-21
Do not strip file from container_volume_mounts

### LINKED ISSUE #1409: ansible-runner destroys container_volume_mounts
If I want to mount a single file (in this case, a GPG agent socket) into an EE with `container_volume_mounts`, ansible-runner strips the filename away and tries to mount the whole directory. This results in a non-usable container.

I don't understand why the code in https://github.com/ansible/ansible-runner/blob/ec628a36072a0f573d162e48c7a464019aec3cd9/src/ansible_runner/config/_base.py#L422-L445 strips off the filename. If the user explicitly specifies a filename, that should be respected.

This is a minimal bugfix:
```diff
diff --git a/src/ansible_runner/config/_base.py b/src/ansible_runner/config/_base.py
index 819640d..3711577 100644
--- a/src/ansible_runner/config/_base.py
+++ b/src/ansible_runner/config/_base.py
@@ -432,7 +432,7 @@ class BaseConfig:
         self._ensure_path_safe_to_mount(dst_dir)
 
         # format the src dest str
-        volume_mount_path = f"{src_dir}:{dst_dir}"
+        volume_mount_path = f"{src_path}:{dst_path}"
 
         # add labels as needed
         if labels:
```

felixfontein:
Fixes #1409.

### DIFF
--- a/src/ansible_runner/config/_base.py
+++ b/src/ansible_runner/config/_base.py
@@ -395,7 +395,8 @@ class BaseConfig:
                                    args_list: list[str],
                                    src_mount_path: str | None,
                                    dst_mount_path: str | None = None,
-                                   labels: str | None = None
+                                   labels: str | None = None,
+                                   strip_file: bool = True,
                                    ) -> None:
 
         if src_mount_path is None or not os.path.exists(src_mount_path):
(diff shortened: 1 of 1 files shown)


oraNod (commented):
Looks good to me for what that's worth. I didn't go so far as building off this branch and verifying the fix but the changes are really clear. I was also able to observe the issue of files getting stripped out with runner version 2.4.0 and podman.

I won't add approval because I'm not a maintainer and don't have much knowledge of the codebase. Just thought I'd give a +1 for what seems like a straightforward and useful fix from the community. Thanks @felixfontein

Shrews:
Closed by #1408 merge.

PR #1411, opened 2025-01-22
Do not lock around writing to stdout, do not flush

sivel:
Do not lock around writing to stdout, do not flush

### DIFF
--- a/src/ansible_runner/display_callback/callback/awx_display.py
+++ b/src/ansible_runner/display_callback/callback/awx_display.py
@@ -34,10 +34,12 @@ import uuid
 from copy import copy
 
 # Ansible
+from ansible import __version__ as ansible_version_str
 from ansible import constants as C
 from ansible.plugins.callback import CallbackBase
 from ansible.plugins.loader import callback_loader
 from ansible.utils.display import Display
+from ansible.utils.multiprocessing import context as multiprocessing_context
 
 
 DOCUMENTATION = '''
@@ -66,6 +68,9 @@ else:
 DefaultCallbackModule: CallbackBase = callback_loader.get(default_stdout_callback).__class__
 
 CENSORED = "the output has been hidden due to the fact that 'no_log: true' was specified for this result"
+
+_ANSIBLE_VERSION = tuple(int(p) for p in ansible_version_str.split('.')[:2])
+_ANSIBLE_217 = _ANSIBLE_VERSION >= (2, 17)
 
 
 def current_time():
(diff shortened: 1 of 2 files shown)


src/ansible_runner/display_callback/callback/awx_display.py
```diff
@@ -222,31 +230,40 @@ def get_begin_dict(self):
...
+@functools.cache
+def _getsignature(f: t.Callable) -> inspect.Signature:
+    return inspect.signature(f)
+
+
+@functools.cache
+def _getcallargs(sig: inspect.Signature, *args: P.args, **kwargs: P.kwargs) -> types.MappingProxyType:
+    ba = sig.bind(*args, **kwargs)
+    ba.apply_defaults()
+    return types.MappingProxyType(ba.arguments)
```
sivel, line 263: This stuff is unnecessary for this PR, and if desired I can remove the last commit. Looking at things like this just drove me a little crazy:

```python
log_only = args[5] if len(args) >= 6 else kwargs.get('log_only', False)
```

sivel:
Reproducer for deadlock:

<details>
<summary><code>runme.py</code></summary>
<p>

```python
import os
import ansible_runner

r = ansible_runner.run(
    private_data_dir=os.path.abspath(os.getcwd()),
    playbook='playbook.yml',
    inventory='hosts',
    forks=50,
    settings=dict(
        suppress_ansible_output=True,
    ),
)
```

</p>
</details>

<details>
<summary><code>project/action_plugins/chatty.py</code></summary>
<p>

```python
from __future__ import annotations

from ansible.plugins.action import ActionBase
from ansible.utils.display import Display

display = Display()


class ActionModule(ActionBase):
    def run(self, tmp=None, task_vars=None):
        host = task_vars['inventory_hostname']
        for i in range(100):
            for a in ('display', 'v', 'vv', 'vvv', 'vvvv', 'vvvvv', 'vvvvvv', 'verbose', 'warning', 'debug'):
                getattr(display, a)(f'<{host}> chatty {i=} {a=}')
        return {}
```

</p>
</details>

Create an inventory with 100 hosts, and a playbook just like:

```yaml
- hosts: all
  gather_facts: false
  tasks:
    - chatty:
```

src/ansible_runner/display_callback/callback/awx_display.py
nitzmahone: You can't safely `functools.cache` the bound args this way- any non-hashable value passed to a display method would cause a runtime error.  There isn't really a benefit to caching this one in real-world usage anyway, since every unique combination of args is a new cache entry, so this would basically just be a memory leak bound by the size of the cache. Several of the display methods happily accept unhashable args, and last I checked there are a number of callers that blindly pass arbitrary objects to display methods.

Caching the signature this way should be fine (though you might want to verify that `inspect` isn't already doing so), since method objects are usually immutable/hashable.

sivel: `Signature` objects are hashable, and as far as I have been able to confirm they are not cached within `inspect`.

I was also skeptical of caching the `_getcallargs` function, purely for the concern of how much it would cache.  I'll remove that one.

sivel:
Closing in favor of https://github.com/ansible/ansible-runner/pull/1414

Issue #1415, opened 2025-02-21
make rotate_artifacts accessible from env/settings


flku-snp:
https://github.com/ansible/ansible-runner/blob/f95972b530045a0bf2701240ea4e5bcc03eced50/src/ansible_runner/config/_base.py#L151

Is that all needed to make the option available to the settings file? 

`        self.container_options = self.settings.get('rotate_artifacts ', self.rotate_artifacts)`

fkuep:
> ` self.container_options = self.settings.get('rotate_artifacts ', self.rotate_artifacts)`

        self.rotate_artifacts = self.settings.get('rotate_artifacts ', self.rotate_artifacts)

fkuep:
@Shrews Can You advise me on how to approach this, I dunno if I am doing this right.

Abhijeet213:
Clarify This Issueeee.....

PR #1417, opened 2025-02-21
Make rotate_artifacts accessible from env/settings

fkuep:
Related to #1415

Make `rotate_artifacts` accessible from the settings file.

* Modify `src/ansible_runner/config/_base.py`:
  - Retrieve `rotate_artifacts` from `settings` dictionary in `prepare_env` method.
  - Set `rotate_artifacts` in `__init__` method only if not provided in `settings` dictionary.
  - Ensure `rotate_artifacts` is set to 0 if not provided in command line or settings file.
  - Make `rotate_artifacts` default to 0 in `__init__` method.
  - Set `rotate_artifacts` from command line if provided.

* Modify `demo/env/settings`:
  - Add `rotate_artifacts` option to the settings file.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/ansible/ansible-runner/pull/1417?shareId=b4cd0f0e-8105-4ed0-b110-c7f379e7df3f).

### DIFF
--- a/src/ansible_runner/config/_base.py
+++ b/src/ansible_runner/config/_base.py
@@ -81,7 +81,7 @@ class BaseConfig:
                  container_workdir: str | None = None,
                  container_auth_data=None,
                  ident: str | None = None,
-                 rotate_artifacts: int = 0,
+                 rotate_artifacts: int | None = None,
                  timeout: int | None = None,
                  ssh_key: str | None = None,
                  quiet: bool = False,
@@ -110,7 +110,7 @@ class BaseConfig:
         self.container_options = container_options
 
         # runner params
-        self.rotate_artifacts = rotate_artifacts
+        self.rotate_artifacts = rotate_artifacts if rotate_artifacts is not None else 0
         self.quiet = quiet
         self.json_mode = json_mode
         self.passwords = passwords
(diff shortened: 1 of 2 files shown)


fkuep:
This one seems to work, only I wonder if the settings file should really win over the command line argument.

src/ansible_runner/config/_base.py
```diff
@@ -81,7 +81,7 @@ def __init__(self,
                  container_workdir: str | None = None,
                  container_auth_data=None,
                  ident: str | None = None,
-                 rotate_artifacts: int = 0,
+                 rotate_artifacts: int | None = None,
```
Shrews, line 84: I'm not seeing a reason for this change. If you revert it back, you can also just revert your change below on line 113, and it will work as you expect.

Shrews (changes requested):
Thanks for the PR!

I think you're on the right track. I think all you really need is your change on line 230 of `_base.py`. The other changes seem unnecessary, unless there is something I'm not seeing?

As for the settings file contents winning over the command line, I probably would have reversed that if I had had a say at the time the project was created, but it is what it is, so let's just be consistent with the other settings values.

You'll also need to update the documentation in `intro.rst` for the `env/settings` section.

Also, you'll want to add some tests  to validate this works as expected (we won't accept it without tests). You might need to be a bit creative here. When my group took over this code base, we worked to improve the existing set of tests, but we are lacking in certain areas still, and I'm not sure if we currently have a good example for you to use for a new test. I took a very quick look through the existing tests and didn't see any good candidates. See what you can come up with. If you have problems, I can try to assist you, but may be delayed in getting around to it.

PR #1418, opened 2025-02-24
feat: add option to mask potential secrents in env vars 

kokosnuss:
Ansible-Runner is used within AWX to run ansible-playbooks in transmit and worker mode.
Since AWX passes the contents (user + password) of some credential types (e.g. VMware vcenter) to the playbook as environment variables, these environment variables are written to stdout in `streaming.py`:
https://github.com/ansible/ansible-runner/blob/devel/src/ansible_runner/streaming.py#L228
`self._output.write(json.dumps(printed_status_data).encode('utf-8'))`
  
In environments such as Red Hat Openshift where global log forwarding to a log streaming server is enabled for all pods and their stdout, this immediately leads to the logging of sensitive data into log management.
This PR offers the possibility to suppress the logging of all environment variables by ansible-runner via CLI flag (`--suppress-env-print`) or environment variable (`SUPPRESS_ENV_PRINT`).
  
Due to the fact that awx uses a hard-coded CLI-call for the ansible-runner, it's important to allow log-suppressipon not only via CLI flag but also via environment variable. Otherwise, it would not be configurable:
https://github.com/ansible/awx/blob/devel/awx/main/tasks/receptor.py#L646
The suppress option can here be achieved via setting the environmet variable `SUPPRESS_ENV_PRINT=True` using the pod specification for the ansible-runner.

### DIFF
--- a/src/ansible_runner/__main__.py
+++ b/src/ansible_runner/__main__.py
@@ -875,7 +875,7 @@ def main(sys_args=None):
                 if vargs.get('suppress_env_print'):
                     suppress_env_print = vargs.get('suppress_env_print')
                 else:
-                    suppress_env_print = os.getenv('SUPPRESS_ENV_PRINT', 'False') == 'True'
+                    suppress_env_print = os.getenv('SUPPRESS_ENV_PRINT', 'False').lower() == 'true'
                 run_options = {
                     "private_data_dir": vargs.get('private_data_dir'),
                     "ident": vargs.get('ident'),
--- a/src/ansible_runner/__main__.py
+++ b/src/ansible_runner/__main__.py
@@ -56,7 +56,6 @@ DEFAULT_RUNNER_BINARY = os.getenv('RUNNER_BINARY', None)
 DEFAULT_RUNNER_PLAYBOOK = os.getenv('RUNNER_PLAYBOOK', None)
 DEFAULT_RUNNER_ROLE = os.getenv('RUNNER_ROLE', None)
 DEFAULT_RUNNER_MODULE = os.getenv('RUNNER_MODULE', None)
-DEFAULT_RUNNER_SUPPRESS_ENV_PRINT = os.getenv('SUPPRESS_ENV_PRINT', None)
 DEFAULT_UUID = uuid4()
 
 DEFAULT_CLI_ARGS = {
@@ -628,6 +627,12 @@ def main(sys_args=None):
         dest="worker_info",
         action="store_true",
         help="show the execution node's Ansible Runner version along with its memory and CPU capacities"
+    )
+    worker_subparser.add_argument(
+        "--suppress-env-print",
+        dest="suppress_env_print",
+        action="store_true",
+        help="add flag to prevent the printing of env vars on stdout. Also set via SUPPRESS_ENV_PRINT"
     )
     worker_subparser.add_argument(
         "--delete",
--- a/src/ansible_runner/__main__.py
+++ b/src/ansible_runner/__main__.py
@@ -633,7 +633,6 @@ def main(sys_args=None):
         "--suppress-env-print",
         dest="suppress_env_print",
         action="store_true",
-        default=DEFAULT_RUNNER_SUPPRESS_ENV_PRINT,
         help="add flag to prevent the printing of env vars on stdout. Also set via SUPPRESS_ENV_PRINT"
     )
     worker_subparser.add_argument(
@@ -874,6 +873,10 @@ def main(sys_args=None):
 
         with context:
             with role_manager(vargs) as vargs:
+                if vargs.get('suppress_env_print'):
+                    suppress_env_print = vargs.get('suppress_env_print')
+                else:
+                    suppress_env_print = os.getenv('SUPPRESS_ENV_PRINT', 'False') == 'True'
                 run_options = {
                     "private_data_dir": vargs.get('private_data_dir'),
                     "ident": vargs.get('ident'),
@@ -908,7 +911,7 @@ def main(sys_args=None):
                     "limit": vargs.get('limit'),
                     "streamer": streamer,
                     "suppress_env_files": vargs.get("suppress_env_files"),
-                    "suppress_env_print": vargs.get('suppress_env_print') if vargs.get('suppress_env_print') else DEFAULT_RUNNER_SUPPRESS_ENV_PRINT,
+                    "suppress_env_print": suppress_env_print,
                     "keepalive_seconds": vargs.get("keepalive_seconds"),
                 }
                 try:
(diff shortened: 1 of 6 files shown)


Shrews:
Thanks for your PR.

On changes like this, we have to ask someone from the Controller/AWX team to test the change to validate that it won't adversely affect that system (ping @AlanCoding or @john-westcott-iv). It might take some time for this change to be tested, so until that feedback can be given, I'd suggest investing some time in adding some unit/integration tests to this PR as we will not merge a PR that is not testing its own changes.

Also, I'd like to point out that the AWX UI provides a way to use encrypted VMWare credentials (see https://ansible.readthedocs.io/projects/awx/en/24.6.1/userguide/credentials.html#vmware-vcenter), although I'm not personally able to guide you in that aspect. This is likely to be a more secure option for you than using unencrypted environment variables.

kokosnuss:
> On changes like this, we have to ask someone from the Controller/AWX team to test the change to validate that it won't adversely affect that system (ping @AlanCoding or @john-westcott-iv). It might take some time for this change to be tested, so until that feedback can be given, I'd suggest investing some time in adding some unit/integration tests to this PR as we will not merge a PR that is not testing its own changes.

Yes sure, we will provide some unit/integration tests.
> 
> Also, I'd like to point out that the AWX UI provides a way to use encrypted VMWare credentials (see https://ansible.readthedocs.io/projects/awx/en/24.6.1/userguide/credentials.html#vmware-vcenter), although I'm not personally able to guide you in that aspect. This is likely to be a more secure option for you than using unencrypted environment variables.

That is exactly the way we are working. Unfortunately AWX passes the encrypted VMWare credentials ans env vars instead of ansible extra_vars to the play.

Klaas-:
This sounds like a security issue :)

kokosnuss:
@Shrews thanks for starting the action workflow.
We've now added a integration tests.

src/ansible_runner/__main__.py
```diff
@@ -866,6 +872,10 @@ def main(sys_args=None):
 
         with context:
             with role_manager(vargs) as vargs:
+                if vargs.get('suppress_env_print'):
+                    suppress_env_print = vargs.get('suppress_env_print')
+                else:
+                    suppress_env_print = os.getenv('SUPPRESS_ENV_PRINT', 'False') == 'True'
```
Akasurde:
```suggestion
                    suppress_env_print = os.getenv('SUPPRESS_ENV_PRINT', 'False').lower() == 'true'
```

This will match all `True`, `true` and other combinations of cases.
Not a blocker though.

Shrews:
Random out loud thought... might be worth considering removing `env` at [this point in the code](https://github.com/ansible/ansible-runner/blob/f95972b530045a0bf2701240ea4e5bcc03eced50/src/ansible_runner/runner.py#L106) so that suppressing will work even outside of the streaming protocol.

Also, unclear to me at this point that _ever_ printing out env vars at the above point is going to be useful.

kokosnuss:
> Random out loud thought... might be worth considering removing `env` at [this point in the code](https://github.com/ansible/ansible-runner/blob/f95972b530045a0bf2701240ea4e5bcc03eced50/src/ansible_runner/runner.py#L106) so that suppressing will work even outside of the streaming protocol.
> 
> Also, unclear to me at this point that _ever_ printing out env vars at the above point is going to be useful.

Thanks for that not so random thought. Initially i had the same idea but there are components in AWX/tower which are requiring that ansible-runner prints at least one key-value pair in env since awx is iterating through the env vars at some point.

Anyways, we discovered that a project sync in awx did not work anymore after we included this fix in our execution environment:
`2025-02-28 17:32:33,562 ERROR    [54203f1c4a454517afe9a8c15ab62e10] awx.main.tasks.jobs inventory_update 55180 (running) Post run hook errored.
  File "/var/lib/awx/venv/awx/lib64/python3.11/site-packages/awx/main/tasks/jobs.py", line 637, in run
    self.post_run_hook(self.instance, status)
  File "/var/lib/awx/venv/awx/lib64/python3.11/site-packages/awx/main/tasks/jobs.py", line 1645, in post_run_hook
    private_data_dir = inventory_update.job_env['AWX_PRIVATE_DATA_DIR']
                       ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: 'AWX_PRIVATE_DATA_DIR'`
Turns out that awx is reading the ansible_private_data_dir from the env key printed from ansible-runner:
https://github.com/ansible/awx/blob/7b8b37d9a8e46b3cda101f3c5691c870b9de46da/awx/main/tasks/jobs.py#L1664
That's why we have rewritten the PR into a “mask-secrets-in-env”. We copied the code for this from awx itself.

Shrews:
Thanks for the PR. It's a little problematic (through no fault of your own) in that it suffers from a problem we've been trying to address within runner when introducing a new `interface.run()` job argument, as your PR does. Within streaming mode, if the `transmit` runner is not the same version as the `worker` runner, the worker could crash when it receives a job argument that it does not understand. We currently have a [proposed fix up](https://github.com/ansible/ansible-runner/pull/1397) that is awaiting testing from the Controller team. I think we should reconsider this fix _after_ that big fix merges.

gannaramu:
+1 on this issue. need a way to suppress env variables to stdout. leaking secrets to upstream logging.

AlanCoding:
At first glance I don't think this should break anything with AWX. You made the changes to `Worker` which reports the env vars in a callback. All that AWX does with that is used it to populate `job_env`. This is not fed into anything else in the server logic itself, merely presented in the API details of the job.

You're supporting it as a ansible-runner setting, which is how support for it would be enabled for AWX. There is some additional sanitization/hiding on the AWX layer. After all, it is what provides the env vars to begin with, so it already tracks what it knows should be hidden. The only value I could see for AWX would be that the system or podman configuration adds additional secrets that need to be hidden.

src/ansible_runner/utils/__init__.py
```diff
@@ -520,3 +520,23 @@ def _handler(number, frame):
...
 
     return signal_event.is_set
+
+
+def build_safe_env(env):
+    """
+    Build environment dictionary, hiding potentially sensitive information
+    such as passwords or keys.
+    """
+    hidden_re = re.compile(r'API|TOKEN|KEY|SECRET|PASS|PWD|CRED|AUTH', re.I)
```
gannaramu:
```suggestion
    hidden_re = re.compile(r'API|TOKEN|KEY|SECRET|PASS|PWD|CRED|AUTH|PAT', re.I)
```

sivel:
I meant to come back to this issue a while ago, and I apologize for that. However, the end result of the discussion we have had are that we have no intentions of implementing or accepting such a feature at this time.

Our recommendation is to implement security restrictions within the logging platform to prevent access to the sensitive data for anyone who does not otherwise have access to view the info.

Ultimately, the only way we could approach the needs of the platform, and prevent this from happening would be to implement a new non-stdout mechanism for communication between the execution node and ansible-runner.  However using stdout has been determined to give us the highest level of reliability for communication, due to reliability implemented via k8s and other container platforms in ensuring stdout is not lost.

PR #1434, opened 2025-07-26
Python 3.14 compat: replace codecs.open with open

AdamWill:
Python 3.14 deprecates codecs.open:
https://docs.python.org/3.14/whatsnew/3.14.html#deprecated This makes the tests fail, because we get a bunch of warnings about the deprecation and I guess the tests are set to fail on warnings.

It's never been recommended to use codecs.open on Python 3. It was the "right way" to open text content on Python 2 for a bit, but when Python 3 came around, io.open was supposed to be used to be agnostic between 2 and 3. Since ansible-runner only targets 3 nowadays, let's just use open.

### DIFF
--- a/src/ansible_runner/loader.py
+++ b/src/ansible_runner/loader.py
@@ -21,7 +21,6 @@ from __future__ import annotations
 
 import os
 import json
-import codecs
 
 from typing import Any, Dict
 from yaml import safe_load, YAMLError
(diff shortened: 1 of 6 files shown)


Shrews:
FYI, tests are currently broken with latest core 2.19 release. Working on resolution to that....

Shrews:
If you would, go ahead and rebase against `devel` to pick up the CI changes we just merged. I was going to go ahead and add Python 3.14 testing as well, but tests keep timing out with that version, so probably best to just go ahead and get your changes in before adding 3.14 to the matrix.

AdamWill:
Done.

Shrews (approved):
Thanks for the contribution.

PR #1435, opened 2025-07-29
Fix CI for core 2.19.0

Shrews:
Update selected CI tests to skip when tested against `ansible-core` 2.19.0 or higher.

Assisted-by: Cursor

### DIFF
--- a/test/integration/conftest.py
+++ b/test/integration/conftest.py
@@ -4,15 +4,38 @@ import subprocess
 import pathlib
 import random
 
+from functools import cache
 from string import ascii_lowercase
 
 import pexpect
 import pytest
 import yaml
+from packaging.version import parse as parse_version
 
 from ansible_runner.config.runner import RunnerConfig
+from ansible_runner.utils.importlib_compat import importlib_metadata
+
 
 here = pathlib.Path(__file__).parent
+
+
+@cache
+def get_ansible_version():
+    """Get the ansible version string."""
+    return importlib_metadata.version("ansible-core")
+
+
+@pytest.fixture(scope='session')
+def is_ansible_219_or_higher():
+    """Return True if ansible-core version is >= 2.19.0."""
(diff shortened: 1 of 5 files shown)


sivel (approved):
I'm good with this.

fwiw for others, this is an attempt at a temporary fix for CI to get tests passing, while understanding there are some bits of functionality incompatible with 2.19 around fact caching.

The fact cache poking needs to be deprecated/removed, but we have some decisions to make in core that could impact this, and determine what method we take for removing/deprecating the fact cache poking.

test/integration/test_events.py
```diff
@@ -125,7 +125,7 @@ def test_playbook_on_stats_summary_fields(project_fixtures):
         assert runner_stats[stat]  # expected at least 1 host in each stat type
 
 
-def test_include_role_events(project_fixtures):
+def test_include_role_events(project_fixtures, skipif_ansible_219_or_higher):  # pylint: disable=W0613
```
webknjaz, line 128: It's usually nicer to have a `@pytest.mark.usefixtures('skipif_ansible_219_or_higher')` for unused arguments as it will make it clear what you actually access in the test function + you wouldn't have to deal with pointless suppressions. Additionally, I wanted to call out that suppressions with rule numbers in pylint aren't easy to follow and it's best to use human-readable identifiers instead.

test/integration/conftest.py
```diff
@@ -4,17 +4,40 @@
...
+from packaging.version import parse as parse_version
 
 from ansible_runner.config.runner import RunnerConfig
+from ansible_runner.utils.importlib_compat import importlib_metadata
+
 
 here = pathlib.Path(__file__).parent
 
 
+@cache
```
webknjaz, line 22: Looks like this is only called once per `pytest` invocation. So this caching is rather premature...

PR #1442, opened 2025-08-18
Add non-voting tests against ansible-core devel branch

Shrews:
New unit and integration tests using `devel` branch of `ansible-core`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,6 +156,72 @@ jobs:
     secrets:
       codecov-token: ${{ secrets.CODECOV_TOKEN }}
 
+  devel-tests:
+    name: 🧪 Devel-Tests${{ '' }}
+
+    needs:
+    - pre-setup
+
+    strategy:
+      fail-fast: >-  # ${{ runner.debug }} is unavailable in this context
+        ${{
+          fromJSON(needs.pre-setup.outputs.is-debug-mode)
+          && false
+          || true
+        }}
+      matrix:
+        python-version:
+        - 3.13
+        runner-vm-os:
+        - ubuntu-24.04
+        toxenv:
+        - unit-devel
+        - integration-devel
+        xfail:
+        - true
+
+    uses: tox-dev/workflow/.github/workflows/reusable-tox.yml@89de3c6be3cd179adf71e28aa4ac5bef60804209  # yamllint disable-line rule:line-length
+    with:
+      cache-key-for-dependency-files:  # FIXME
+      check-name: >-
+        ${{ matrix.toxenv }}
+        under 🐍${{ matrix.python-version }}
+      job-dependencies-context: >-  # context for hooks
+        ${{ toJSON(needs) }}
+      python-version: >-
+        ${{ matrix.python-version }}
+      require-successful-codecov-uploads: >-
+        ${{
+          toJSON(
+          needs.pre-setup.outputs.upstream-repository-id
+          == github.repository_id
+          )
+        }}
+      runner-vm-os: >-
+        ${{ matrix.runner-vm-os }}
+      timeout-minutes: >-
+        ${{ matrix.toxenv == 'unit-devel' && 2 || 5 }}
+      toxenv: >-
+        ${{ matrix.toxenv }}
+      tox-run-posargs: >-
+        --cov-report=xml:.tox/.tmp/.test-results/pytest-${{
+          matrix.python-version
+        }}/cobertura.xml
+        --junitxml=.tox/.tmp/.test-results/pytest-${{
+          matrix.python-version
+        }}/test.xml
+      tox-rerun-posargs: >-
+        --no-cov
+        -vvvvv
+        --lf
+      tox-tool-deps: >-
+        'tox == 4.25.0'
+        'tox-uv == 1.25.0'
+      xfail: >-
+        ${{ fromJSON(matrix.xfail) }}
+    secrets:
+      codecov-token: ${{ secrets.CODECOV_TOKEN }}
+
   check:  # This job does nothing and is only used for the branch protection
     name: ✅ All tests passed
     if: always()
@@ -164,6 +230,7 @@ jobs:
     - pre-setup  # transitive, for accessing settings
     - sanity
     - tests
+    - devel-tests
 
     runs-on: ubuntu-latest
 
@@ -174,3 +241,4 @@ jobs:
       uses: re-actors/alls-green@release/v1
       with:
         jobs: ${{ toJSON(needs) }}
+        allowed-failures: devel-tests
--- a/tox.ini
+++ b/tox.ini
@@ -12,8 +12,10 @@ pytest_cov_args = --cov --cov-report html --cov-report term --cov-report xml
 
 [testenv]
 description = Run tests with {basepython}
-deps = ansible-core
-       -r {toxinidir}/test/requirements.txt
+deps =
+    !devel: ansible-core
+    {unit,integration}-devel: ansible-core @ https://github.com/ansible/ansible/archive/devel.tar.gz
+    -r {toxinidir}/test/requirements.txt
 passenv =
     HOME
     RUNNER_TEST_IMAGE_NAME
@@ -30,15 +32,13 @@ commands =
     mypy src/ansible_runner
     pylint src/ansible_runner test
 
-[testenv:unit{,-py310,-py311,-py312,-py313,-py314}]
+[testenv:unit{,-py310,-py311,-py312,-py313,-py314,-devel}]
 description = Run unit tests
 commands =
     pytest test/unit -vv -n auto {[shared]pytest_cov_args} {posargs}
 
-[testenv:integration{,-py310,-py311,-py312,-py313,-py314}]
+[testenv:integration{,-py310,-py311,-py312,-py313,-py314,-devel}]
 description = Run integration tests
-deps =
-    {[testenv]deps}
 commands =
     pytest test/integration -vv -n auto {[shared]pytest_cov_args} {posargs}
 
@@ -51,6 +51,7 @@ commands =
 [testenv:clean]
 description = Erase docs and coverage artifacts
 skip_install = True
+base =
 allowlist_externals = /bin/sh
 commands =
     /bin/sh -c "rm -rf {toxinidir}/test/coverage/*"
--- a/test/integration/test_core_integration.py
+++ b/test/integration/test_core_integration.py
@@ -5,7 +5,6 @@ from ansible_runner.interface import run
 
 
 TEST_BRANCHES = (
-    'devel',
     'milestone',
     'stable-2.19',   # current stable
     'stable-2.18',   # stable - 1


.github/workflows/ci.yml
```diff
@@ -157,6 +157,72 @@ jobs:
...
+      matrix:
+        python-version:
+        - 3.13
+        runner-vm-os:
+        - ubuntu-latest
+        toxenv:
+        - unit-devel
+        - integration-devel
+        xfail:
+        - false
```
webknjaz: If you make this `true`, it'll make the CI ignore the job outcomes.

.github/workflows/ci.yml
```diff
@@ -157,6 +157,72 @@ jobs:
...
+        --lf
+      tox-tool-deps: >-
+        'tox == 4.25.0'
+        'tox-uv == 1.25.0'
+      xfail: >-
+        ${{ fromJSON(matrix.xfail) }}
+    secrets:
+      codecov-token: ${{ secrets.CODECOV_TOKEN }}
+
   check:  # This job does nothing and is only used for the branch protection
```
webknjaz, line 225: Make `devel-tests` a dependency and then add it to `allowed-failures` under `with:` below.

.github/workflows/ci.yml
Shrews: Done

.github/workflows/ci.yml
Shrews: Done

tox.ini
```diff
@@ -12,8 +12,8 @@ pytest_cov_args = --cov --cov-report html --cov-report term --cov-report xml
 
 [testenv]
 description = Run tests with {basepython}
-deps = ansible-core
-       -r {toxinidir}/test/requirements.txt
+deps =
+    -r {toxinidir}/test/requirements.txt
```
webknjaz, line 18: Try this:
```suggestion
    !devel: ansible-core
    devel: ansible-core @ https://github.com/ansible/ansible/archive/devel.tar.gz
    -r {toxinidir}/test/requirements.txt
```

You shouldn't need to  have this much copied env definitions. Using factors it's more concise.

Those docs got lost in v4 refactoring but it still works: https://github.com/tox-dev/tox/issues/3276.

tox.ini
```diff
@@ -32,20 +35,40 @@ commands =
...
+deps =
+    git+https://github.com/ansible/ansible.git@devel
+    {[testenv]deps}
+commands =
+    pytest test/unit -vv -n auto {[shared]pytest_cov_args} {posargs}
+
+[testenv:integration-devel]
+description = Run integration tests against ansible-core devel
+deps =
+    git+https://github.com/ansible/ansible.git@devel
```
webknjaz:
```suggestion
```

tox.ini
```diff
@@ -32,20 +35,40 @@ commands =
 
 [testenv:unit{,-py39,-py310,-py311,-py312,-py313,-py314}]
 description = Run unit tests
+deps =
+    ansible-core
+    {[testenv]deps}
```
webknjaz:
```suggestion
```

tox.ini
```diff
@@ -22,6 +22,9 @@ commands = pytest -vv -n auto {posargs}
 
 [testenv:linters{,-py39,-py310,-py311,-py312,-py313,-py314}]
 description = Run code linters
+deps =
+    ansible-core
+    {[testenv]deps}
```
webknjaz:
```suggestion
```

tox.ini
```diff
@@ -32,20 +35,40 @@ commands =
...
+    git+https://github.com/ansible/ansible.git@devel
     {[testenv]deps}
 commands =
     pytest test/integration -vv -n auto {[shared]pytest_cov_args} {posargs}
 
 [testenv:docs]
 description = Build documentation
-deps = -r{toxinidir}/docs/requirements.txt
+deps =
+    -r{toxinidir}/docs/requirements.txt
```
webknjaz:
```suggestion
deps = -r{toxinidir}/docs/requirements.txt
```

test/integration/conftest.py
```diff
@@ -154,21 +146,18 @@ def container_image(request, cli, tmp_path):  # pylint: disable=W0621
...
-ARG WHEEL
-COPY $WHEEL /$WHEEL
+FROM quay.io/centos/centos:stream10
 
-# Need python 3.11 minimum for devel
-RUN dnf install -y python3.11 python3.11-pip git
-RUN alternatives --install /usr/bin/python3 python3 /usr/bin/python3.11 0
-RUN python3 -m pip install /$WHEEL git+https://github.com/ansible/ansible@{branch}
+RUN dnf install -y python3-pip git
+RUN python3 -m pip install git+https://github.com/ansible/ansible@{branch}
```
webknjaz:
```suggestion
RUN python3 -m pip install 'ansible-core @ https://github.com/ansible/ansible/archive/{branch}.tar.gz'
```

test/integration/conftest.py
```diff
@@ -154,21 +146,18 @@ def container_image(request, cli, tmp_path):  # pylint: disable=W0621
...
-FROM quay.io/centos/centos:stream9
-ARG WHEEL
-COPY $WHEEL /$WHEEL
+FROM quay.io/centos/centos:stream10
 
-# Need python 3.11 minimum for devel
-RUN dnf install -y python3.11 python3.11-pip git
-RUN alternatives --install /usr/bin/python3 python3 /usr/bin/python3.11 0
-RUN python3 -m pip install /$WHEEL git+https://github.com/ansible/ansible@{branch}
+RUN dnf install -y python3-pip git
```
webknjaz:
```suggestion
RUN dnf install -y python3-pip
```

test/integration/Dockerfile
```diff
@@ -1,11 +1,9 @@
 # This is not fully compatible builder Dockerfile, but meets the needs of tests
-FROM quay.io/centos/centos:stream9
-ARG WHEEL
-COPY $WHEEL /$WHEEL
+FROM quay.io/centos/centos:stream10
 RUN dnf install -y python3-pip
-RUN python3 -m pip install /$WHEEL ansible-core
+RUN python3 -m pip install ansible-core
```
webknjaz: Don't we need to have runner in the env still?
```suggestion
RUN python3 -m pip install . ansible-core
```

.github/workflows/ci.yml
```diff
@@ -157,6 +157,72 @@ jobs:
...
+        ${{
+          toJSON(
+          needs.pre-setup.outputs.upstream-repository-id
+          == github.repository_id
+          )
+        }}
+      runner-vm-os: >-
+        ${{ matrix.runner-vm-os }}
+      timeout-minutes: >-
+        ${{ matrix.toxenv == 'unit' && 2 || 10 }}
```
webknjaz: It's possible that with switching to installing from Git archives over HTTP will make the job run time fast again.

test/integration/Dockerfile
webknjaz: Ah, never mind. I just saw that it's not needed in the PR description.

webknjaz:
> Removes the runner wheel building inside pytest. This was used to install runner within a testing container image, but we never actually invoke `ansible-runner` in the container, only `ansible-playbook`. This unnecessary work was causing tests to run slower

Does this mean that the increased timeout is no longer needed? Or maybe it could be increased less dramatically?

> and the occasional build race condition, because it was being done for every test that was done within a container.

@Shrews would it make sense to change the `container_image` scope to `'session'`, then?

tox.ini
Shrews: That's an interesting idea, though I generally dislike using factors because if you're not familiar with them, they're a bit difficult to understand and aren't as obvious as explicitly defined test envs. I might try this though, only slightly altered (don't want `-devel` available for every test env).

.github/workflows/ci.yml
Shrews: I'll try that!

Shrews:
> > Removes the runner wheel building inside pytest. This was used to install runner within a testing container image, but we never actually invoke `ansible-runner` in the container, only `ansible-playbook`. This unnecessary work was causing tests to run slower
> 
> Does this mean that the increased timeout is no longer needed? Or maybe it could be increased less dramatically?
 
I hope so. Keeping it higher for right now so I can get complete tests until this is no longer a draft.


> > and the occasional build race condition, because it was being done for every test that was done within a container.
> 
> @Shrews would it make sense to change the `container_image` scope to `'session'`, then?

I think that still needs to be at `function` scope.

webknjaz:
> > @Shrews would it make sense to change the `container_image` scope to `'session'`, then?
> 
> I think that still needs to be at `function` scope.

Yeah, I see now that it's a bunch of DI spaghetti. I believe it can be refactored to allow for switching the scope, though.

CONTRIBUTING.md
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
 ```
-## Linting and Unit Tests
 
-`tox` is used to run linters (`flake8` and `yamllint`) and tests.
+## Running Tests
 
+`tox` is used to run the linting, unit, and integration tests. You can install it with:
+
+```
+$ pip install tox
```
webknjaz: Can we suggest app managers instead of generic `pip`? With `pip`, the users would have to know they have to maintain isolated per-app venvs.
```suggestion
$ pipx install tox
```

Oftentimes, the best option for people is to use the distro package manager and do `dnf install tox`.

This should also link upstream instead of trying to maintain a copy of a subset of instructions: https://tox.wiki/en/stable/installation.html

CONTRIBUTING.md
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
+```
+
+### Linting Tests
+
+To run all linting tests:
+
+```
+$ tox -e linters
```
webknjaz: Use tox4 syntax instead of the legacy fallbacks:
```suggestion
$ tox r -e linters
```

CONTRIBUTING.md
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
+```
+
+### Unit Tests
+
+To run all unit tests:
+
+```
+$ tox -e unit
```
webknjaz:
```suggestion
$ tox r -e unit
```

CONTRIBUTING.md
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
 ```
-(host)$ pip install tox
-(host)$ tox
+
+To execute a subset of unit tests, supply the specific test(s) to `tox`:
+
 ```
+$ tox -e unit -- test/unit/test_runner.py
+$ tox -e unit -- test/unit/test_cleanup.py::test_registry_auth_cleanup
```
webknjaz:
```suggestion
$ tox r -e unit -- test/unit/test_runner.py
$ tox r -e unit -- test/unit/test_cleanup.py::test_registry_auth_cleanup
```

CONTRIBUTING.md
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
+```
+
+To run unit tests against the `devel` version of `ansible-core`, append `-devel` to the
+tox test environment name:
 
+```
+$ tox -e unit-devel
```
webknjaz:
```suggestion
$ tox r -e unit-devel
```

CONTRIBUTING.md
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
+
 ```
+$ tox -e unit -- test/unit/test_runner.py
+$ tox -e unit -- test/unit/test_cleanup.py::test_registry_auth_cleanup
+```
+
+To run unit tests against the `devel` version of `ansible-core`, append `-devel` to the
+tox test environment name:
 
+```
```
webknjaz:
````suggestion
```console
````

File: CONTRIBUTING.md [hunk 18]
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
+$ tox -e unit-devel
+```
+
+### Integration Tests
+
+To run all integration tests:
+
+```
+$ tox -e integration
```
webknjaz:
````suggestion
```console
$ tox r -e integration
````

File: CONTRIBUTING.md [hunk 19]
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
+```
+$ tox -e integration -- test/integration/test_runner.py
+$ tox -e integration -- test/integration/test_runner.py::test_run_command_ansible
+```
+
+To run integration tests against the `devel` version of `ansible-core`, append `-devel` to the
+tox test environment name:
+
+```
+$ tox -e integration-devel
```
webknjaz:
````suggestion
```console
$ tox r -e integration-devel
````

File: CONTRIBUTING.md [hunk 20]
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
+
+```
+$ tox -e integration
+```
+
+To execute a subset of integration tests, supply the specific test(s) to `tox`:
+
+```
+$ tox -e integration -- test/integration/test_runner.py
+$ tox -e integration -- test/integration/test_runner.py::test_run_command_ansible
```
webknjaz:
````suggestion
```console
$ tox r -e integration -- test/integration/test_runner.py
$ tox r -e integration -- test/integration/test_runner.py::test_run_command_ansible
````

File: CONTRIBUTING.md [hunk 21]
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
+
+```
+$ tox -e unit
 ```
-(host)$ pip install tox
-(host)$ tox
+
+To execute a subset of unit tests, supply the specific test(s) to `tox`:
+
 ```
```
webknjaz:
````suggestion
```console
````

File: CONTRIBUTING.md [hunk 22]
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
+
+```
+$ tox -e linters
+```
+
+### Unit Tests
+
+To run all unit tests:
+
+```
```
webknjaz:
````suggestion
```console
````

File: CONTRIBUTING.md [hunk 23]
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
+$ pip install tox
+```
+
+### Linting Tests
+
+To run all linting tests:
+
+```
```
webknjaz:
````suggestion
```console
````

File: CONTRIBUTING.md [hunk 24]
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
+$ workon ansible-runner
 ```
-## Linting and Unit Tests
 
-`tox` is used to run linters (`flake8` and `yamllint`) and tests.
+## Running Tests
 
+`tox` is used to run the linting, unit, and integration tests. You can install it with:
+
+```
```
webknjaz:
````suggestion
```console
````

File: CONTRIBUTING.md [hunk 25]
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
 When done making changes, run:
 
 ```
-(host)$ deactivate
+$ deactivate
 ```
 
 To reactivate the virtual environment:
 
 ```
```
webknjaz:
````suggestion
```console
````

File: CONTRIBUTING.md [hunk 26]
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
 When done making changes, run:
 
 ```
```
webknjaz:
````suggestion
```console
````

File: CONTRIBUTING.md [hunk 27]
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
-(host)$ deactivate
+$ deactivate
 ```
 
 To reactivate the virtual environment:
 
 ```
-(host)$ workon ansible-runner
+$ workon ansible-runner
```
webknjaz: FWIW, I think that we should show how to use `tox devenv` for this.

CONTRIBUTING.md
```diff
@@ -24,23 +24,74 @@ In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readt
...
 ```
-## Linting and Unit Tests
 
-`tox` is used to run linters (`flake8` and `yamllint`) and tests.
+## Running Tests
 
+`tox` is used to run the linting, unit, and integration tests. You can install it with:
```
webknjaz: It'd be a good idea to mention that `tox l` would output the list of envs in a generic way.

test/integration/conftest.py
```diff
@@ -140,7 +132,7 @@ def container_image(request, cli, tmp_path):  # pylint: disable=W0621
     image_name = f'ansible-runner-{random_string}-event-test'
 
     cli(
-        [runtime, 'build', '--build-arg', f'WHEEL={wheel.name}', '--rm=true', '-t', image_name, '-f', str(dockerfile_path), str(tmp_path)],
+        [runtime, 'build', '--build-arg', '--rm=true', '-t', image_name, '-f', str(dockerfile_path), str(tmp_path)],
```
webknjaz: Is this breaking things?
```suggestion
        [runtime, 'build', '--rm=true', '-t', image_name, '-f', str(dockerfile_path), str(tmp_path)],
```

test/integration/conftest.py
```diff
@@ -184,22 +173,14 @@ def container_image_devel(request, cli, tmp_path):  # pylint: disable=W0621
...
-
     runtime = request.getfixturevalue('runtime')
     dockerfile_path = tmp_path / 'Dockerfile'
     dockerfile_path.write_text(DOCKERFILE)
     random_string = ''.join(random.choice(ascii_lowercase) for i in range(10))
     image_name = f'ansible-runner-{random_string}-event-test'
 
     cli(
-        [runtime, 'build', '--build-arg', f'WHEEL={wheel.name}', '--rm=true', '-t', image_name, '-f', str(dockerfile_path), str(tmp_path)],
+        [runtime, 'build', '--build-arg', '--rm=true', '-t', image_name, '-f', str(dockerfile_path), str(tmp_path)],
```
webknjaz:
```suggestion
        [runtime, 'build', '--rm=true', '-t', image_name, '-f', str(dockerfile_path), str(tmp_path)],
```

CONTRIBUTING.md
Shrews: > It'd be a good idea to mention that `tox l` would output the list of envs in a generic way.

Because of the use of factors, the list output is kind of incorrect (another reason I'm not fond of using them). It mentions a `devel` target that doesn't really exist. But I'm trying not to make this a tox tutorial anyway, just giving folks (and AI) the basics.

webknjaz:
@Shrews would you increase the timeout so we could see the output of `--durations` in the slow job? I'm guessing that this is probably related to building a new container and that's probably happening multiple times. Maybe the network/IO/CPU gets throttled due to this...
Interestingly, the old integration test jobs have like 487 lines of output, while the devel job produces 4332 lines before it gets cancelled.

P.S. I'd go for refactoring the fixture before this PR.

CONTRIBUTING.md
webknjaz: It kinda does exist, though. But it's good to mention it being "fake" in the description field.

tox.ini
```diff
@@ -12,8 +12,10 @@ pytest_cov_args = --cov --cov-report html --cov-report term --cov-report xml
 
 [testenv]
 description = Run tests with {basepython}
-deps = ansible-core
-       -r {toxinidir}/test/requirements.txt
+deps =
+    !devel: ansible-core
+    unit-devel,integration-devel: ansible-core @ https://github.com/ansible/ansible/archive/devel.tar.gz
```
webknjaz:
```suggestion
    {unit,integration}-devel: ansible-core @ https://github.com/ansible/ansible/archive/devel.tar.gz
```

tox.ini
```diff
@@ -12,8 +12,10 @@ pytest_cov_args = --cov --cov-report html --cov-report term --cov-report xml
 
 [testenv]
 description = Run tests with {basepython}
```
webknjaz, line 14:
```suggestion
description = Run tests with {basepython}
```
```suggestion
description =
    Run tests with {basepython}
    !unit-devel,!integration-devel: DO NOT USE the `devel` factor without `unit` or `integration`
```

tox.ini
```diff
@@ -52,6 +51,7 @@ commands =
 [testenv:clean]
```
webknjaz, line 51: FYI tox4 has a `base` env setting now: https://tox.wiki/en/latest/config.html#base

Wonder if it's possible to reset it to "nothing" with an empty string:
```suggestion
[testenv:clean]
base =
```

Shrews:
This PR is getting too large. Going to break it up into smaller changes, some of which we should be able to backport to 2.4.

test/integration/conftest.py
Shrews: Good spot. No, didn't break anything, simply ignored.

Shrews:
> @Shrews would you increase the timeout so we could see the output of `--durations` in the slow job? I'm guessing that this is probably related to building a new container and that's probably happening multiple times. Maybe the network/IO/CPU gets throttled due to this... Interestingly, the old integration test jobs have like 487 lines of output, while the devel job produces 4332 lines before it gets cancelled.
> 

FWIW, these tests depend on rebuilding new containers, so that's expected. The interesting thing is that even the unit tests are much slower, and there's no container building going on there.  I bumped both timeouts, so they should not get cancelled for this latest push  (*crossing fingers*).

tox.ini
Shrews: Is there documentation on how factors work within a `description`? I couldn't find any and find this confusing. I would expect that warning to NOT show up with `unit-devel` or `integration-devel` just based on what I know about factors, but they do.

webknjaz: I think they work the same in every field. But I was making suggestions from browser via GH UI so it's possible I missed something. Maybe, it's because of the comma, I haven't tested locally.

Factors definitely work in description: https://github.com/sphinx-doc/sphinx/blob/e8ab5cf1c57d106f57cb6c77b60dc4a5ae2c9a37/tox.ini#L21-L22. But perhaps I messed up the negation..

webknjaz: > I would expect that warning to NOT show up with `unit-devel` or `integration-devel` just based on what I know about factors, but they do.

I think that `!unit-devel` means anything that has a factor of `devel` that does not have a factor of `unit`. I wouldn't expect negation to work for both factors, you'd have to negate each. Is this what confuses you? Have you checked locally?

CONTRIBUTING.md
```diff
@@ -2,45 +2,138 @@
...
+## Running Tests
+
+`tox` is used to run the linting, unit, and integration tests. It automatically creates isolated testing environments
+and executes the appropriate test tools.
 
+**Common test commands:**
+- `tox run -e linters` - Code style and quality checks
+- `tox run -e unit` - Fast unit tests
+- `tox run -e integration` - Slower end-to-end tests
+- `tox run -e linters,unit` - Run linting and unit tests together
```
webknjaz:
```suggestion
- `tox run -e linters,unit` - Run linting and unit tests sequentially
- `tox parallel -e integration,unit` - Run integration and unit tests simultaneously
```

CONTRIBUTING.md
```diff
@@ -2,45 +2,138 @@
...
 - All code and doc submissions are done through pull requests against the `devel` branch.
 - Take care to make sure no merge commits are in the submission, and use `git rebase` vs `git merge` for this reason.
-- We ask all of our community members and contributors to adhere to the [Ansible code of conduct]. If you have questions, or need assistance, please reach out to our community team at [codeofconduct@ansible.com].
+- We ask all of our community members and contributors to adhere to the [Ansible code of conduct]. If you have questions,
+  or need assistance, please reach out to our community team at [codeofconduct@ansible.com].
 
 ## Setting up your development environment
 
-In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readthedocs.io/en/latest/), but any virtual environment will do.
+The recommended way to set up a development environment is using `tox --devenv` which creates a virtual environment
```
webknjaz: `tox --devenv` is actually `tox legacy --devenv` (tox3). It's best to use the tox4 interface:
```suggestion
The recommended way to set up a development environment is using `tox devenv` which creates a virtual environment
```

CONTRIBUTING.md
```diff
@@ -2,45 +2,138 @@
...
+with all necessary dependencies. You will need to either install it with your system package manager, or use `pip`:
+
+```console
+$ pip install tox
+```
+
+Once installed, create your own development environment and activate it with:
+
+```console
+$ tox --devenv venv
```
webknjaz:
```suggestion
$ tox devenv venv
```

CONTRIBUTING.md
```diff
@@ -2,45 +2,138 @@
...
+$ deactivate
 ```
 
-When done making changes, run:
+## Quick Start Workflow
+
+Here's a typical workflow for contributing:
+
+1. **Fork and clone** the repository
+2. **Set up environment**: `pip install tox && tox --devenv venv && source venv/bin/activate`
```
webknjaz:
```suggestion
2. **Set up environment**: `pip install tox && tox devenv venv && source venv/bin/activate`
```

CONTRIBUTING.md
```diff
@@ -2,45 +2,138 @@
...
+  or need assistance, please reach out to our community team at [codeofconduct@ansible.com].
 
 ## Setting up your development environment
 
-In this example we are using [virtualenvwrapper](https://virtualenvwrapper.readthedocs.io/en/latest/), but any virtual environment will do.
+The recommended way to set up a development environment is using `tox --devenv` which creates a virtual environment
+with all necessary dependencies. You will need to either install it with your system package manager, or use `pip`:
+
+```console
+$ pip install tox
```
webknjaz: I still think that it's a generally harmful advice to guide people towards installing `tox` into an arbitrary env. In general, pointing towards `pipx` / `uvx` by default is safer and is what the upstream community are doing more these days.
Additionally, it might be a good idea to suggest installing `tox-uv` in the same env.

`pipx install tox && pipx inject tox tox-uv` or `uv tool install tox --with tox-uv` woudld be so much nicer in here.

CONTRIBUTING.md
```diff
@@ -2,45 +2,138 @@
...
+
+Once installed, create your own development environment and activate it with:
+
+```console
+$ tox --devenv venv
+$ source venv/bin/activate
+```
+
+This creates a virtual environment in the `venv` directory with ansible-runner installed in development mode along
+with all dependencies.
```
webknjaz: Note that you can select the environment just like with other subcommands. This would look like `tox devenv -e units-devel devel-experiments-venv`. It might be useful to mention in the context of troubleshooting tests in a venv directly.

webknjaz (approved):
Are you planning to take my contrib doc notes to a separate PR?

Shrews:
> Are you planning to take my contrib doc notes to a separate PR?

Most likely, yes. It was becoming a distraction from the main changes.

PR #1451, opened 2025-09-23
Docs: correct/extend password regex (#1398)

### LINKED ISSUE #1398: Stuck on "SSH password" prompt
Greetings.

Version: 2.4.0 and 2.3.6 (at least)

If we supply passwords (ssh, become and vault) it gets stuck at the SSH prompt.

We bascially supply the secrets like this:
```python3
args['ssh_key'] = creds.ssh_key
args['passwords'] = {}
args['passwords'][r'^SSH\s[pP]assword:\s*$'] = creds.connect_pass
args['passwords'][r'^BECOME\s[pP]assword:\s*$'] = creds.become_pass
args['passwords'][r'^Vault\s[pP]assword:\s*$'] = creds.vault_pass
RunnerConfig(**args)
```

I've debugged it and the password is actually sent here: https://github.com/ansible/ansible-runner/blob/2.4.0/src/ansible_runner/runner.py#L321

But the execution is stuck forever (not even a timeout?) at this point.

I've tried adding `\r\n` to the password and even using pexpect's `sendcontrol` - but without any change.
If another `sendline` is performed directly after the SSH prompt - it is unstuck. But the next prompt fails as it seems this empty-input is used for that one..

I've got it working for now as seen in this fork: https://github.com/ansibleguy/ansible-runner/blob/release_2.4/src/ansibleguy_runner/runner.py#L323
But that solution is not really clean.

superstes:
This PR updates/corrects the documented password-regex(es) that users needs to use.
It can close #1398

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -102,9 +102,9 @@ In order for **Runner** to respond with the correct password, it needs to be abl
 by providing a yaml or json formatted file with a regular expression and a value to emit, for example::
 
   ---
-  "^SSH\s[pP]assword:\s*$": 'some_password',
-  "^BECOME\s[pP]assword.*:\s*$": 'become_password',
-  "^Vault\s[pP]assword:\s*$": 'vault_password',
+  r"^SSH\s[pP]assword:\s*$": 'some_password',
+  r"^BECOME\s[pP]assword.*:\s*$": 'become_password',
+  r"^Vault\s[pP]assword:\s*$": 'vault_password',
 
 ``env/cmdline``
 ---------------
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -102,9 +102,9 @@ In order for **Runner** to respond with the correct password, it needs to be abl
 by providing a yaml or json formatted file with a regular expression and a value to emit, for example::
 
   ---
-  r"^SSH\s[pP]assword:\s*$": 'some_password',
-  r"^BECOME\s[pP]assword.*:\s*$": 'become_password',
-  r"^Vault\s[pP]assword:\s*$": 'vault_password',
+  r"^SSH\spassword:\s*$": 'some_password',
+  r"^BECOME\spassword.*:\s*$": 'become_password',
+  r"^Vault\spassword:\s*$": 'vault_password',
 
 ``env/cmdline``
 ---------------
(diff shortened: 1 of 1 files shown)


docs/intro.rst
```diff
@@ -102,8 +102,9 @@ In order for **Runner** to respond with the correct password, it needs to be abl
 by providing a yaml or json formatted file with a regular expression and a value to emit, for example::
 
   ---
-  "^SSH password:\\s*?$": "some_password"
-  "^BECOME password.*:\\s*?$": "become_password"
+  r"^SSH\s[pP]assword:\s*$": 'some_password',
+  r"^BECOME\s[pP]assword.*:\s*$": 'become_password',
+  r"^Vault\s[pP]assword:\s*$": 'vault_password',
```
Shrews:
```suggestion
  r"^SSH password:\s*$": 'some_password',
  r"^BECOME password.*:\s*$": 'become_password',
  r"^Vault password:\s*$": 'vault_password',
```

Shrews (changes requested):
I think we can simplify these regex's slightly (suggestion inline). Otherwise looks good!

docs/intro.rst
superstes: Yes - seems these are always lowercase.
References:
* SSH: https://github.com/ansible/ansible/blob/stable-2.20/lib/ansible/cli/__init__.py#L348
* Become: https://github.com/ansible/ansible/blob/stable-2.20/lib/ansible/cli/__init__.py#L346
  Seems like there could be edge-cases where `BECOME` will be replaced - but the users will be able to catch those themselves
* Vault: https://github.com/ansible/ansible/blob/stable-2.20/lib/ansible/cli/__init__.py#L254

docs/intro.rst
```diff
@@ -102,9 +102,9 @@ In order for **Runner** to respond with the correct password, it needs to be abl
 by providing a yaml or json formatted file with a regular expression and a value to emit, for example::
 
   ---
-  r"^SSH\s[pP]assword:\s*$": 'some_password',
-  r"^BECOME\s[pP]assword.*:\s*$": 'become_password',
-  r"^Vault\s[pP]assword:\s*$": 'vault_password',
+  r"^SSH\spassword:\s*$": 'some_password',
+  r"^BECOME\spassword.*:\s*$": 'become_password',
+  r"^Vault\spassword:\s*$": 'vault_password',
```
Shrews, line 107:
```suggestion
  r"^SSH password:\s*$": 'some_password',
  r"^BECOME password.*:\s*$": 'become_password',
  r"^Vault password:\s*$": 'vault_password',
```

Shrews (changes requested):
The `\s` in between expected words is overly complex. Just use a single space.

PR #1454, opened 2025-09-30
feat: improve eof log messages when streaming

PabloHiro:
The reason of this change is to provide a meaningful error message whenever the streaming of logs gets aborted.

Multiple users have reported that a simple trace mentioning that an empty line is not valid JSON is not helpful to understand that in most cases, the infrastructure under the hood is experiencing problems.

### DIFF
--- a/src/ansible_runner/streaming.py
+++ b/src/ansible_runner/streaming.py
@@ -344,8 +344,26 @@ class Processor:
         while True:
             try:
                 line = self._input.readline()
+                if len(line) == 0:
+                    self.status_callback({
+                        'status': 'error',
+                        'job_explanation': (
+                            'Unexpected empty line encountered during worker stream. '
+                            'Worker did not produce events or streaming was aborted, check execution node health.'
+                        )
+                    })
+                    break
+            except (OSError) as exc:
+                self.status_callback({
+                    'status': 'error',
+                    'job_explanation': (
+                        f'Failed to read from worker stream. Error: {exc}'
+                    )
+                })
+                break
+            try:
                 data = json.loads(line)
-            except (json.decoder.JSONDecodeError, IOError) as exc:
+            except (json.decoder.JSONDecodeError) as exc:
                 self.status_callback({
                     'status': 'error',
                     'job_explanation': (
--- a/test/integration/test_transmit_worker_process.py
+++ b/test/integration/test_transmit_worker_process.py
@@ -517,3 +517,42 @@ def test_unparsable_really_big_line_processor(tmp_path):
         private_data_dir=process_dir,
         status_handler=status_receiver
     )
+
+
+def test_error_on_empty_line_processor(tmp_path):
+    process_dir = tmp_path / 'for_process'
+    process_dir.mkdir()
+    incoming_buffer = io.BytesIO(b'')
+
+    def status_receiver(status_data, runner_config):  # pylint: disable=unused-argument
+        assert status_data['status'] == 'error'
+        assert (
+            'Unexpected empty line encountered during worker stream. '
+            'Worker did not produce events or streaming was aborted, check execution node health.'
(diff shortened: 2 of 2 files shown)


src/ansible_runner/streaming.py
```diff
@@ -342,8 +342,16 @@ def run(self):
             os.makedirs(job_events_path, 0o700, exist_ok=True)
 
         while True:
+            line = self._input.readline()
+            if len(line) == 0:
+                self.status_callback({
+                    'status': 'error',
+                    'job_explanation': (
+                        'Unexpected EOF encountered during worker stream. Worker did not produce events or streaming was aborted, check execution node health.'
```
dleehr: Do we know with certainty that an `EOF` was encountered? I think we need to be precise here since ansible-runner also writes messages to the stream that reference an in-band `eof` like these:

https://github.com/ansible/ansible-runner/blob/1999bc639a550c3c66cf57f22d4da4eecf2313b7/src/ansible_runner/streaming.py#L246
https://github.com/ansible/ansible-runner/blob/1999bc639a550c3c66cf57f22d4da4eecf2313b7/src/ansible_runner/streaming.py#L63

So I'm concerned mentioning EOF here will confuse the issue.

While the underlying cause for that may have been an EOF in another system, I think what we know for certain is the length of the line is `0`. So maybe the message should call out an unexpected empty line rather than calling it an EOF.

PabloHiro: Fully agree. Updated!

src/ansible_runner/streaming.py
```diff
@@ -342,8 +342,17 @@ def run(self):
             os.makedirs(job_events_path, 0o700, exist_ok=True)
 
         while True:
+            line = self._input.readline()
```
sivel: It appears as though where this line was before, it was wrapped in an a `try/except` that caught `IOError`.

I'm assuming that `IOError` was there specifically for the `readline()`.

So it probably needs rewrapped, and the `IOError` removed from the below `except`

Shrews:
This is going to need some tests, too. Maybe have a look at `test/integration/test_transmit_worker_process.py` for inspiration.

src/ansible_runner/streaming.py
PabloHiro: Awesome catch! I changed it to `OSError`, because according to this:
https://docs.python.org/3/library/exceptions.html

`IOError` got merged into `OSError` in python 3.3. I am still testing for `IOError` in the tests though. Thoughts?

Shrews: I would change the test to match the code.

test/integration/test_transmit_worker_process.py
```diff
@@ -6,6 +6,7 @@
 import concurrent.futures
 import time
 import threading
+from unittest.mock import Mock
```
Shrews: We prefer `pytest` over `unittest`. Grep for `mocker` to see usages in the test code.

test/integration/test_transmit_worker_process.py
```diff
@@ -517,3 +518,42 @@ def status_receiver(status_data, runner_config):  # pylint: disable=W0613
...
+        status_handler=status_receiver,
+    )
+
+
+def test_error_on_io_error_processor(tmp_path):
+    process_dir = tmp_path / 'for_process'
+    process_dir.mkdir()
+
+    incoming_buffer = Mock()
+    incoming_buffer.readline.side_effect = IOError("Simulated IO error")
```
Shrews: No reason to use `IOError` here. Best to match the code and use `OSError`.

Shrews (changes requested):

Shrews (approved):

sivel (approved):

dleehr (approved):

PR #1456, opened 2025-10-07
Migrate setup.cfg to pyproject.toml

Shrews:


sivel (approved):
I always forget that flake8 doesn't support pyproject.toml for some reason.

lgtm

PR #1461, opened 2025-10-15
Restore SSH agent automounting for containerized Execution Environments

### LINKED ISSUE #1396: #689 broke SSH key features when running EE as the process isolation
The runner docs seems to indicate that when run using an EE, then it should mount in the SSH agent and/or the host users ~/.ssh (as long as it is not symlinked). Ref: [Using Runner with Execution Environments — Ansible Runner Documentation 1](https://ansible.readthedocs.io/projects/runner/en/2.4.0/execution_environments/#ssh-symlinks)

However when running a playbook that is present inside the EE, against a target, then it fails to authenticate. Adding the debug flag indicates that the runner in fact doesn’t add any mount options when executing podman, that would handle any SSH keys.

I have been trying to get my head around the intent of the code, but it's really hard to match the docs to the code and trying to guess what is intentional and what is dead code or code waiting(?) to be moved over to ansible-navigator.

Anyways, i think i have found two issues?

1. Both [runner.py](https://github.com/ansible/ansible-runner/blob/devel/src/ansible_runner/config/runner.py#L399) and [_base.py](https://github.com/ansible/ansible-runner/blob/devel/src/ansible_runner/config/_base.py#L342) defines the `handle_command_wrap` function.  However, the version in [runner.py#L415](https://github.com/ansible/ansible-runner/blob/devel/src/ansible_runner/config/runner.py#L415) explicitly sets `BaseExecutionMode.NONE` as the execution mode, when calling `wrap_args_for_containerization`, regardless of the `self.execution_mode` being set earlier based on the actual execution mode defined through the CLI options.

2. In #689, the `wrap_args_for_containerization` function in _base.py got refactored, and quite a substantial part of this functions logic got wrapped in the following if statement:
https://github.com/ansible/ansible-runner/blob/1bad4599d9236d5a89f262f65d4f09f97be40b50/src/ansible_runner/config/_base.py#L516-L519
effectively ensuring that none of that code will ever(?) be executed, as the execution mode is enforced to be `NONE` as per finding 1 above.  Specifically it skips the call to the `_handle_automounts` function, which seems to be responsible for trying to handle SSH credentials.

Besides this, isn't it kind of defeating the purpose of the container isolation, when using `--ipc=host`? I assume this is primarily to allow access to the `ssh-agent` socket?

yashx007:
### Root cause

In #689, a refactor changed how containerization was applied for `run()`-based flows by forcing `BaseExecutionMode.NONE`.
This bypassed internal wrapper logic - specifically the automount handling for `SSH_AUTH_SOCK` - which previously ensured the SSH agent socket was bind-mounted into the container.

As a result, containerized Execution Environment (EE) runs no longer received the SSH agent socket, breaking workflows that relied on SSH agent forwarding (e.g. Git over SSH, remote hosts with keyless auth via agent).

This PR restores that behavior using a runtime fallback approach that safely re-inserts the socket mount and environment variable when needed.

####  Code changes

* **`src/ansible_runner/config/_base.py`**

  * Re-enabled socket automount detection for wrapper logic, with ordering preserved.
  * Skips explicit socket mounting for `RunnerConfig` to preserve CLI compatibility.

* **`src/ansible_runner/runner.py`**

  * Added a **runtime fallback** that:

    * Detects if the container is being used.
    * Checks for a present SSH agent socket.
    * Inserts a conservative `-v <sock>:<dest>` and `-e SSH_AUTH_SOCK=<dest>` into the final container command.
    * **Writes these values into artifacts** for audit and reproducibility, even when wrapper logic is skipped.

####  Behavior notes

* Wrapper avoids breaking legacy CLI calls by not injecting mounts for `RunnerConfig`; the fallback ensures the socket is mounted when needed.
* A lenient `/tmp/ssh-*` heuristic supports environments like **WSL** and **rootless Podman**, where socket paths may not exist at prepare-time.
* **Artifact-only audit insertion**: The final `command` and `env.list` in `artifacts/<id>/` always contain the explicit socket mapping and environment variable.
* Fix is **defense-in-depth**: wrapper logic is safe and optional; runtime fallback is authoritative.

####  Local verification

* **Unit tests**:

  ```bash
  pytest -q test/unit --maxfail=1
  # Result: 1950 passed, 1 skipped, 1 xpassed in 23.64s
  ```

* **Containerized run with `--debug`**:

  ```text
  wrap_args_for_containerization: skipping explicit SSH_AUTH_SOCK bind-mount for RunnerConfig wrapper
  runtime-fallback: containerized=True sock=/tmp/ssh-.../agent.NNNN
  runtime-fallback: socket exists at /tmp/ssh-.../agent.NNNN
  runtime-fallback: '--env-file' found at index X, inserting mount at X
  ```

* **Artifacts (redacted)**:
  `artifacts/<id>/command`:

  ```json
  [
    "-v", "/tmp/ssh-.../agent.NNNN:/tmp/ssh-.../agent.NNNN",
    "-e", "SSH_AUTH_SOCK=/tmp/ssh-.../agent.NNNN"
  ]
  ```

  `artifacts/<id>/env.list`:

  ```text
  SSH_AUTH_SOCK=/tmp/ssh-.../agent.NNNN
  ```

* Confirmed socket visible and working inside the container (`SOCKET_OK`).

#### Screenshots

Screenshot 2025-10-15 165312
Screenshot 2025-10-15 165332
Screenshot 2025-10-15 165353

#### Files to review

* `src/ansible_runner/config/_base.py`
* `src/ansible_runner/runner.py`
* `test/unit/config/test_runner.py`
* `unit/test_automounts.py`

#### Tests added/updated

* Updated `test_runner.py` to assert correct podman argument ordering with automounts.
* Added `test_automounts.py` to cover:

  * Automount detection
  * Runtime fallback insertion
  * Edge cases with socket heuristics

#### Security note

Mounting a host’s SSH agent socket into a container grants that container access to the user’s agent.
This is an **intended behavior** when agent forwarding is required (e.g., for remote Git/SSH access), but users should be aware that this exposes their local agent to the containerized process.
This PR ensures this happens **only when the user has set up an agent and run in containerized mode** - preserving security expectations and user intent.

#### Limitations / follow-up

* Mounting of the user’s `~/.ssh` directory is **out of scope** and remains unchanged.
* If needed, that feature should be designed separately, with additional review for security and usability.
* CI limitations may prevent testing socket mounts directly. We recommend:

  * An integration job that asserts `[ -S "$SSH_AUTH_SOCK" ]` inside the container.
  * Review of recorded artifacts to confirm correct fallback and env insertion.
* Docker and Podman differ slightly in runtime flag handling - tests were verified with Podman; CI may need adaptation for both engines.

### Conclusion

This PR:

* **Fixes a confirmed regression** in `run()` → containerized EE → SSH_AUTH_SOCK behavior.
* **Restores socket mounting** using a safe wrapper + runtime fallback strategy.
* **Records artifacts** for auditing and test reproducibility.
* **Adds new unit tests** to prevent future regressions.
* **Supports edge cases** like WSL and rootless Podman with heuristic detection.

Fixes: #1396
Fixes: #689

### DIFF
--- a/src/ansible_runner/config/_base.py
+++ b/src/ansible_runner/config/_base.py
@@ -335,6 +335,16 @@ class BaseConfig:
         # Pexpect will error with non-string envvars types, so we ensure string types
         self.env = {str(k): str(v) for k, v in self.env.items()}
 
+        # If the calling process has an SSH_AUTH_SOCK set, but it was not
+        # included in loaded env files, propagate it into self.env so that
+        # automount logic can detect and mount the socket when containerizing.
+        try:
+            if 'SSH_AUTH_SOCK' not in self.env and 'SSH_AUTH_SOCK' in os.environ:
+                self.env['SSH_AUTH_SOCK'] = os.environ.get('SSH_AUTH_SOCK')
+        except Exception:
+            # best-effort only
+            pass
+
         debug('env:')
         for k, v in sorted(self.env.items()):
             debug(f' {k}: {v}')
@@ -398,7 +408,23 @@ class BaseConfig:
                                    labels: str | None = None
                                    ) -> None:
 
-        if src_mount_path is None or not os.path.exists(src_mount_path):
+        # Debug: log mount attempt and existence check
+        try:
+            exists = os.path.exists(src_mount_path) if src_mount_path is not None else False
+            debug(f"_update_volume_mount_paths: attempting to mount src={src_mount_path} exists={exists} dst={dst_mount_path} labels={labels}")
+        except Exception:
+            pass
+
+        # Allow lenient handling for ssh-agent sockets: some environments
+        # (WSL/rootless podman) create sockets under /tmp/ssh-*/agent.* where
+        # the path may not be visible at prepare time. If the src_mount_path
+        # looks like an ssh-agent socket, don't abort solely on existence.
+        if src_mount_path is None:
+            logger.debug("Source volume mount path is None")
+            return
+        exists = os.path.exists(src_mount_path)
+        is_probable_ssh_socket = isinstance(src_mount_path, str) and src_mount_path.startswith('/tmp/ssh-')
+        if not exists and not is_probable_ssh_socket:
             logger.debug("Source volume mount path does not exist: %s", src_mount_path)
             return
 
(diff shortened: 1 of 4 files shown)


Shrews:
Thank you for taking the time to submit this PR. However, we're declining this contribution. [As I stated in issue #1396](https://github.com/ansible/ansible-runner/issues/1396#issuecomment-2770388462), we aren't looking to add this functionality.

We value contributions from developers who have invested time understanding our architecture, participated in issue discussions, and can provide technical rationale grounded in hands-on experience with the problem space.

If you're interested in contributing to ansible-runner, we encourage you to:
  - Engage in the GitHub issues to understand the problems users are facing
  - Start with smaller, focused changes that demonstrate familiarity with the codebase
  - Participate in technical discussions with maintainers before investing significant development time

We appreciate your interest in the project.

PR #1466, opened 2025-10-24
Update GH runners in CI

Shrews:
Ubuntu 22.04 is now the minimum


CLAUDE.md
webknjaz, line 1: Did you mean to add this to the branch but not `devel`?

Shrews: nope. this was a quick test

PR #1142, opened 2022-10-10
Use `get_option` api from callback plugins

Akasurde:
##### SUMMARY

To load options, callback plugins can use `get_option` api.
Allow this functionality for other callback plugin.

Fixes: #1077

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
ansible_runner/display_callback/callback/awx_display.py

PR #1477, opened 2026-02-04
awx_display: correctly load original stdout callback options

Yannik:
This fixes the issue that options for the original stdout callback are not loaded/cannot be used with awx_display.

(example: even for the `default` callback, the `callback_result_format` option doesn't work, because `awx_display` only loads the options from its own `DOCUMENTATION` string, and not the options of the original stdout callback.)


Shrews:
https://github.com/ansible/ansible-runner/pull/1142 and https://github.com/ansible/ansible-runner/pull/1345 also seem to be fixing the same thing.

Shrews:
Closing in favor of #1142

PR #1479, opened 2026-02-10
fix: make rotate_artifacts accessible from settings file

### LINKED ISSUE #1415: make rotate_artifacts accessible from env/settings

dmzoneill:
fixes #1415

added rotate_artifacts to the settings loading section so it can be configured via the settings file, matching the pattern used by other config options like suppress_output_file and container_options.

now you can set rotate_artifacts in your settings file and it'll work as expected.

### DIFF
--- a/src/ansible_runner/config/_base.py
+++ b/src/ansible_runner/config/_base.py
@@ -281,6 +281,7 @@ class BaseConfig:
 
         self.suppress_output_file = self.settings.get('suppress_output_file', False)
         self.suppress_ansible_output = self.settings.get('suppress_ansible_output', self.quiet)
+        self.rotate_artifacts = self.settings.get('rotate_artifacts', self.rotate_artifacts)
 
         if 'fact_cache' in self.settings:
             if 'fact_cache_type' in self.settings:


Shrews (changes requested):
Please add tests.

src/ansible_runner/config/_base.py
```diff
@@ -281,6 +281,7 @@ def prepare_env(self, runner_mode: str = 'pexpect') -> None:
 
         self.suppress_output_file = self.settings.get('suppress_output_file', False)
         self.suppress_ansible_output = self.settings.get('suppress_ansible_output', self.quiet)
+        self.rotate_artifacts = self.settings.get('rotate_artifacts', self.rotate_artifacts)
```
Akasurde, line 284: Also, update docs here - https://docs.ansible.com/projects/runner/en/stable/intro/#env-settings-settings-for-runner-itself

PR #1490, opened 2026-02-24
Initial testing support for Python 3.15

Shrews:
- Add Python 3.15 to CI.
- Ignore a `DeprecationWarning` emitted from the use of `pexpect`.


Shrews:
Python 3.15 will not work with ansible-core until 2.22 releases. Holding off for a bit...

PR #1496, opened 2026-03-17
Remove fact cache API

Shrews:
* Removes fact cache API functionality
* Adds porting guides to the documentation, with a new guide for unreleased 2.5 version.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -38,6 +38,7 @@ Examples of this could include:
    intro
    install
    community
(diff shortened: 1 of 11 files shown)


Shrews:
Not sure we can remove the interface function params: https://github.com/ansible/awx/blob/643a9849df872e3b7a3f35d605a6042d66710d61/awx/main/tasks/jobs.py#L744-L749

PR #1517, opened 2026-07-01
CI test fixes: update Python for test container and add image prune retry

Shrews:
We haven't had a PR in a while, so tests have decayed:

- The `milestone` branch was failing `test_core_integration.py` tests because the version of Python in the test container image was 3.12, but 3.13 or higher is required.
- The `test_cleanup_images.py::test_cleanup_new_image` was failing `integration-devel` tests because `prune_image()` was failing due to a delay within Podman of cleaning up the stopped container. The change above seems to have exposed this clean up delay for $insertCrazyReasonHere, even though it passes in other integration tests, so build in a retry. Add retries, with a quick interval, for a better user experience. Re-raise the same `RuntimeError` exception after retries are exhausted to not break the API.

### DIFF
--- a/src/ansible_runner/cleanup.py
+++ b/src/ansible_runner/cleanup.py
@@ -179,23 +179,20 @@ def cleanup_images(images: list, runtime: str) -> int:
     return rm_ct
 
 
-def prune_images(runtime: str, retries: int = 1, retry_interval: float | int = .5) -> bool:
-    """Run the prune images command and return changed status"""
-    retries = max(retries, 0)
-    count = retries + 1
-    while True:
+def prune_images(runtime: str) -> bool:
+    """
+    Run the prune images command.
+
+    Since stopped containers may not be removed immediately, just try again, a couple of seconds max.
+    """
+    for attempt in range(1, 5):
         try:
-            stdout = run_command([runtime, 'image', 'prune', '-f'])
-            if not stdout or stdout == "Total reclaimed space: 0B":
-                return False
-            return True
-        except RuntimeError as e:
-            # Image pruning can fail if a container exists that used an image. Since stopped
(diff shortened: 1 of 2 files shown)


src/ansible_runner/cleanup.py
```diff
@@ -178,12 +179,23 @@ def cleanup_images(images: list, runtime: str) -> int:
...
+            if not stdout or stdout == "Total reclaimed space: 0B":
+                return False
+            return True
+        except RuntimeError as e:
+            # Image pruning can fail if a container exists that used an image. Since stopped
+            # containers may not be removed immediately, we might need to just try again.
+            count -= 1
+            time.sleep(retry_interval)
+            if count == 0:
+                raise e
```
nitzmahone: Could tighten up the loop impl a bit with for/range, and since you're still in the exception handler, the `raise` should just be naked, like:

```suggestion
    for attempt in range(0, (retries := max(retries, 0)) + 1):
        try:
            return (stdout := run_command([runtime, 'image', 'prune', '-f'])) and stdout != "Total reclaimed space: 0B"
        except RuntimeError as ex:
            # Image pruning can fail if a container exists that used an image. Since stopped
            # containers may not be removed immediately, we might need to just try again.
            if attempt == retries:
                raise
            time.sleep(retry_interval)
```

If it's not really public API, you could simplify it even further by just hardcoding the values instead of making them args that need validation/coercion. At a glance it doesn't look like someone could change the behavior without changing the code anyway.

Shrews: Valid points. And now that I look at this a bit more in-depth, the whole `stdout` thing makes no sense. Maybe it did at one time, but the `prune` command does not output those strings anymore. Nor does the return value change based on if it did anything or not. Will probably just change this to always return `True`.

Issue #1516, opened 2026-06-30
support apple/container support on MacOS as Container Runtime Environment


jwalzer:
### Summary

tags: 
  - Feature Request
  - MacOS
  - Container Runtime
  - apple/container

---
Apple has Released with MacOS 26 its own Container Runtime Environment:
see https://github.com/apple/container

This is projected to be the superior container solution on MacOS over Podman Desktop and Docker Desktop.

Currently Ansible-Navigator supports Docker and Podman as CRE. 
This Request is to support also the container CRE.

sivel:
Thank you for your interest in ansible-runner. This is not something we have plans to implement. We are largely able to support podman and docker because of the fact that podman intends to almost be a 1:1 implementation of docker for the use cases we have. The apple container system is not close, and would require a separate implementation.

If you have further questions please stop by our forum or Matrix:

   * Forum: https://forum.ansible.com/
   * Matrix: https://matrix.to/#/#users:ansible.com

PR #1518, opened 2026-07-03
limited Support for Apple container

jwalzer:
This PR adds support for apple/container as CRE to ansible runner.

Included Tests and Documentation.

This would solve https://github.com/ansible/ansible-runner/issues/1516 if it wasn't closed already.

### DIFF
--- a/docs/apple_container.rst
+++ b/docs/apple_container.rst
@@ -1,0 +1,199 @@
+.. _apple_container:
+
+Apple Container Notes
+=====================
+
+This document gathers the ansible-runner specific behavior, requirements, and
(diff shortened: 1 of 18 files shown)


Shrews:
As mentioned in #1516, we don't have plans to support this at this time. Support is more than just adding the feature. We'd have to consider the maintenance burden, enhancing our CI environment (currently there are no OSX runners in our CI), and, most importantly, testing the impact of those changes within AAP, which is no small task. Docker and Podman are already difficult to maintain due to several quirks in behavior, many not seen in the code itself, but how it behaves in CI.

This may be something we reconsider in the future, but presently we are unwilling to add this feature.

PR #1523, opened 2026-07-22
docs: clarify container volume mounts vs bwrap isolation paths

### LINKED ISSUE #757: `container_volume_mounts` should be documented
It would be helpful if the following two points were documented.

- The `process_isolation_*` options are only for `bwrap` and not for `docker` or `podman`.
- `docker` and `podman` should be able to use `container_volume_mounts`.

[The documentation lists options](https://ansible-runner.readthedocs.io/en/stable/intro.html#env-settings-settings-for-runner-itself) such as `process_isolation_show_paths` in order to make arbitrary paths available to the playbook.

However, according to the code base, this seems to be for `bwrap` only, and does not work with `docker` or `podman`.

For `docker` and `podman`, we can use `container_volume_mounts` instead, but this is not documented.

Atharva-Ramawat:
### Summary
Fixes #757

This PR updates the documentation and API docstrings to clarify the distinction between process isolation paths and container volume mounts. 

Previously, the documentation implied that `process_isolation_*_paths` applied universally, which caused confusion for users running execution environments in Docker or Podman. 

### Changes made
* **`docs/intro.rst`**: Added a note clarifying that `process_isolation_hide_paths`, `process_isolation_show_paths`, and `process_isolation_ro_paths` only apply when using `bwrap`. Explicitly documented `container_volume_mounts` as the required alternative for container engines.
* **`src/ansible_runner/interface.py`**: Updated the `:param` docstrings to reflect this exact same distinction so developers referencing the Python API have the correct context. Also fixed a missing closing quote on the `container_volume_mounts` example string.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -155,6 +155,10 @@ The process isolation settings are meant to control the process isolation featur
 * ``process_isolation_hide_paths``: ``None`` Path or list of paths on the system that should be hidden from the playbook run.
 * ``process_isolation_show_paths``: ``None`` Path or list of paths on the system that should be exposed to the playbook run.
 * ``process_isolation_ro_paths``: ``None`` Path or list of paths on the system that should be exposed to the playbook run as read-only.
+* ``container_volume_mounts``: ``None`` List of volume mounts to use when running inside a container engine (Docker or Podman). This should be used instead of the ``process_isolation_*_paths`` options for container environments.
+
+.. note::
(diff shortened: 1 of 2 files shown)


s-hertel (approved):

Shrews (approved):

Shrews:
@Atharva-Ramawat Could you please rebase and fix the linting errors?

PR #1524, opened 2026-07-22
[ci]: Fix container usage in tests

Shrews:
The `integration-devel` tests were not taking advantage of the pre-built container image that is specified by RUNNER_TEST_IMAGE_NAME, so it was not being reused.

The `container_image_devel` test fixture, meant for the Ansible core integration tests, wasn't properly being used because it was still honoring RUNNER_TEST_IMAGE_NAME. We had a mixed testing state where these tests DO work correctly for `integration-devel` since it wasn't receiving that env var, but for the other `integration` tests, they were not functioning correctly. This set of changes should make the testing world right again.

### DIFF
--- a/.github/reusables/tox-dev/workflow/reusable-tox/hooks/post-src-checkout/action.yml
+++ b/.github/reusables/tox-dev/workflow/reusable-tox/hooks/post-src-checkout/action.yml
@@ -19,12 +19,16 @@ runs:
   using: composite
   steps:
   - name: Build a wheel
-    if: fromJSON(inputs.calling-job-context).toxenv == 'integration'
+    if: >-
+      contains(fromJSON('["integration", "integration-devel"]'),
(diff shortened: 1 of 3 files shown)


jborean93 (approved):

.github/reusables/tox-dev/workflow/reusable-tox/hooks/post-src-checkout/action.yml
```diff
@@ -19,12 +19,16 @@ runs:
   using: composite
   steps:
   - name: Build a wheel
-    if: fromJSON(inputs.calling-job-context).toxenv == 'integration'
+    if: >-
+      contains(fromJSON('["integration", "integration-devel"]'),
```
webknjaz, line 23: @Shrews an alternative (a little less precise) idea could be using a `startsWith()`.

PR #1528, opened 2026-08-06
docs: improve run_async documentation and examples (#1071)

### LINKED ISSUE #1071: Improve run_async documentation and examples
We are trying to get logs [data](https://ansible-runner.readthedocs.io/en/stable/ansible_runner/#ansible_runner.runner.Runner.events) about an event when it occurs directly in Python.
Because we are using 

`runner_async_thread, runner_object = ansible_runner.interface.run_async(...` 

There appears to be no clear way in the documentation to see the output of playbook tasks as they run.

The [documentation for the `run_async()` helper function](https://ansible-runner.readthedocs.io/en/stable/python_interface/#run-async-helper-function) says:

> The **Runner** object can be inspected during execution.

… but does not elaborate or provide any clues regarding precisely how that inspection can be performed or what information can be inspected.

I started looking at [event_callback](https://ansible-runner.readthedocs.io/en/stable/ansible_runner/#ansible_runner.runner.Runner.event_callback) but as far as I can understand it's not really usable outside of `Runner`'s context.

The only plausible way to get those data that I found is to create a custom plugin similar to `ansible-runner-http`. Is that the only way?

There is also an [old open question](https://stackoverflow.com/questions/56434195/how-to-read-the-events-from-ansible-runner-programatically) on StackOverflow that talks about this need.

Atharva-Ramawat:
### Summary
Fixes #1071

This PR updates the Python interface documentation for `run_async()` to clarify how users can inspect the `Runner` object during execution. Previously, the documentation stated the object could be inspected but did not provide guidance on how to do so or how to handle duplicate event reads.

### Changes Made
* **Clarified Event Inspection:** Added explicit instructions explaining that users can query the `runner_object.events` generator to read event data asynchronously.
* **Added Duplicate Warning:** Added a `.. warning::` block highlighting that reading the `events` attribute repeatedly will yield duplicate events, advising users to parse and coalesce them using the `uuid` and `counter` keys.
* **Added Practical Example:** Provided a concrete Python code block demonstrating how to properly poll for events in a `while runner_object.rc is None:` loop while filtering out already-processed events using a set of seen UUIDs.

### DIFF
--- a/docs/python_interface.rst
+++ b/docs/python_interface.rst
@@ -31,7 +31,49 @@ foreground and return the :class:`Runner <ansible_runner.runner.Runner>` object 
 :meth:`ansible_runner.interface.run_async`
 
 Takes the same arguments as :meth:`ansible_runner.interface.run` but will launch **Ansible** asynchronously and return a tuple containing
-the ``thread`` object and a :class:`Runner <ansible_runner.runner.Runner>` object. The **Runner** object can be inspected during execution.
+the ``thread`` object and a :class:`Runner <ansible_runner.runner.Runner>` object. 
+
+The returned **Runner** object can be inspected while the playbook is executing. You can query the ``runner_object.events`` generator to read event data as it becomes available.
+
+.. warning::
+    Because the ``events`` attribute reads all event files present in the artifacts directory at the time of access, accessing it repeatedly in a loop will yield duplicate events. You must parse and coalesce the events yourself using the ``uuid`` and ``counter`` keys present in the JSON event data to filter out duplicates and maintain the correct order.
+
+The playbook execution is complete when the ``rc`` attribute of the ``Runner`` object changes from ``None`` to an integer representing the final return code.
+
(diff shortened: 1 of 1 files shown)


Shrews (commented):
Thanks for giving this some attention. However, as I was fairly new to the project at the time of my comment on the referenced issue, my statement there wasn't entirely correct.

The `events` property returns a generator that one can iterate through while the process is running. It will handle filtering out already "seen" events, and even do the final read to get events after the playbook run finishes. So its use is for tailing the event stream as the playbook runs, so it doesn't just automatically return events at the point of property access.

So this example code isn't correct. Rather than example code, I'd rather just see verbage explaining that concept. And I don't think it belongs here under `run_async()` documentation. Better to explain that in the `Runner.events` documentation. I've opened [PR #1531](https://github.com/ansible/ansible-runner/pull/1531) that should close this out for us. As such, I'm just going to close this PR.

Shrews:
Closing in favor of #1531

PR #1529, opened 2026-08-06
Add process_isolation_hide_envvars to unset variables in the bwrap sandbox

akinarcak:
##### SUMMARY

When process isolation uses bwrap, the sandboxed command inherits the environment of the process that started Runner. For an application that embeds Runner as a library, that means every playbook run receives the application's own environment — database credentials, session signing keys, API tokens — even though nothing in the playbook asked for them, and even though the caller has carefully limited which paths the run can see.

Callers can already hide filesystem paths from a run (`process_isolation_hide_paths`). This adds the equivalent for the environment.

`process_isolation_hide_envvars` takes an `fnmatch` pattern, or a list of them, and emits `--unsetenv` for each matching variable in the environment the sandboxed command would otherwise inherit:

```python
ansible_runner.run(
    private_data_dir='/tmp/demo',
    playbook='site.yml',
    process_isolation=True,
    process_isolation_executable='bwrap',
    process_isolation_hide_envvars=['*_TOKEN', 'DATABASE_URL', 'DJANGO_SECRET_KEY'],
)
```

It defaults to `None`, so nothing changes for anyone who does not opt in. The variables also remain visible to Runner itself — only the sandboxed command loses them — so this does not interfere with variables Runner needs, nor with anything a caller deliberately passes through `envvars`.

Like the other `process_isolation_*` path settings, it applies to bwrap only; the docs note has been extended to say so. It is also settable through `env/settings`.

##### ORIGIN

The idea comes from a downstream patch the JumpServer project carries on a fork of this repository ([`8f9316545`](https://github.com/jumpserver-dev/ansible-runner/commit/8f9316545), by @w940853815 and ibuler, both credited as co-authors on the commit). That version strips every variable whose name contains one of `KEY`, `PASSWORD`, `TOKEN`, `SECRET`, `PASSWD`, `REDIS_` or `DB_`, unconditionally.

That is the right instinct but the wrong default for upstream: matching on `PASSWORD` as a substring would also drop `ANSIBLE_VAULT_PASSWORD_FILE`, and `KEY` would drop `AWS_SECRET_ACCESS_KEY` — including when a caller passed it deliberately. Making the patterns a caller-controlled setting keeps the capability while leaving the default behaviour untouched.

The fork carries a second change in the same commit, replacing `--symlink usr/lib /lib` with `--ro-bind /lib /lib`. That one is deliberately **not** included here: `--ro-bind` fails when the source does not exist, so it would break on systems without a real `/lib64`, and it is unrelated to the environment question.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

`ansible_runner.config.runner`

##### ADDITIONAL INFORMATION

Two unit tests are added alongside the existing bwrap tests in `test/unit/config/test_runner.py`, covering a list of patterns and a bare string, and asserting that the hidden variable is still present in `rc.env`.

`test/unit/config/test_runner.py` on this branch: 40 passed, against 38 passed on unmodified `devel` in the same environment. `flake8` is clean on both changed files.

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -155,10 +155,11 @@ The process isolation settings are meant to control the process isolation featur
 * ``process_isolation_hide_paths``: ``None`` Path or list of paths on the system that should be hidden from the playbook run.
 * ``process_isolation_show_paths``: ``None`` Path or list of paths on the system that should be exposed to the playbook run.
 * ``process_isolation_ro_paths``: ``None`` Path or list of paths on the system that should be exposed to the playbook run as read-only.
+* ``process_isolation_hide_envvars``: ``None`` Name pattern, or list of name patterns, of environment variables to unset inside the isolated playbook run. Patterns use :mod:`fnmatch` syntax and are matched case-sensitively against the environment the run would otherwise inherit, so an application embedding **Runner** can keep its own credentials (for example ``*_TOKEN`` or ``DATABASE_URL``) out of the sandbox without having to sanitize its own process environment.
 * ``container_volume_mounts``: ``None`` List of volume mounts to use when running inside a container engine (Docker or Podman). This should be used instead of the ``process_isolation_*_paths`` options for container environments.
 
 .. note::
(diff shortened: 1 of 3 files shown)


w940853815 (approved):

Shrews:
We are planning to deprecate `bwrap` in the next release, and then eventually remove it. As such, we don't want to add anymore features for bubblewrap at this point.

PR #1546, opened 2026-09-08
fix: Count bytes actually read in unstream_dir and ensure decoding handles quantum boundaries

### LINKED ISSUE #1545: unstream_dir: short read corrupts the transfer or stalls it forever (remaining decremented by chunk_size instead of len(data))
**Environment**

- ansible-runner `2.4.1.dev70+g8dcdadbd8` (as shipped in the AWX/Ascender 25.6.x control plane image)
- Ascender 25.6.1, RKE2 cluster, job executed on the control plane node

**Symptom**

A long-running job (8.5h, ~1.6M events, ~75 MB of job output) stopped producing events mid-run and stayed in `running` state for 38+ hours. No error in any log. The last events were `playbook_on_task_start` / `runner_on_start` with no completion. Cancelling and re-running did not help until the file below was patched.

**Root cause**

`unstream_dir()` in `src/ansible_runner/utils/streaming.py` decrements `remaining` by the requested chunk size instead of the number of bytes actually read:

```python
while remaining != 0:
    chunk_size = min(chunk_size, remaining)

    data = source.read(chunk_size)
    target.write(data)

    remaining -= chunk_size
```

`Base64IO.read()` over a pipe may legally return fewer bytes than requested. On a short read `remaining` goes negative, `min(chunk_size, remaining)` turns negative, and the transfer either produces a truncated zip (`BadZipFile`, `Incorrect padding`, `OSError: EINVAL` — the exception variant reported in #998) or blocks indefinitely on a read that can never be satisfied (silent hang). Which of the two you get depends on where the short read lands.

**Fix**

Decrement by the number of bytes actually read:

```diff
--- a/src/ansible_runner/utils/streaming.py
+++ b/src/ansible_runner/utils/streaming.py
@@
                 while remaining != 0:
                     chunk_size = min(chunk_size, remaining)
 
                     data = source.read(chunk_size)
                     target.write(data)
 
-                    remaining -= chunk_size
+                    remaining -= len(data)
```

With this one-line change the same workload (same job, same artifacts) completes normally; without it the same job froze reproducibly at the artifact transfer. Happy to send this as a PR if you take fixes in this area.

**Related**

- #998 reports the exception variant of the same accounting bug in `unstream_dir()` on large artifacts. It was closed without a change to this line, which is still present on `devel` as of today.

Shrews:
## Summary

Contains fixes for:

- Correcting the remaining bytes to read in `unstream_dir()`
- Makes Base64IO recognize and respect quantum boundaries so that only decoded data is returned from a `read()`, avoiding "Incorrect padding" errors.

### Details

unstream_dir() decremented `remaining` by the requested chunk size rather than the number of bytes returned. Base64IO.read() issues a single read() on the wrapped stream and does not loop to fill the request, so any short read - which is what a raw pipe or an unbuffered socket does once the payload outgrows the kernel buffer - silently drops data while still being counted as transferred.

The archive is truncated (BadZipFile, or "Incorrect padding" when the short read does not land on a base64 boundary), and worse, the reader stops draining the stream while the writer still has data to send, so the writer blocks forever in stream_dir() on a full send buffer. That surfaces as a job wedged in `running` with no error logged anywhere.

Decrement by len(data) and raise on a premature EOF. The guard matters: without it an early EOF makes the loop spin, since read() keeps returning b'' and `remaining` never reaches zero.

Adds an end-to-end regression test that drives Transmitter -> Worker over a socketpair whose read side is unbuffered, plus unit tests covering short reads over a real os.pipe and the truncated-stream case.

Fixes #1545 

Co-Authored-By: Claude Code

### DIFF
--- a/src/ansible_runner/utils/base64io.py
+++ b/src/ansible_runner/utils/base64io.py
@@ -76,6 +76,7 @@ class Base64IO(io.IOBase):
         """
         # set before the attr check as we may reach close() after that check fails
         self.__read_buffer = b""
+        self.__encoded_read_buffer = b""
         self.__write_buffer = b""
 
         required_attrs = ("read", "write", "close", "closed", "flush")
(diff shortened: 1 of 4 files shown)


Akasurde (commented):
Code wise LGTM

PR #1547, opened 2026-09-09
Remove the python-daemon dependency

### LINKED ISSUE #379: runner is dependent upon a deprecated package - lockfile
Runner's python dependency chain requires daemon, which in turn require lockfile (at least that's how the rpms are built!). The issue is that lockfile is deprecated - https://pypi.org/project/lockfile/ so any security related issues relating to that package may not be addressed leading to a potential exposure in runner itself.

This is also covered in a downstream BZ - https://bugzilla.redhat.com/show_bug.cgi?id=1758507

Shrews:
Remove the `python-daemon` dependency
    
`python-daemon` is used in exactly one place, and only with its defaults. It depends on `lockfile`, which has been unmaintained since 2015. Every drop-in replacement on PyPI is staler still, so implement the PEP 3143 essentials we actually need in a new private module instead.
    
Dropping `python-daemon`'s defaults also fixes several bugs in `start`:
    
* A stale pid file no longer blocks all further use of a private_data_dir. PIDLockFile had no stale detection at all, so a kill -9 or a reboot left `start` failing with a traceback until someone deleted the file by hand.
* `--logfile` works again. DaemonContext.open() closed every inherited file descriptor, including the one belonging to the logging FileHandler that `output.set_logfile()` had just opened.
* Artifacts are no longer world writable. The umask of the invoking process is inherited rather than reset to 0.
* Starting a second daemon reports the running PID instead of raising lockfile.AlreadyLocked.
* The daemon's stderr is appended to `<private_data_dir>/daemon.log`, which was already being created for it, rather than discarded. It is line buffered, because `stop` reaps the daemon with SIGKILL and a block buffered stream would discard up to a bufferful of it.

Fixes #379

### DIFF
--- a/src/ansible_runner/utils/_daemonize.py
+++ b/src/ansible_runner/utils/_daemonize.py
@@ -76,11 +76,10 @@ def pid_is_running(pid: int) -> bool:
         return False
     try:
         os.kill(pid, 0)
-    except ProcessLookupError:
-        return False
     except PermissionError:
-        # The process exists, we are simply not allowed to signal it.
-        return True
+        # The process exists, we are simply not allowed to signal it. This clause must
+        # come first: PermissionError is a subclass of OSError.
+        pass
     except OSError:
         return False
     return True
--- a/src/ansible_runner/utils/_daemonize.py
+++ b/src/ansible_runner/utils/_daemonize.py
@@ -1,0 +1,511 @@
+'''
+Minimal, dependency free POSIX daemonization support.
+
+This module is private to ``ansible_runner`` and covers only what the
+``ansible-runner start`` command needs. It is not part of the public API.
+'''
+
+from __future__ import annotations
+
+import atexit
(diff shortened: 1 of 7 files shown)


src/ansible_runner/utils/_daemonize.py
```diff
@@ -0,0 +1,511 @@
...
+    try:
+        os.kill(pid, 0)
+    except ProcessLookupError:
+        return False
+    except PermissionError:
+        # The process exists, we are simply not allowed to signal it.
+        return True
+    except OSError:
+        return False
+    return True
```
Akasurde:
```suggestion
    except (OSError, ProcessLookupError):
        return False
    except PermissionError:
        # The process exists, we are simply not allowed to signal it.
        pass
    return True
```

Akasurde (commented):
LGTM

PR #1549, opened 2026-09-18
AAP-78908 Explicitly pass tls-verify value

lallen92:
## Description:
With a registry credential set to `verify_ssl: false`, pulling an EE image from a self-signed-cert registry in containerized AAP still failed with `x509: certificate signed by unknown authority`.

ansible-runner only disabled TLS verification by writing a `registries.conf` file and pointing a `CONTAINERS_REGISTRIES_CONF` env var at it on the *client* process, but containerized AAP's task container talks to podman over a remote socket (`CONTAINER_HOST`), so the actual image pull runs in a separate podman service process that never sees that client-side env var or file.

## Proposed Fix:
Adding `--tls-verify=false` as an explicit podman CLI argument (alongside the existing `--authfile=`) sends the setting as part of the request itself, which podman's remote/service API does honour, unlike a local env var pointing at a config file only the client process can see.

## Testing:
Local e2e testing (Not yet validated against a real two-host AAP 2.6 containerized install).
Run against a self-signed registry on `localhost:5443` behind macOS's podman-machine VM, which is client/server just like `CONTAINER_HOST`:
```
$ CONTAINERS_REGISTRIES_CONF=/…/registries.conf podman pull localhost:5443/test/busybox:latest
Error: unable to copy from source docker://localhost:5443/test/busybox:latest: initializing source
docker://localhost:5443/test/busybox:latest: fetching manifest latest in localhost:5443/test/busybox:
pinging container registry localhost:5443: Get "https://localhost:5443/v2/": tls: failed to verify
certificate: x509: certificate signed by unknown authority
```
registries.conf contained `[[registry]] location = "localhost:5443" \n insecure = true`, exactly what ansible-runner writes, and had zero effect, reproducing the reported error verbatim.
```
$ podman pull --tls-verify=false localhost:5443/test/busybox:latest
Trying to pull localhost:5443/test/busybox:latest...
Getting image source signatures
...
(succeeds)
```


djulich (approved):
Change looks great: Exactly scoped, guarded well, consistent with existing patterns, good test coverage - good job!

nitzmahone:
I think this does actually cause a security regression: `podman-remote` (by design, it seems) relies on the server-side registry config for short-name resolution, mirror config, etc. The current mechanism disables TLS verification only for the primary registry host- any other default or server-configured short-names/mirrors continue to use the configured/default verification settings for those registries. IIUC this change disables verification for all pulls, so a container manifest with includes/links/base layers from other registries would now skip verification on all those connections, instead of only for the primary registry.

Is that fatal to this approach? :shrug: It opens up new attack vectors that weren't possible before, but OTOH disabling verification on the host serving the primary manifest blows a pretty big hole in the entire chain anyway...

lallen92:
> I think this does actually cause a security regression: `podman-remote` (by design, it seems) relies on the server-side registry config for short-name resolution, mirror config, etc. The current mechanism disables TLS verification only for the primary registry host- any other default or server-configured short-names/mirrors continue to use the configured/default verification settings for those registries. IIUC this change disables verification for all pulls, so a container manifest with includes/links/base layers from other registries would now skip verification on all those connections, instead of only for the primary registry.
> 
> Is that fatal to this approach? 🤷 It opens up new attack vectors that weren't possible before, but OTOH disabling verification on the host serving the primary manifest blows a pretty big hole in the entire chain anyway...

Would it be better then if we only added `--tls-verify=false` when podman is actually running remote? That would cover the scenario where registries.conf doesn't work and would leave local podman alone. However it still wouldn't fully address your concern...

